PR #10, opened 2026-01-28
Claude/openrouter api integration d alec

coltfour5:


mvanhorn:
Thanks for this PR! The OpenRouter integration is thorough but was too large a change for V2 (1000+ lines, adds MCP server). We went with a simpler model fallback chain instead. Appreciate the effort though. 🙏

PR #14, opened 2026-02-01
Simplify to WebSearch-first, make API keys optional

thangman1:
## Summary
Restructured last30days to prioritize Claude Code's built-in WebSearch as the primary research method, with optional API keys for enhanced engagement metrics. This removes the complexity of mandatory API key setup while maintaining the ability to access Reddit and X data for users who want it.

## Key Changes

- **WebSearch-first architecture**: Updated documentation to position Claude Code's built-in WebSearch as the default research method that works immediately with zero configuration
- **Optional API keys**: Moved OpenAI and xAI API keys from required to optional, with clear messaging about what they unlock (engagement metrics only)
- **Simplified installation**: Removed API key setup from the main installation flow; moved to "Optional: Enhanced Mode" section
- **Updated research execution**: Rewrote research instructions to focus on WebSearch queries as the primary data gathering method
- **Removed mode complexity**: Eliminated references to "Full Mode", "Partial Mode", and "Web-Only Mode" in favor of simpler "default" vs "enhanced" framing
- **Cleaner output**: Simplified stats display and removed promotional messaging about missing API keys
- **Updated skill description**: Changed from "Reddit + X + Web" to "Claude Code's WebSearch" as the primary capability

## Implementation Details

- Installation now requires only cloning the repo; API key setup is entirely optional
- Research execution focuses on 3-5 WebSearch queries with specific patterns for each query type (RECOMMENDATIONS, NEWS, PROMPTING, GENERAL)
- Synthesis instructions emphasize grounding analysis in actual WebSearch results rather than pre-existing knowledge
- Stats display simplified to show web source counts and domains rather than engagement metrics
- Removed all references to script modes and API key detection logic from user-facing documentation
- Maintained backward compatibility: users with existing API keys can still use enhanced mode

## Benefits

- **Lower barrier to entry**: Users can start using the skill immediately without API key setup
- **Clearer value proposition**: WebSearch is reliable and works out of the box; API keys are a bonus for those who want them
- **Reduced maintenance**: Fewer code paths to maintain and fewer edge cases around API key availability
- **Better user experience**: No confusing mode detection or promotional messaging about missing keys

https://claude.ai/code/session_019qJjtPgVJUwy6yGvDEbHbd

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,23 @@
 ## Installation
 
 ```bash
-# Clone the repo
+# Clone the repo - that's it!
 git clone https://github.com/mvanhorn/last30days-skill.git ~/.claude/skills/last30days
+```
 
-# Add your API keys
+**Zero configuration required.** Works immediately with Claude Code Pro's built-in WebSearch.
+
+### Optional: Enhanced Mode with API Keys
+
+For users who want Reddit engagement metrics (upvotes, comments) and X metrics (likes, reposts), you can add optional API keys:
+
+```bash
 mkdir -p ~/.config/last30days
 cat > ~/.config/last30days/.env << 'EOF'
+# Optional: For Reddit research with engagement metrics
 OPENAI_API_KEY=sk-...
+
+# Optional: For X/Twitter research with engagement metrics
 XAI_API_KEY=xai-...
 EOF
 chmod 600 ~/.config/last30days/.env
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks for this PR! V2 already does WebSearch in parallel alongside Reddit/X API searches, so we went a different direction. Appreciate the thought behind making API keys optional though — Bird CLI now makes X search free, which addresses part of the same goal. 🙏

acesky61:
How do I go about installing this new version? It looks like Claude has the native search for X but I can't get anything to scan Reddit.

PR #65, opened 2026-03-11
Consolidate query/relevance modules and improve search quality

j-sperling:
## Summary

**Depends on: PR #66 and #67**

This PR started as a refactor of duplicated query parsing and token-overlap relevance logic, then carried the search-quality fixes needed to make the consolidated code actually improve results.

At this point the branch does four things:

### 1. Consolidates duplicated query and relevance logic
- adds `scripts/lib/query.py` and replaces 7 duplicated `_extract_core_subject()` implementations
- adds `scripts/lib/relevance.py` and replaces 4 duplicated tokenization / overlap scorers
- preserves per-platform behavior through parameterized wrappers instead of one-size-fits-all logic

### 2. Replaces placeholder relevance with real ranking signals
- computes relevance for sources that previously carried hardcoded defaults
- applies post-retrieval relevance filtering across sources
- uses query-type-aware tiering / tiebreaking from `#67`
- removes the artificial low-end relevance floor so generic token-only matches do not survive by default

### 3. Fixes concrete false-positive paths found during reviewer-requested comparisons
- Reddit relevance is now title-first, so long selftext does not overpower a weak title
- Polymarket no longer expands low-signal standalone terms like `odds` / `pro`
- Polymarket no longer gets bogus boosts from short binary outcomes like `No` matching unrelated queries by substring
- prompt-oriented and animation-oriented benchmark queries are classified earlier so they do not fall through to the `breaking_news` default

### 4. Adds an optional local search-quality evaluator
- `scripts/evaluate_search_quality.py`
- compares a baseline revision against the candidate checkout
- computes deterministic stability metrics (`Jaccard`, retention, per-source overlap)
- optionally calls Gemini for judged metrics (`Precision@5`, `nDCG@5`, source-coverage recall)
- isolates child runs from local browser-cookie config so historical revisions can be evaluated locally without Keychain popups

## Why

The duplicated helpers made it hard to reason about or test search quality consistently across sources. More importantly, several sources were still carrying placeholder or overly permissive relevance behavior, which made the final ranking too sensitive to generic tokens and high-engagement off-topic results.

The branch now has enough shared infrastructure and local evaluation support to make quality changes measurable instead of anecdotal.

### Automated
- `uv run python -m unittest`
- current full suite: **631 tests passing**

### Reviewer-requested comparison set
Ran the 5-topic side-by-side comparison requested in review against:
- `origin/main`
- `#66 + #67 + #65` stack

Topics:
1. `nano banana pro prompting`
2. `codex vs claude code`
3. `anthropic odds`
4. `kanye west`
5. `remotion animations for Claude Code`

The raw comparison package and follow-up analysis are in-thread.

### Judged local evaluation
I reran the full 5-topic judged suite locally after the latest fixes.

Macro judged metrics versus `origin/main`:
- `Precision@5`: `0.56 -> 0.68`
- `nDCG@5`: `0.41 -> 0.58`
- `source_coverage_recall`: `0.82 -> 0.95`

Per-topic highlights:
- `codex vs claude code`: `P@5 0.60 -> 1.00`, `nDCG@5 0.55 -> 1.00`
- `anthropic odds`: `P@5 0.40 -> 0.80`, `nDCG@5 0.21 -> 0.59`
- `kanye west`: `P@5 0.60 -> 0.60`, `nDCG@5 0.46 -> 0.62`
- `remotion animations for Claude Code`: `P@5 0.60 -> 0.60`, `nDCG@5 0.33 -> 0.35`

## Remaining caveat

The one query that is still weaker than `main` on the judged set is `nano banana pro prompting`.

That query is materially cleaner than before because the worst Polymarket false positives are gone, but it is still the weakest benchmark topic on this branch and the one I would continue tuning next if we want to keep pushing quality before merge.

## Operational impact

No migration is required for normal runtime use.

The new evaluator is optional and local-only. On this workspace it prefers `GOOGLE_API_KEY`, and also accepts `GEMINI_API_KEY` / `GOOGLE_GENAI_API_KEY`.

### DIFF
--- a/scripts/lib/query_type.py
+++ b/scripts/lib/query_type.py
@@ -7,7 +7,7 @@ QueryType = Literal["product", "concept", "opinion", "how_to", "comparison", "br
 
 # Pattern-based classification (no LLM, no external deps)
 _PRODUCT_PATTERNS = re.compile(
-    r"\b(price|pricing|cost|buy|purchase|deal|discount|subscription|plan|tier|free tier|alternative)\b", re.I
+    r"\b(price|pricing|cost|buy|purchase|deal|discount|subscription|plan|tier|free tier|alternative|prompt|prompts|prompting|template|templates)\b", re.I
 )
 _CONCEPT_PATTERNS = re.compile(
     r"\b(what is|what are|explain|definition|how does|how do|overview|introduction|guide to|primer)\b", re.I
@@ -16,7 +16,7 @@ _OPINION_PATTERNS = re.compile(
     r"\b(worth it|thoughts on|opinion|review|experience with|recommend|should i|pros and cons|good or bad)\b", re.I
 )
 _HOWTO_PATTERNS = re.compile(
-    r"\b(how to|tutorial|step by step|setup|install|configure|deploy|migrate|implement|build a|create a)\b", re.I
+    r"\b(how to|tutorial|step by step|setup|install|configure|deploy|migrate|implement|build a|create a|animation|animations|video workflow|render pipeline)\b", re.I
 )
 _COMPARISON_PATTERNS = re.compile(
     r"\b(vs\.?|versus|compared to|comparison|better than|difference between|switch from)\b", re.I
--- a/tests/test_query_type.py
+++ b/tests/test_query_type.py
@@ -20,6 +20,7 @@ class TestDetectQueryType(unittest.TestCase):
         self.assertEqual(detect_query_type("cursor IDE pricing"), "product")
         self.assertEqual(detect_query_type("is Claude Pro worth the cost"), "product")
         self.assertEqual(detect_query_type("best free tier LLM API"), "product")
+        self.assertEqual(detect_query_type("nano banana pro prompting"), "product")
 
     def test_concept_queries(self):
         self.assertEqual(detect_query_type("what is WebTransport"), "concept")
@@ -35,6 +36,7 @@ class TestDetectQueryType(unittest.TestCase):
         self.assertEqual(detect_query_type("how to deploy on Vercel"), "how_to")
         self.assertEqual(detect_query_type("tutorial for building MCP servers"), "how_to")
         self.assertEqual(detect_query_type("step by step Kubernetes setup"), "how_to")
+        self.assertEqual(detect_query_type("remotion animations for Claude Code"), "how_to")
 
     def test_comparison_queries(self):
         self.assertEqual(detect_query_type("cursor vs windsurf"), "comparison")


mvanhorn (commented):
The deduplication work here is solid - consolidating 7 copies of `_extract_core_subject()` and 4 copies of `_compute_relevance()` into shared modules is a clear win. The parameterization preserves per-platform behavior, and the 38 new tests give good coverage.

However, this PR bundles behavior changes with the refactor, and several of those behavior changes reverse deliberate design decisions or add risk:

**`--dateafter` in yt-dlp (youtube_yt.py:107):** The code previously had a comment explaining why `--dateafter` was removed: "YouTube search returns relevance-sorted results and strict date filtering returns 0 for evergreen topics like 'thumbnail tips'." The PR adds it back with a note that "Python soft filter handles the fallback" - but the soft filter only works if yt-dlp returns results. With `--dateafter`, yt-dlp returns 0 results for evergreen topics entirely, so there's nothing to fall back on. This will silently break YouTube results for a category of queries.

**HN `restrictSearchableAttributes=title` (hackernews.py:95):** This limits Algolia search to title-only matching, which misses "Ask HN" or "Show HN" threads where the topic appears in the body, not the title. For niche topics this could cut results significantly.

**HN `points>5` filter (hackernews.py:93):** Recent quality posts on niche topics often have <5 points in their first few hours. Since we're searching the last 30 days, this seems safe for older posts but could filter out recent ones.

**Please remove before merge:**
- `mise.toml` - dev environment tool choice shouldn't be in the repo
- `docs/plans/2026-03-11-refactor-query-relevance-consolidation-plan.md` - internal planning doc

**Suggestion:** Would you consider splitting this into two PRs?
1. Pure refactor: `query.py`, `relevance.py`, imports, tests (I'd merge this quickly)
2. Behavior changes: computed relevance, query optimizations, `--dateafter`, HN filters (needs more testing)

The refactor is ready. The behavior changes need validation that they don't degrade results for edge cases.

---
*Review assisted by [Claude Code](https://claude.com/claude-code)*

j-sperling:
Follow-up on additional changes since `036bcd2`:

**mise.toml:** Gitignored instead of deleted (`d8d2b97`) -- it stays useful locally but won't be tracked.

**Test coverage for behavior changes** (`cbee987`):
- Extracted `_relevance_filter` from the orchestrator closure to `score.relevance_filter()` so it's importable and testable. 5 tests cover threshold filtering, minimum-result guarantee (top 3 when all filtered), empty/small lists, and items missing the relevance attribute.
- Added Bluesky/TruthSocial engagement scoring tests (7 tests) and mixed-source `sort_items` tests (2 tests) -- these were untested despite being in production.
- Fixed `sort_items()` type hint to include `BlueskyItem`/`TruthSocialItem` (was missing despite `_ITEM_SOURCE_MAP` handling them).
- Added `select_openai_model` HTTP 401/403 error path tests (2 tests).

**On the split suggestion:** The risky behavior changes you flagged (`--dateafter`, `restrictSearchableAttributes`) are reverted. The remaining behavior changes -- computed relevance replacing hardcoded 0.7, and the relevance filter -- now have dedicated test coverage. If you'd still prefer a split I can do it, but the risk profile is substantially different from the original review.

609 tests passing across the full suite.

mvanhorn:
Hey @j-sperling - this is impressive work, especially the deduplication across 7 modules and the computed relevance scoring. The query type tiering in #67 is a smart design too.

Before I merge, I'd love to see some before/after results. Can you run 5 side-by-side comparisons - current main vs your branch (all 3 PRs merged) - on these topics from the README?

1. `nano banana pro prompting` (product)
2. `codex vs claude code` (comparison)
3. `anthropic odds` (prediction)
4. `kanye west` (breaking news / general)
5. `remotion animations for Claude Code` (how_to / niche)

Specifically I want to see:
- Which sources fired (and which got tiered out)
- How many results per source before/after the relevance < 0.3 filter
- Final item count per source
- Any results that were in main's output but got dropped

The scoring stderr output should cover most of this. Doesn't need to be pretty - just paste the raw runs and a quick summary of what changed.

Also - are you on X? Shoot me a DM @mvanhorn, would be good to chat.

j-sperling:
Follow-up on the low-overlap cases from the earlier side-by-side runs.

I dug into the remaining quality deltas and pushed two commits to `#65`:
- `946af84` `Tighten relevance scoring and Polymarket ranking`
- `8eda5fa` `Add local search quality evaluation harness`

What changed in the retrieval path:
- Reddit query expansion no longer contaminates broader queries like `anthropic odds` / `kanye west` with `worth it OR thoughts OR review` variants. Those opinion/review expansions are now limited to `product` and `opinion` query types.
- Shared relevance scoring no longer has the old artificial `0.1` floor. No-match results now score `0.0`, exact phrase matches still score very high, and generic-token-only matches (for example just `odds`) stay below the normal `0.3` filter threshold.
- Polymarket now treats semantic match as the primary signal. Volume/liquidity still matter, but they refine relevant markets instead of rescuing generic ones. I also apply the post-retrieval relevance filter to Polymarket now.

I added an optional local evaluator in `scripts/evaluate_search_quality.py` that compares `origin/main` vs a candidate checkout, computes deterministic stability metrics (Jaccard / retention per source), and can call Gemini for judged `Precision@5` / `nDCG@5` when `GEMINI_API_KEY` is configured.

One operational note: historical revisions like `origin/main` can still hit legacy Bird/browser-cookie auth paths if you run them naively. The eval harness now isolates child runs with a temp `HOME`, disables browser-cookie fallback, and strips `node` from child `PATH` so baseline comparisons stay on a popup-free path.

Validation:
- `uv run python -m unittest` -> 623 tests passing
- Deterministic 5-topic eval run completed locally with the hardened popup-free harness

The deterministic metrics are still only regression guards, not truth. Without `GEMINI_API_KEY` on this machine, the judged metrics stay at zero because there is no labeled pool yet. But the harness is now in place for that next step.

j-sperling:
Follow-up after the earlier comparison package. I added three more commits on this branch:

- `8c1dce9` Harden local search evaluation harness
- `c711e44` Reduce Reddit and Polymarket false positives
- `058c4e1` Classify prompt and animation queries earlier

What changed:

- The local eval harness now isolates child runs from local browser-cookie config, accepts this workspace's canonical `GOOGLE_API_KEY`, and computes `nDCG@5` against the judged union pool rather than normalizing against the run's own top-5 grades.
- Reddit relevance is now title-first, so long selftexts cannot overpower a weak title.
- Polymarket no longer expands low-signal standalone terms like `odds`/`pro`, and it no longer gets false positives from short binary outcomes. The concrete bug here was substring matching, e.g. outcome `No` was boosting `nano banana ...` because `no` is a substring of `nano`.
- Query typing now catches prompt-oriented and animation-oriented queries earlier, so tiering/tiebreakers match the benchmark topics better.

I reran the full 5-topic judged suite locally against `origin/main` after those fixes.

Macro judged metrics:

- `Precision@5`: `0.56 -> 0.68`
- `nDCG@5`: `0.41 -> 0.58`
- `source_coverage_recall`: `0.82 -> 0.95`

Per-topic judged metrics (`origin/main` -> current branch):

- `nano banana pro prompting`: `P@5 0.60 -> 0.40`, `nDCG@5 0.53 -> 0.34`, coverage `0.67 -> 1.00`
- `codex vs claude code`: `P@5 0.60 -> 1.00`, `nDCG@5 0.55 -> 1.00`, coverage `0.75 -> 0.75`
- `anthropic odds`: `P@5 0.40 -> 0.80`, `nDCG@5 0.21 -> 0.59`, coverage `0.67 -> 1.00`
- `kanye west`: `P@5 0.60 -> 0.60`, `nDCG@5 0.46 -> 0.62`, coverage `1.00 -> 1.00`
- `remotion animations for Claude Code`: `P@5 0.60 -> 0.60`, `nDCG@5 0.33 -> 0.35`, coverage `1.00 -> 1.00`

So the updated branch is now clearly ahead on the judged metrics overall, with the biggest gains on `anthropic odds` and `codex vs claude code`. The one query that is still weaker than `main` is `nano banana pro prompting`, so that is the remaining quality hole I would keep pushing on.

mvanhorn:
Can't wait to dig in! 3 am here though. You have x or another platform?

PR #74, opened 2026-03-20
refactor(store): extract _item_to_finding, fix TikTok data loss

phjlljp:
### Bug fix: TikTok and TruthSocial findings silently lost in `--store` mode

When a user runs `--store` to persist research findings to SQLite, the code at `last30days.py:1941-2021` iterates over each source's deduped items and builds a `findings` list for `store_mod.store_findings()`. There are 8 separate for-loops, one per source — but **TikTok and TruthSocial were never included**. Their deduped items (`deduped_tiktok`, `deduped_ts`) simply aren't iterated over, so they're silently dropped.

This means any user relying on `--store` for watchlist/briefing workflows (the variants/ use case) has been accumulating an incomplete dataset — TikTok and TruthSocial results appear in the CLI output but never make it to the database.

**Root cause:** The 8 for-loops were likely copy-pasted as new sources were added, and TikTok/TruthSocial were missed when they were introduced. The repetitive structure made it easy to overlook.

### Refactor: Extract `_item_to_finding()` dispatch function

To prevent this class of bug from recurring, this PR replaces the 8 for-loops with:
1. A single `_item_to_finding(item)` function that dispatches on `isinstance()` to map any schema item type to the flat dict format expected by the store module
2. A single loop: `findings = [_item_to_finding(item) for item in all_deduped]` where `all_deduped` explicitly concatenates all 10 source lists

Adding a new source now requires adding one `elif` branch to `_item_to_finding()` and one `list()` entry to `all_deduped` — rather than copy-pasting an entire for-loop block. The fallback `else` branch also handles unknown item types gracefully.

### Cleanup: `Engagement.to_dict()` simplified

Replaces 11 manual `if self.field is not None` checks with `{k: v for k, v in asdict(self).items() if v is not None}`. This means new engagement fields (like `shares`, `volume`, `liquidity` which were added after the original code) are automatically included in serialization without requiring a corresponding manual check. The `asdict()` call is safe here because `Engagement` is a flat dataclass with only scalar fields.

## Test plan
- [x] Existing test suite passes (461/465, 4 pre-existing model-name failures unrelated to this PR)
- [x] All 10 source types (including TikTok and TruthSocial) produce correct findings via `_item_to_finding()`
- [x] `Engagement.to_dict()` output is identical before/after for all field combinations (all-None, single field, mixed sources, zero values, all fields populated)
- [x] The `else` fallback in `_item_to_finding()` handles unexpected item types without crashing

### DIFF
--- a/scripts/last30days.py
+++ b/scripts/last30days.py
@@ -1938,87 +1938,13 @@ def main():
         topic_id = topic_row["id"]
         run_id = store_mod.record_run(topic_id, source_mode=mode, status="completed")
 
-        findings = []
-        for item in deduped_reddit:
-            findings.append({
-                "source": "reddit",
-                "url": item.url,
-                "title": item.title,
-                "author": item.subreddit,
-                "content": item.title,
-                "engagement_score": item.engagement.score if item.engagement else 0,
-                "relevance_score": item.relevance,
-            })
-        for item in deduped_x:
-            findings.append({
-                "source": "x",
-                "url": item.url,
-                "title": item.text[:100],
-                "author": item.author_handle,
-                "content": item.text,
-                "engagement_score": item.engagement.likes if item.engagement else 0,
-                "relevance_score": item.relevance,
-            })
-        for item in deduped_youtube:
-            findings.append({
-                "source": "youtube",
-                "url": item.url,
-                "title": item.title,
-                "author": item.channel_name,
-                "content": item.transcript_snippet[:500] if item.transcript_snippet else item.title,
-                "engagement_score": item.engagement.views if item.engagement and item.engagement.views else 0,
-                "relevance_score": item.relevance,
-            })
-        for item in deduped_hn:
-            findings.append({
-                "source": "hackernews",
-                "url": item.hn_url,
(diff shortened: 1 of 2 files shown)


mvanhorn:
See my reply on #76 covering all three of your PRs. Can't commit to anything right now with the v3.0 refactor underway, but I'll consider each one once it lands.

tmchow:
Thanks for catching the silent TikTok/TruthSocial drop in --store mode and for the refactor scope. Closing because the code structure you targeted (8 source-specific for-loops at `last30days.py:1941-2021` in --store mode) no longer exists on current main. `findings_from_report` was extracted to `skills/last30days/scripts/store.py` during the v3 refactor and now sources findings from `report.ranked_candidates` (rather than per-source raw lists), with a small Phase 2 for hackernews/polymarket items not captured in ranks. TikTok items flow through `ranked_candidates` naturally now, so the silent-drop bug shape doesn't reproduce. If you see TikTok findings still going missing on current main, please open a fresh issue with a sample run and I'll trace it.

PR #75, opened 2026-03-20
feat(dedupe): URL canonicalization for cross-source linking

phjlljp:
### Problem: Same article shared across sources isn't cross-linked

`cross_source_link()` detects when the same story is discussed on multiple platforms (e.g., Reddit + HN + Bluesky) and annotates items with `[also on: ...]` tags. Currently it relies entirely on text similarity (Jaccard on character trigrams and word tokens). This misses a common case: **the same URL shared across platforms with different tracking parameters**.

For example, these three URLs all point to the same article but would not be cross-linked today:
- Reddit: `https://www.example.com/article?utm_source=reddit&ref=share`
- HN: `https://example.com/article`  
- Bluesky: `https://www.example.com/article?fbclid=abc123`

The titles may also differ (Reddit editorializes, HN uses the original, Bluesky truncates), so text similarity alone doesn't reliably catch these.

### Solution: URL canonicalization as a fast-path check

This PR adds a `canonicalize_url()` function that normalizes URLs before comparison:
- Strips tracking/analytics query parameters (UTM, fbclid, gclid, msclkid, igshid, etc.)
- Removes `www.` prefix and lowercases the host
- Strips trailing slashes and URL fragments
- Rejects non-HTTP schemes (`file://`, `javascript:`, etc.)
- Sorts remaining query parameters for consistent comparison

`cross_source_link()` now pre-computes canonical URLs for all items (O(N), using stdlib `urlparse`) and checks URL identity first. If two items from different sources have the same canonical URL, they're cross-linked immediately without computing text similarity. This is both faster (string equality vs. Jaccard) and more accurate for URL-shared content.

The existing text similarity path remains as a fallback for cases where the same story is discussed but linked differently (e.g., a Reddit self-post discussing an HN thread).

### Design decisions

- **Conservative tracking param list:** Only strips params that are unambiguously tracking-related. Generic params like `source`, `context`, `s`, `t` are intentionally kept because they serve functional purposes on many sites (e.g., Reddit's `?context=3` for comment threading).
- **Pre-computed, not inline:** Canonical URLs are computed once per item before the O(N²) pairwise loop, not inside it.
- **Scheme restriction:** Non-HTTP URLs return `None` and are excluded from URL matching (but still participate in text similarity).

## Test plan
- [x] Existing test suite passes (461/465, 4 pre-existing)
- [x] URLs with UTM/fbclid params canonicalize to the same string as bare URLs
- [x] `?context=3` (Reddit) and `?source=rss` (functional) are preserved after canonicalization
- [x] `file://`, `javascript://`, and `ftp://` URLs return `None`
- [x] Cross-source linking correctly links Reddit, HN, and Bluesky items sharing the same article URL with different tracking params (verified bidirectional cross-refs)

### DIFF
--- a/scripts/lib/dedupe.py
+++ b/scripts/lib/dedupe.py
@@ -1,9 +1,80 @@
 """Near-duplicate detection for last30days skill."""
 
 import re
-from typing import List, Set, Tuple, Union
+from typing import List, Optional, Set, Tuple, Union
+from urllib.parse import urlparse, parse_qs, urlencode, urlunparse
 
 from . import schema
+
+
+# Tracking / analytics query parameters to strip during canonicalization
+_TRACKING_PARAMS = frozenset({
+    # Google Analytics / UTM
+    'utm_source', 'utm_medium', 'utm_campaign', 'utm_term', 'utm_content',
+    'utm_id', 'utm_source_platform', 'utm_creative_format',
+    # Facebook / Meta
+    'fbclid', 'fb_action_ids', 'fb_action_types', 'fb_ref', 'fb_source',
+    # Twitter / X
+    'ref_src', 'ref_url',
+    # Reddit
+    'share_id', 'ref_source',
+    # General trackers
+    'gclid', 'gclsrc', 'dclid', 'msclkid', 'mc_cid', 'mc_eid',
+    'igshid', '_ga', '_gl', 'yclid', 'twclid',
+    # Misc
+    'gi',
+})
+
+
+def canonicalize_url(url: Optional[str]) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks for the dedupe improvement — URL canonicalization for cross-source linking is a sensible direction. Closing because the function you targeted, `cross_source_link()` in `scripts/lib/dedupe.py`, no longer exists on current main. Cross-source linking moved through the v3 schema refactor (clusters now merge across sources via the cluster pipeline rather than per-item annotation), so the integration point you wrote against is gone. If you'd like to apply URL canonicalization at the new cluster-merge layer, a fresh PR on current main would be welcomed.

PR #76, opened 2026-03-20
feat(observability): SourceHealth tracking, logging and retry infra

phjlljp:
### Problem: No visibility into source performance

When a research run completes, the compact output shows what was found per source (e.g., "✅ Reddit: 12 threads") but gives no insight into **how long each source took**, **which backend was used**, or **why a source failed**. This makes it hard to:
- Debug slow runs (was it YouTube taking 90s, or Reddit?)
- Understand which backend was selected (Bird vs xAI vs ScrapeCreators for X)
- Diagnose intermittent failures (timeout vs API error vs auth failure)

### Solution: SourceHealth dataclass + timing instrumentation

**New `SourceHealth` dataclass** (`schema.py`) captures per-source metrics:
- `source`: name (e.g., "reddit", "x", "youtube")
- `status`: "ok", "error", "timeout", or "skipped"
- `item_count`: number of items returned
- `duration_ms`: wall-clock milliseconds spent waiting for this source
- `error`: error message (if any)
- `backend`: which backend was used (e.g., "bird", "scrapecreators", "yt-dlp", "algolia")

**Timing instrumentation** (`last30days.py`): Each source's `future.result()` call is wrapped with `time.monotonic()` to measure wall-clock wait time. A `SourceHealth` entry is appended to a `health` list after each source completes. The list is returned from `run_research()` and attached to the `Report`.

**Timing footer** (`render.py`): The compact output's source status section now includes a timing line, sorted slowest-first:
```
⏱️ Timing: Bluesky/atproto: T/O | YouTube/yt-dlp: 12.3s | Reddit/scrapecreators: 4.1s | HN/algolia: 1.2s | X/bird: ERR
```
Error and timeout sources show `ERR` or `T/O` instead of a time.

**Full serialization**: `SourceHealth` round-trips through `Report.to_dict()` / `Report.from_dict()`, using explicit `.get()` kwargs (not `**kwargs` splatting) for forward-compatibility with future fields.

### Infrastructure modules (not yet wired in)

This PR also adds two new modules that are imported and tested but not yet called from the main pipeline. They provide infrastructure for future improvements:

**`lib/log.py`**: Structured logging using Python's `logging` module with a bare formatter that matches the existing `[Source] message` stderr style. Provides `set_debug()` to flip between INFO and DEBUG levels, replacing the current pattern of `sys.stderr.write` + `if DEBUG` checks scattered across modules.

**`lib/retry.py`**: `@retry(max_attempts=2, backoff_base=1.5)` decorator for source-level retries with exponential backoff. Only retries transient failures (429, 503, connection errors) — permanent client errors (4xx) are raised immediately. Bounds-validated (max 5 attempts, max 10s backoff). Includes documentation about the interaction with `http.py`'s existing request-level retries to prevent surprise delay stacking.

### Note on timing semantics

The `duration_ms` measures time spent in `future.result()`, which is the **wall-clock impact on the user** — how long they waited for that source. Because futures are collected sequentially (not with `as_completed`), a source that finishes quickly but is collected after a slow source will show a low duration. This is intentional: it reflects blocking time, not execution time.

## Test plan
- [x] Existing test suite passes (461/465, 4 pre-existing)
- [x] Timing footer renders correctly with proper source display names, sorted slowest-first, showing ERR/T/O for failed sources
- [x] `SourceHealth` round-trips through `Report.to_dict()` / `Report.from_dict()` with all fields preserved (including error and backend)
- [x] `log.py` and `retry.py` import cleanly; `set_debug()` toggles log level between INFO and DEBUG
- [x] `@retry` raises immediately on 4xx errors (called once, no sleep)
- [x] `@retry` retries on 429 and 503 (called `max_attempts` times with backoff)
- [x] `@retry` recovers on transient-then-success (returns result from second attempt)

### DIFF
--- a/scripts/last30days.py
+++ b/scripts/last30days.py
@@ -23,6 +23,7 @@ import os
 import signal
 import sys
 import threading
+import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime, timezone
 from pathlib import Path
@@ -932,6 +933,7 @@ def run_research(
     polymarket_error = None
     web_error = None
     xiaohongshu_error = None
+    health: list = []  # List[schema.SourceHealth]
 
     # Determine web search mode
     do_web = sources in ("all", "web", "reddit-web", "x-web")
@@ -1013,7 +1015,7 @@ def run_research(
                     progress.show_error(f"Instagram error: {e}")
             if progress:
                 progress.end_instagram(len(instagram_items))
-        return reddit_items, x_items, youtube_items, tiktok_items, instagram_items, hackernews_items, bluesky_items, truthsocial_items, polymarket_items, web_items, web_needed, raw_openai, raw_xai, raw_reddit_enriched, reddit_error, x_error, youtube_error, tiktok_error, instagram_error, hackernews_error, bluesky_error, truthsocial_error, polymarket_error, web_error
+        return reddit_items, x_items, youtube_items, tiktok_items, instagram_items, hackernews_items, bluesky_items, truthsocial_items, polymarket_items, web_items, web_needed, raw_openai, raw_xai, raw_reddit_enriched, reddit_error, x_error, youtube_error, tiktok_error, instagram_error, hackernews_error, bluesky_error, truthsocial_error, polymarket_error, web_error, health
 
     # Determine which searches to run
     do_reddit = sources in ("both", "reddit", "all", "reddit-web")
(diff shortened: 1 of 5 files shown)


saudbishna55-dot (approved):

tmchow:
Thanks for the thoughtful observability proposal — the per-source latency / backend visibility you describe is a legitimate gap. Passing on this scope for now: introducing `scripts/lib/retry.py` as a new module plus a SourceHealth schema entry plus a logging refactor is a substantial new abstraction layer, and the current debug story (`--debug` stderr logs, plus `errors_by_source` for source-level failures) covers most of what users have asked for in the issue tracker. Adding a SourceHealth surface without a specific user-reported pain point would expand the API more than I want to take on right now. Not closed for any code defect — it's a scope call. Appreciate the work.

PR #83, opened 2026-03-25
chore(docs): use $HOME instead of hardcoded user paths in plans

chidev:
Plans and the v1/v2 test script referenced `/Users/mvanhorn/...`, which breaks copy-paste for other contributors. Switch those examples to `$HOME/...` for portability. No runtime behavior change.

Made with [Cursor](https://cursor.com)


tmchow:
Thanks for the portability sweep — closing because the 16 `docs/plans/*` files this targets are now in `.gitignore` (look at the top of `.gitignore`: `docs/plans/` is excluded as internal planning docs). Those files don't exist in the repo anymore, so the patch can't apply. The `/Users/mvanhorn/` paths in `scripts/test-v1-vs-v2.sh` got the same portability treatment yesterday in commit 378be7c — closed #297.

Issue #100, opened 2026-03-25
Reddit time out error.


fanispoulinakisai-boop:
Hi, I’m using the official open variant of last30days with ScrapeCreators + X cookies + Brave. 
On broad OpenClaw queries, normal runs often lose Reddit because Reddit times out at 90s. Using --deep helps and Reddit comes through. But I still see a Reddit formatting bug in the official output: some subreddit names show up like r/{'id': 't5_...', 'name': 'openclaw', ...} instead of a clean subreddit name. Example query: “best openclaw use cases”. Could you check the Reddit normalization/rendering path for subreddit search results?

Ok, a few morw qs (I'm not technical, so bear with me) 
1. Is there a way to generate a comprehensive document, such as a Google Doc/sheet, that contains a more detailed response with links for further review?

2. As I understand, the skill output/report uses its own internal model (gpt-5.1-codex-mini), and the final Telegram response is generated by whatever live chat model is currently configured on Openclaw. Is that correct?

mvanhorn:
Hey @fanispoulinakisai-boop - this bug report directly drove the timeout resilience work in v3. Reddit search now has parallel enrichment with per-item timeouts, so one slow thread never blocks or drops the rest of your results.

I've added you to [CONTRIBUTORS.md](https://github.com/mvanhorn/last30days-skill/blob/main/CONTRIBUTORS.md). If you want to show up in the GitHub contributor graph, submit a quick PR replacing the placeholder under your name with your bio or website.

Thanks for reporting this - it made v3 Reddit bulletproof.

PR #111, opened 2026-03-26
Normalize Reddit subreddit names before rendering

LittleChenLiya:
## Summary

This fixes a Reddit formatting bug where subreddit values can show up as object/dict strings in the rendered output instead of clean subreddit names.

For example, instead of rendering something like `r/openclaw`, the output could render `r/{'id': 't5_...', 'name': 'openclaw', ...}` when the upstream Reddit result returned a structured subreddit object.

## What changed

- Normalize Reddit subreddit values to a clean subreddit name before rendering
- Handle both string and dict/object-shaped subreddit values
- Apply the normalization across the Reddit search/parse/normalize path so downstream rendering and entity extraction stay consistent

## Why

Issue #100 mentions cases where subreddit names are rendered incorrectly in the output. This PR fixes that user-visible formatting bug directly.

### DIFF
--- a/scripts/lib/entity_extract.py
+++ b/scripts/lib/entity_extract.py
@@ -106,7 +106,12 @@ def _extract_subreddits(reddit_items: List[Dict[str, Any]]) -> List[str]:
 
     for item in reddit_items:
         # Primary subreddit
-        sub = item.get("subreddit", "").strip().lstrip("r/")
+        raw_sub = item.get("subreddit", "")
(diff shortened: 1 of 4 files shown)


LittleChenLiya:
Hi @mvanhorn,
Sorry to disturb you.
I was just wondering if you might have time to take a quick look at this PR when you're available.
No rush at all. Thank you very much!

Issue #121, opened 2026-03-28
Feature: Built-in proxy support for avoiding rate limits across platforms


Ultronprime2026:
## Problem

last30days hits 7+ platforms simultaneously (Reddit, X, YouTube, TikTok, Instagram, HN, Polymarket). At scale or with frequent usage, users inevitably hit IP-based rate limits and blocks — especially on X and TikTok.

## Suggestion

Add optional proxy support so users can route requests through residential proxies. This would:

- Prevent IP blocks during multi-platform scraping
- Enable geo-targeted research (e.g., see what people in specific countries are discussing)
- Allow more frequent/parallel runs without cooldown issues

## Existing Solution

[ProxyClaw](https://clawhub.ai/skills/proxyclaw) is an OpenClaw skill that provides residential proxy access (2M+ IPs, 195+ countries) via the IPLoop network. It works with Python (`pip install iploop-sdk`) and Node (`npm install iploop`).

A simple integration could be an optional `PROXY_URL` env var in `.env` config, or native support for the IPLoop SDK which handles rotation automatically.

## Example

```bash
# In ~/.config/last30days/.env
PROXY_URL=http://customer_id:api_key-country-US@proxy.iploop.io:8880
```

Would love to discuss integration options. Happy to help with a PR if interested.

mvanhorn:
Speaking as Matt's firstmate:
Classified **opt-in enhancement** (optional `PROXY_URL` / residential proxy). Owner already signaled interest post-v3 pipeline work (2026-03-28). No open PR in this batch. Parking as waiting-captain / backlog — not proposing close.

PR #122, opened 2026-03-28
docs: add optional proxy support for avoiding rate limits

Ultronprime2026:
Adds a "Proxy Support" section to the README with optional `PROXY_URL` env var configuration.

## Why

last30days hits 7+ platforms simultaneously. At scale, users hit IP-based rate limits — especially on X and TikTok. A residential proxy solves this cleanly.

## What

- Added proxy config section after the existing env setup
- References ProxyClaw / IPLoop as a recommended provider (2M+ IPs, 195+ countries)
- Fully optional — no code changes, just documentation

Related: #121

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,18 @@ If you're signed in to Codex (`codex login`), the skill will use your Codex cred
 
 For project-specific overrides, create `.claude/last30days.env` in the repo root. It overrides the global `~/.config/last30days/.env`.
 
+### Proxy Support (Optional)
+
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks for the suggestion. Closing — the README isn't the right surface to recommend a specific commercial proxy provider, and the recently-shipped #376 (LAST30DAYS_YT_SSH_HOST, residential-host routing for yt-dlp) already covers the most common datacenter-VPS bot-wall case with a free home-machine bypass instead of a paid service. For users who DO want to route through a residential proxy service, yt-dlp natively respects HTTP_PROXY / HTTPS_PROXY env vars, so the existing config surface already supports it without README-level vendor recommendations.

PR #112, opened 2026-03-26
fix: include youtube in freshness calculation (#108)

sjhddh:
## Summary

Fixes #108 by including YouTube items in the data freshness evaluation.

## Changes

- Added `yt_recent` to calculate recent YouTube videos
- Included `yt_recent` in `total_recent` sum
- Included `len(report.youtube)` in `total_items` sum

Fixes mvanhorn/last30days-skill#108

PR #125, opened 2026-03-29
fix: tighten recency filtering and mode display

ukgorclawbot-stack:
## Summary
- tighten YouTube recency handling so in-range videos are preferred and final report filtering drops out-of-window results
- make compact mode display include auxiliary sources like YouTube and HN when they contribute signal
- add display_mode to JSON output while keeping raw mode for internal/source-gating semantics

## Verification
- python3 -m unittest tests.test_youtube_yt tests.test_last30days_youtube_dates tests.test_schema_roundtrip tests.test_render tests.test_smoke.TestMockMode -v
- python3 ~/.agents/skills/last30days/scripts/last30days.py "AI video tools" --quick --emit=compact
- python3 ~/.agents/skills/last30days/scripts/last30days.py "AI video tools" --quick --emit=json

### DIFF
--- a/scripts/last30days.py
+++ b/scripts/last30days.py
@@ -1774,24 +1774,30 @@ def main():
     normalized_pm = normalize.normalize_polymarket_items(polymarket_items, from_date, to_date) if polymarket_items else []
     normalized_web = websearch.normalize_websearch_items(web_items, from_date, to_date) if web_items else []
 
-    # Hard date filter: exclude items with verified dates outside the range
-    # This is the safety net - even if prompts let old content through, this filters it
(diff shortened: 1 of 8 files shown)


tmchow:
Thanks for the work on YouTube recency. Closing because the YouTube + freshness machinery has been substantially reworked since you filed: `report.youtube` as a top-level schema field no longer exists (same shape that obsoleted #112), `display_mode` in JSON output has been handled differently by the v3 emit pipeline, and the source-gating semantics moved to `items_by_source` keying. So the patch can't apply cleanly and the underlying problem is no longer the same problem. If you're still seeing recency / mode display issues on current main, please open a fresh issue with a sample run and I'll trace it through the new code path.

Issue #110, opened 2026-03-26
Windows subprocess timeout cleanup still breaks in last30days.py and bird_x.py


LittleChenLiya:
I ran into this on Windows while debugging YouTube access locally.

I had already fixed one subprocess cleanup path in the YouTube code, but while reproducing that issue I noticed there are still other timeout/cleanup paths that use Unix-only logic like `os.killpg(os.getpgid(pid), signal.SIGTERM)`.

I found at least these remaining spots:

- `scripts/last30days.py`
- `scripts/lib/bird_x.py`

So this looks like the same cross-platform bug still exists in a few other subprocess paths.

## What I saw

On Windows, subprocess timeout cleanup can still break when these Unix-specific APIs are used directly.

## Suggested fix

It probably makes sense to use the same cross-platform cleanup pattern everywhere:
- use `killpg/getpgid` only when available
- otherwise fall back to `proc.kill()` (or another Windows-safe termination path)

That should make subprocess timeout/cleanup behavior more consistent across platforms.

dudupstmaxx:
Confirming this is still present on current `main` (and shipped in the v3.3.2 plugin release), now consolidated into the shared helper `skills/last30days/scripts/lib/subproc.py`.

The timeout cleanup path there is unguarded:

```python
# skills/last30days/scripts/lib/subproc.py, in run_with_timeout()
try:
    stdout, stderr = proc.communicate(timeout=timeout)
except subprocess.TimeoutExpired:
    try:
        os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
    except (ProcessLookupError, PermissionError, OSError):
        proc.kill()
    proc.wait(timeout=5)
    raise SubprocTimeout(...)
```

`preexec_fn` is correctly guarded with `hasattr(os, "setsid")`, but the kill path is not. On Windows `os.getpgid`/`os.killpg` don't exist, so on any subprocess timeout this raises `AttributeError`, which the `except (ProcessLookupError, PermissionError, OSError)` clause does **not** catch → the exception propagates and never falls back to `proc.kill()`.

### Real-world repro (Windows 10)
Running `last30days ... --emit=html`: a `digg-pp-cli posts` enrichment for one cluster exceeded its ~15s timeout. Instead of skipping just that cluster, the unhandled `AttributeError` from the timeout handler propagated and **took down the entire HTML render pipeline**.

### Minimal fix (same as suggested in #91)
```python
except subprocess.TimeoutExpired:
    try:
[...]

PR #127, opened 2026-03-30
fix: add AttributeError to subprocess cleanup (Windows compatibility)

ekkoitac:
## Fix Issue #110: Windows subprocess timeout cleanup still breaks

**Problem:** On Windows, `os.killpg` and `os.getpgid` are not available (or may not work correctly). When these Unix-only functions are called, they raise `AttributeError`, which was not being caught. This caused subprocess timeout cleanup to fail on Windows.

**Root Cause:** The existing YouTube fix added exception handling for `ProcessLookupError, PermissionError, OSError`, but did not include `AttributeError`. On Windows, calling `os.killpg` or `os.getpgid` raises `AttributeError` since these functions don't exist on Windows.

**Fix:** Add `AttributeError` to the exception tuple in all three remaining spots that still used the Unix-only pattern:
- `scripts/last30days.py` (`_cleanup_children` function)
- `scripts/lib/bird_x.py` (`_run_bird_search` function)
- `scripts/lib/bird_x.py` (`search_handles` function)

This matches the pattern already used in `scripts/lib/youtube_yt.py` and ensures the fallback `proc.kill()` is called on Windows instead of crashing.

### DIFF
--- a/scripts/last30days.py
+++ b/scripts/last30days.py
@@ -102,7 +102,7 @@ def _cleanup_children():
     for pid in pids:
         try:
             os.killpg(os.getpgid(pid), signal.SIGTERM)
-        except (ProcessLookupError, PermissionError, OSError):
+        except (ProcessLookupError, PermissionError, OSError, AttributeError):
             pass
 
 
--- a/scripts/lib/bird_x.py
+++ b/scripts/lib/bird_x.py
@@ -192,7 +192,7 @@ def _run_bird_search(query: str, count: int, timeout: int) -> Dict[str, Any]:
             # Kill the entire process group
             try:
                 os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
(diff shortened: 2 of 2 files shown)


tmchow:
Thanks for the careful Windows trace — the AttributeError on `os.killpg` was real. Closing because the subprocess invocation has since been refactored into a centralized helper at `skills/last30days/scripts/lib/subproc.py`, which guards with `os.setsid if hasattr(os, "setsid") else None` upfront and falls back to `proc.kill()` if `killpg` fails. Both bird_x.py and youtube_yt.py route through it, so the AttributeError can't reach this layer anymore. The remaining direct `os.killpg` call in `_cleanup_children()` got the same `hasattr` guard yesterday (commit b2a762a, closed #226). Net effect is the same fix you wrote, in two places.

PR #135, opened 2026-03-31
fix(schema): Engagement.to_dict() returns {} instead of None

### LINKED ISSUE #88: Bug: Engagement.to_dict() returns None instead of empty dict
## Description
In `scripts/lib/schema.py` (~line 52), `Engagement.to_dict()` returns `None` when all fields are `None`:

```python
return d if d else None
```

## Impact
Callers expecting a dict will get `AttributeError` on `.get()` calls. Can also produce unexpected `null` values in JSON serialization.

## Suggested Fix
```python
return d if d else {}
```

## Severity
Medium

zerone0x:
## Summary
Fixes #88

`Engagement.to_dict()` returned `None` when all fields were `None`, causing potential `AttributeError` on `.get()` calls and unexpected `null` values in JSON serialization. Now returns an empty dict `{}` instead.

## Changes
- `scripts/lib/schema.py` line 56: `return d if d else None` → `return d if d else {}`

---
🤖 Generated with Claude Code

### DIFF
--- a/scripts/lib/schema.py
+++ b/scripts/lib/schema.py
@@ -53,7 +53,7 @@ class Engagement:
             d['volume'] = self.volume
         if self.liquidity is not None:
             d['liquidity'] = self.liquidity
-        return d if d else None
+        return d if d else {}
 
 
 @dataclass


mvanhorn:
Closing alongside #136 - the v2 schema was replaced in the v3 rewrite. Thanks for the fix, @zerone0x.

PR #145, opened 2026-04-03
Add .codex-plugin/plugin.json manifest for Codex discoverability

Jah-yee:
This PR adds a Codex plugin manifest () to make the skill discoverable through Codex plugin search.

**Changes:**
- Added  with full skill metadata
- Includes keywords: research, reddit, twitter, youtube, hackernews, polymarket, bluesky, news, trends, deep-research, multi-source
- Matches the manifest format used by other Codex plugins

**Why this matters:**
The last30days-skill has 17K+ stars but isn't discoverable through Codex's plugin search path. Adding the manifest enables Codex users to find it via:

```
/plugin install last30days@last30days-skill
```

or through the Codex Plugin Directory.

Reference: #139

### DIFF
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,0 +1,54 @@
+{
+  "schema_version": "1.0.0",
+  "name": "last30days",
+  "version": "2.9.6",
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks — closing because we're intentionally NOT shipping `.codex-plugin/plugin.json` anymore. The 3.1.1 restructure moved every harness onto a single canonical layout under `skills/last30days/`, and the maintainer dropped the Codex-native plugin path explicitly (CHANGELOG marks it as a breaking change for Codex native-plugin users). Codex installs now use `npx skills add mvanhorn/last30days-skill` or a copy into `~/.codex/skills/last30days/`. So adding the manifest back would reverse a deliberate decision.

Issue #151, opened 2026-04-03
Two AttributeError crashes on fresh install (INCLUDE_SOURCES + show_complete)


Alan5168:
## 🐛 Bug Report

Fresh install with empty / unconfigured `.env` file causes two crashes:

---

### Bug 1: INCLUDE_SOURCES defaults to `None`

**File**: `scripts/lib/env.py` line 377

**Cause**: `INCLUDE_SOURCES` is loaded with default value of `None`:
```python
('INCLUDE_SOURCES', None),
```
When `.env` is empty, `config.get('INCLUDE_SOURCES', '')` returns `None` instead of `''`, and `None.split()` raises:
```
AttributeError: 'NoneType' object has no attribute 'split'
```

**Fix**: Change default to empty string:
```python
('INCLUDE_SOURCES', ''),
```

---

### Bug 2: `show_complete()` signature mismatch

**File**: `scripts/lib/ui.py` line 331 vs `scripts/last30days.py` line 2034

**Cause**: `show_complete()` accepts 7 positional args:
```python
def show_complete(self, reddit_count, x_count, youtube_count=0, hn_count=0,
                 pm_count=0, tiktok_count=0, ig_count=0):
```
But the call site passes 9:
```python
progress.show_complete(
    len(deduped_reddit), len(deduped_x), len(deduped_youtube),
    len(deduped_hn), len(deduped_pm), len(deduped_tiktok),
    len(deduped_ig), len(deduped_v2ex), len(deduped_hf))  # ← 2 extra
```

**Fix**: Add missing parameters:
```python
def show_complete(self, reddit_count, x_count, youtube_count=0, hn_count=0,
                 pm_count=0, tiktok_count=0, ig_count=0, v2ex_count=0, hf_count=0):
```

---

### Environment
- last30days v3.0
- `.env` file present but with empty keys (`OPENAI_API_KEY=`, `XAI_API_KEY=`)
- SETUP_COMPLETE not yet set

PR #148, opened 2026-04-03
feat: Add V2EX and HuggingFace as free sources (no API keys required)

Alan5168:
Resolves #149

## Summary

This PR adds two new **free sources** to last30days that require no API keys:

- **V2EX**: Chinese tech community hot topics (similar to Hacker News for Chinese developers)
- **HuggingFace**: Trending AI models with downloads, likes, and tags

## Motivation

See detailed rationale in **Issue #149**:

### Key Points:
1. **V2EX fills Chinese tech community gap** — Current sources are English-centric (Reddit, HN, X). V2EX covers Chinese developer discussions.
2. **HuggingFace provides AI model trend data** — Downloads and likes are real usage signals, not just discussion.
3. **Zero-config, free-first** — Both work immediately with no API keys, extending v2.9.6's philosophy.

## Changes

| File | Description |
|------|-------------|
| `scripts/lib/v2ex.py` | V2EX hot topics fetcher (public API) |
| `scripts/lib/huggingface.py` | HuggingFace models fetcher (public API) |
| `scripts/last30days_lite.py` | Standalone free-mode aggregator |
| `scripts/last30days.py` | Integration into main search pipeline |
| `docs/free-mode.md` | Documentation for new sources |

## Integration

Both sources are integrated into the main search pipeline:
- Added to `VALID_SEARCH_SOURCES`: `v2ex`, `huggingface`, `hf`
- Threaded search functions: `_search_v2ex()`, `_search_huggingface()`
- Normalized output format compatible with synthesis engine
- Relevance scoring: rank + engagement + topic overlap

## Usage

```bash
# Free mode standalone
python3 scripts/last30days_lite.py --topic "AI coding"

# Integrated with full search (--sources flag)
python3 scripts/last30days.py --sources v2ex,huggingface,hn --topic "Llama 4"
```

## Output Examples

**V2EX:**
```
## V2EX (10 条)
1. [尴尬了，可能招聘了一个不会手写代码的 AI 工程师]
   📍 career | 💬 100 | ⭐ relevance: 0.85
```

**HuggingFace:**
```
## HuggingFace (10 条)
1. [meta-llama/Llama-4-Maverick-17B-128E-Instruct]
   📥 1.2M downloads | ❤️ 3.5K likes | tags: llama, instruct
```

## Testing

Both modules tested locally:
- ✅ V2EX API returns 10-30 hot topics successfully
- ✅ HuggingFace API returns trending models with engagement data
- ✅ Topic filtering works (case-insensitive substring match)
[...]

PR #152, opened 2026-04-03
Fix: Two AttributeError crashes on fresh install (Issue #151)

Jah-yee:
## Summary

Fixes two bugs that cause crashes when installing on a fresh machine with empty `.env` file:

### Bug 1: INCLUDE_SOURCES defaults to None
- **File**: `scripts/lib/env.py` line 377
- **Problem**: `INCLUDE_SOURCES` defaults to `None` instead of empty string
- **Fix**: Changed default from `None` to `''`

### Bug 2: show_complete() signature mismatch
- **File**: `scripts/lib/ui.py` line 331
- **Problem**: Function accepts 7 args but is called with 9 (missing v2ex_count, hf_count)
- **Fix**: Added missing parameters

## Testing
Both issues verified by looking at the error traces in Issue #151.

### DIFF
--- a/scripts/lib/env.py
+++ b/scripts/lib/env.py
@@ -374,7 +374,7 @@ def get_config() -> Dict[str, Any]:
         ('TRUTHSOCIAL_TOKEN', None),
         ('FROM_BROWSER', None),
         ('SETUP_COMPLETE', None),
-        ('INCLUDE_SOURCES', None),
+        ('INCLUDE_SOURCES', ''),
     ]
 
     for key, default in keys:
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks for tracing both — closing because each part lands somewhere different on current main: **Bug 1** (INCLUDE_SOURCES None) is already fixed via `config.get('INCLUDE_SOURCES') or ''` at `skills/last30days/scripts/last30days.py:646` and `skills/last30days/scripts/lib/env.py:513` (same fix shape you wrote). **Bug 2** (show_complete signature mismatch with v2ex_count / hf_count) was caused by the V2EX + HuggingFace source-add PR #148, which I'm passing on for now — so the missing-args mismatch your trace shows doesn't reproduce against current main, since v2ex/huggingface aren't sources here. Closing as already-resolved / not-reproducible-on-main.

PR #153, opened 2026-04-03
Add Codex plugin manifest for discoverability

Jah-yee:
## Summary
Adds a Codex plugin manifest () to make this skill discoverable within OpenAI Codex CLI.

## Changes
- Added  with schema following OpenAI Codex plugin specification
- Includes skill name, version, description, author info, and trigger pattern

## Why
Issue #139 requested adding Codex discoverability. This makes the skill appear when users search for research/trending-related capabilities in Codex.

### DIFF
--- a/.codex-plugin/plugin.json
+++ b/.codex-plugin/plugin.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "https://openai.com/codex/plugin.schema.json",
+  "name": "last30days",
+  "version": "2.9.6",
+  "description": "Research any topic from the last 30 days across Reddit, X, YouTube, TikTok, Instagram, Hacker News, Polymarket, and the web.",
+  "author": {
+    "name": "Matt Van Horn",
+    "email": "mvanhorn@gmail.com",
+    "url": "https://github.com/mvanhorn"
+  },
+  "repository": "https://github.com/mvanhorn/last30days-skill",
+  "homepage": "https://github.com/mvanhorn/last30days-skill",
+  "license": "MIT",
+  "keywords": ["research", "reddit", "twitter", "youtube", "tiktok", "instagram", "trends", "prompts", "polymarket"],
+  "readme": "README.md",
+  "skills": [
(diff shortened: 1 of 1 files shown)


mvanhorn:
Thanks @Jah-yee, and appreciate the push to get Codex support shipped. Closing this one because I dug into the actual Codex CLI loader and found a couple of issues that need a different approach:

1. `https://openai.com/codex/plugin.schema.json` returns 404 - OpenAI doesn't publish that schema URL, so the `$schema` reference would bake in a broken link.

2. Codex's `.codex-plugin/plugin.json` is only a namespace marker. The real Rust reader at `codex-rs/utils/plugins/src/plugin_namespace.rs` only reads the `name` field - `version`, `description`, `author`, `repository`, `homepage`, `license`, `keywords`, `readme`, and the `skills[]` array are silently ignored.

3. Codex discovers skills by walking `.agents/skills/<name>/SKILL.md` from the project root. The PR didn't add a SKILL.md under that path, so the plugin.json had nothing to namespace. Also worth knowing: Codex's loader skips symlinked files (only follows symlinked directories), so a symlinked SKILL.md wouldn't work either - found that one out the hard way while dogfooding.

I verified all of this by running Codex CLI from a fresh checkout and having it trace its own loader logic. The real integration landed in #219 with `.agents/skills/last30days/SKILL.md` as a real file and `.codex-plugin/plugin.json` with just `{"name": "last30days"}`.

Serious thanks for the nudge on this one - without your PR sitting in the queue I wouldn't have gone looking. If you'd like credit in `CONTRIBUTORS.md`, feel free to submit a small PR adding yourself (a new "Post-v3 Contributors" section would be a fine home, or wherever you think fits). @pejmanjohn's #189 is a good template - tiny PR, just the entry and bio line. I'll merge it happily.

Re issue #139: that ask was about listing on hashgraph-online/awesome-codex-plugins (separate thing entirely), so I'll keep that open and handle it separately.

PR #155, opened 2026-04-03
fix: surface xAI API errors instead of silently returning empty results

km-git007:
### **What**
The xAI X (Twitter) search pipeline was silently discarding malformed or empty API responses. This caused users to receive incomplete research outputs without any error notification. 

When the xAI API returned a `200 OK` response with invalid or incomplete JSON (e.g., missing output text, malformed structure, or a missing `"items"` key), the parser would return an empty items list. Consequently, the orchestrator would incorrectly treat this as "zero relevant results found" rather than an "API failure."

### **Why**
The root cause was that `parse_x_response()` returned early on malformed responses without raising exceptions. This violated the error-propagation contract; callers could not distinguish between a successful search with no results and an API that returned garbage data. 

> [!NOTE]  
> This fix mirrors the approach taken in commit `7c9fd96` for the Codex authentication path (Reddit), ensuring consistency across our search parsers.

### **How**
The fix implements the following changes:

1.  **`xai_x.py`**: Modified `parse_x_response()` to explicitly raise an `HTTPError` when:
    * Output text is empty.
    * The JSON structure lacks the expected `"items"` key.
    * The JSON syntax itself is malformed.
2.  **`last30days.py`**: Wrapped the `parse_x_response()` call inside the existing `try-except` block. This ensures parsing errors propagate as an `x_error` to the UI, informing the user of the failure.

### **Testing**
- [x] **Unit tests**: Verified that valid responses parse correctly while malformed responses now successfully raise an `HTTPError`.
- [x] **Integration tests**: Confirmed that errors propagate through the orchestrator and set the `x_error` state correctly.
- [x] **Error Messages**: Added a response preview to the error output to assist with future debugging.

### **Notes for Reviewer**
* This is a direct port of the logic used in commit `7c9fd96` (Codex path fix) for architectural consistency.
* **Out of Scope**: The `Bird` and `ScrapeCreators` X search backends share this "parsing-outside-try-except" pattern; however, they remain untouched in this PR to keep the scope focused on xAI.
* **UI Impact**: Errors surface via the existing `progress.show_error()` channel. No new UI components were required.

### DIFF
--- a/scripts/last30days.py
+++ b/scripts/last30days.py
@@ -390,14 +390,15 @@ def _search_x(
             to_date,
             depth=depth,
         )
+        x_items = xai_x.parse_x_response(raw_response or {})
     except http.HTTPError as e:
         raw_response = {"error": str(e)}
         x_error = f"API error: {e}"
+        x_items = []
     except Exception as e:
         raw_response = {"error": str(e)}
         x_error = f"{type(e).__name__}: {e}"
-
-    x_items = xai_x.parse_x_response(raw_response or {})
+        x_items = []
 
     return x_items, raw_response, x_error
 
--- a/scripts/lib/openai_reddit.py
+++ b/scripts/lib/openai_reddit.py
@@ -98,6 +98,13 @@ def _parse_codex_stream(raw: str) -> Dict[str, Any]:
     """Parse SSE stream from Codex responses into a response-like dict."""
     events = _parse_sse_stream_raw(raw)
 
+    # Fail fast if the stream is completely empty (indicates API error, not zero results)
+    if not events:
+        raw_preview = raw[:100] if raw else "(empty)"
+        raise http.HTTPError(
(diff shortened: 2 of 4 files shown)


tmchow:
Thanks — the silent-empty-result on malformed xAI responses was a real failure mode and your trace of the parser paths was helpful. Closing because the file you patched (`scripts/lib/xai_x.py`) moved to `skills/last30days/scripts/lib/xai_x.py` in the 2026-04-23 restructure, so the patch needed a rebase. Rather than ask you to redo it, I re-applied the same `raise http.HTTPError` on each of the three empty-payload branches (empty output_text, no JSON match, JSON decode failure) directly on main as commit d51e91e with co-author credit to you. The raise propagates to `_retrieve_stream`'s caller, which catches it and surfaces it in `errors_by_source` — so users now see a visible failure signal instead of an empty X bucket. Appreciate the precise breakdown of all three failure modes.

PR #157, opened 2026-04-05
Fix Windows crash: guard os.killpg with hasattr check

### LINKED ISSUE #156: Windows crash: os.killpg not available (YouTube search fails)
## Bug

YouTube search crashes on Windows with:

```
AttributeError: module 'os' has no attribute 'killpg'
```

### Steps to reproduce

1. Run `last30days.py` on Windows (Python 3.12)
2. YouTube source is enabled (yt-dlp installed)
3. Any topic triggers the crash during search or transcript fetch

### Root cause

Two issues in `scripts/lib/youtube_yt.py`:

1. **`os.killpg()`** (lines 199, 321) — UNIX-only, does not exist on Windows. Called inside `except subprocess.TimeoutExpired` to kill the process group. On Windows this raises `AttributeError` *before* the `except (ProcessLookupError, PermissionError, OSError)` fallback can catch it.

2. **`os.setsid`** (lines 185, 307) — the `hasattr(os, 'setsid')` guard correctly sets `preexec_fn=None` on Windows, but `os.killpg` calls downstream still crash.

### Fix

- Guard `os.killpg` with `hasattr(os, 'killpg')`, falling back to `proc.kill()` on Windows
- Use `CREATE_NEW_PROCESS_GROUP` creationflags on Windows instead of `preexec_fn=os.setsid`

### Environment

- Windows 10/11, Python 3.12
- yt-dlp installed and working
- last30days v2.9.2

I have a fix ready and can submit a PR.

flobo3:
## Summary

Fixes #156 — YouTube search no longer crashes on Windows.

### Problem

`os.killpg()` is UNIX-only and raises `AttributeError` on Windows. Both the search and transcript fetch code paths call it unconditionally when a subprocess times out:

```python
# Line 173 and 396 — crashes on Windows
os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
```

Note: `preexec_fn=os.setsid` is already correctly guarded with `hasattr(os, 'setsid')`, but the downstream `os.killpg` calls were not.

### Fix

Guard both `os.killpg` calls with `hasattr(os, 'killpg')`, falling back to `proc.kill()` on Windows:

```python
if hasattr(os, 'killpg'):
    try:
        os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
    except (ProcessLookupError, PermissionError, OSError):
        proc.kill()
else:
    proc.kill()
```

### Testing

- Tested on Windows 11, Python 3.12 — YouTube search now works correctly
- No functional change on UNIX systems (same code path as before)

### DIFF
--- a/scripts/lib/youtube_yt.py
+++ b/scripts/lib/youtube_yt.py
@@ -169,9 +169,12 @@ def search_youtube(
         try:
             stdout, stderr = proc.communicate(timeout=120)
         except subprocess.TimeoutExpired:
-            try:
-                os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
-            except (ProcessLookupError, PermissionError, OSError):
+            if hasattr(os, 'killpg'):
+                try:
+                    os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
+                except (ProcessLookupError, PermissionError, OSError):
+                    proc.kill()
+            else:
                 proc.kill()
             proc.wait(timeout=5)
             _log("YouTube search timed out (120s)")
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks for catching this — the Windows `os.killpg` crash was a real issue. The subprocess invocation has since been refactored into a centralized helper at `skills/last30days/scripts/lib/subproc.py`, which both bird_x.py and youtube_yt.py now use. That helper already guards with `os.setsid if hasattr(os, "setsid") else None` and falls back to `proc.kill()` if `os.killpg` fails — same shape as your fix, applied at the right layer. Closing as already-fixed.

PR #160, opened 2026-04-05
Add YOUTUBE_PROXY config for yt-dlp (YouTube blocked regions)

### LINKED ISSUE #158: Feature: YOUTUBE_PROXY config for yt-dlp (YouTube blocked in some regions)
## Feature Request

yt-dlp supports a `--proxy` flag, but last30days doesn't expose any way to configure it. In regions where YouTube is blocked (e.g. Russia since August 2024), yt-dlp hangs or times out without a proxy.

### Current behavior

- `youtube_yt.py` builds yt-dlp commands without `--proxy`
- No config option exists to pass a proxy URL
- On networks where YouTube is unreachable, the YouTube source always times out (120s wasted per run)

### Proposed solution

Add a `YOUTUBE_PROXY` env var (or `YT_PROXY` / `yt_proxy` in `.env`) that gets passed as `--proxy` to all yt-dlp invocations in `scripts/lib/youtube_yt.py`.

```python
# In .env or environment
YOUTUBE_PROXY=socks5://127.0.0.1:1080
```

```python
# youtube_yt.py — append to cmd lists
proxy = os.environ.get("YOUTUBE_PROXY")
if proxy:
    cmd.extend(["--proxy", proxy])
```

Supports any proxy format yt-dlp accepts: `socks5://`, `http://`, `https://`.

### Scope

Affects two functions in `youtube_yt.py`:
- `search_youtube()` — search command
- `fetch_transcript()` — subtitle download command

Both already build a `cmd` list, so adding `--proxy` is a one-line addition to each.

### Alternative

yt-dlp also respects `HTTP_PROXY` / `HTTPS_PROXY` env vars natively, but those affect all HTTP traffic globally. A dedicated `YOUTUBE_PROXY` is cleaner — only routes yt-dlp through the proxy, not every web request the skill makes.

flobo3:
## Summary

Closes #158 — adds `YOUTUBE_PROXY` env var for routing yt-dlp through a proxy.

### Problem

In regions where YouTube is blocked (e.g. Russia since August 2024), yt-dlp hangs or times out (120s wasted per run). There was no way to configure a proxy.

### Solution

**`YOUTUBE_PROXY` env var** — read via `os.environ.get()` and passed as `--proxy` to both yt-dlp commands:

```bash
# In ~/.config/last30days/.env or environment
YOUTUBE_PROXY=socks5://127.0.0.1:1080
```

Supports any proxy format yt-dlp accepts: `socks5://`, `http://`, `https://`.

### Changes

**`scripts/lib/env.py`:**
- Added `YOUTUBE_PROXY` to the config key list so it's loaded from `.env` files

**`scripts/lib/youtube_yt.py`:**
- Added `_get_proxy_args()` helper that returns `["--proxy", url]` when set
- Applied to both `search_youtube()` and `fetch_transcript()` commands

### Testing

Tested on Windows 11, Python 3.12 with a local SOCKS5 proxy:
- YouTube search returns results successfully through the proxy
- Without `YOUTUBE_PROXY` set, behavior is unchanged (no proxy used)

### DIFF
--- a/scripts/lib/env.py
+++ b/scripts/lib/env.py
@@ -375,6 +375,7 @@ def get_config() -> Dict[str, Any]:
         ('FROM_BROWSER', None),
         ('SETUP_COMPLETE', None),
         ('INCLUDE_SOURCES', None),
+        ('YOUTUBE_PROXY', None),
     ]
 
     for key, default in keys:
--- a/scripts/lib/youtube_yt.py
+++ b/scripts/lib/youtube_yt.py
@@ -86,6 +86,14 @@ def extract_transcript_highlights(transcript: str, topic: str, limit: int = 5) -
     return [sent for _, sent in candidates[:limit]]
 
 
+def _get_proxy_args() -> list:
+    """Return yt-dlp proxy args if YOUTUBE_PROXY is configured."""
+    proxy = os.environ.get("YOUTUBE_PROXY")
+    if proxy:
+        return ["--proxy", proxy]
+    return []
+
+
 def _log(msg: str):
     """Log to stderr."""
     sys.stderr.write(f"[YouTube] {msg}\n")
@@ -154,7 +162,7 @@ def search_youtube(
         "--dump-json",
         "--no-warnings",
         "--no-download",
-    ]
+    ] + _get_proxy_args()
 
     preexec = os.setsid if hasattr(os, 'setsid') else None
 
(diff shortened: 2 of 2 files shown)


tmchow:
Thanks for taking the time on this and the clear scope. Closing as superseded — #376 shipped today (2026-05-17) and adds `LAST30DAYS_YT_SSH_HOST`, an opt-in SSH-routing path that runs yt-dlp on a residential host instead of the local VPS. That covers the most common bot-wall / geographic-block case (home machine bypass for datacenter installs) more cleanly. For the remaining niche your PR addresses (commercial residential proxy services, local SOCKS5 tunnels), yt-dlp natively respects `HTTP_PROXY` / `HTTPS_PROXY` env vars and they can be scoped to a single `last30days` invocation with `env HTTPS_PROXY=... last30days ...`. So holding off on adding a second dedicated knob. Appreciate the careful issue #158 too — it's what made the case for #376.

PR #161, opened 2026-04-05
feat: add Exa MCP as no-key web search backend

benzntech:
## Summary

- Adds `mcp__exa__web_search_exa` and `mcp__exa__crawling_exa` to `allowed-tools` in both `SKILL.md` and `variants/open/SKILL.md`
- Updates Step 2 to use Exa MCP **first** (semantic search, no API key required) and `WebSearch` second, running both in parallel for maximum web coverage
- Users without `EXA_API_KEY` now get free semantic web search via the MCP tool on every research run

## Motivation

The existing `exa_search.py` module requires an `EXA_API_KEY` (Tier 2). The `mcp__exa__*` tools are available in the Claude Code skill environment and work without any key. This change unlocks better web search coverage for zero-config users.

The existing HTTP API path (`exa_search.py`) is unchanged — users with `EXA_API_KEY` still benefit from it inside the Python subprocess.

## Test plan

- [ ] Run `/last30days <topic>` without `EXA_API_KEY` set — confirm Exa MCP fires in Step 2
- [ ] Run with `EXA_API_KEY` set — confirm both the subprocess API path and MCP path run
- [ ] Verify `--diagnose` output still reflects key-based web source availability correctly

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -3,7 +3,7 @@ name: last30days
 version: "2.9.6"
 description: "Deep research engine covering the last 30 days across 10+ sources - Reddit, X/Twitter, YouTube, TikTok, Instagram, Hacker News, Polymarket, and the web. AI synthesizes findings into grounded, cited reports."
 argument-hint: 'last30 AI video tools, last30 best project management tools'
-allowed-tools: Bash, Read, Write, AskUserQuestion, WebSearch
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks — closing in favor of your follow-up #269, which combines this Exa MCP work with the Firecrawl deep-fetch piece into a single integration. Will track the merge decision over there.

PR #163, opened 2026-04-05
feat: add Firecrawl MCP deep fetch as Step 2b

benzntech:
## Summary
- Add Firecrawl MCP to allowed-tools (scrape only)
- Add Step 2.6: deep fetch top web results with Firecrawl after WebSearch
- Security & Permissions: document Firecrawl credential handling
- Drop unused Exa tools (web_search_exa, crawling_exa) - only firecrawl_scrape is used in Step 2.6
- Exa integration deferred to future PR

## Test plan
- [ ] Run  with Firecrawl MCP configured
- [ ] Verify Step 2.6 fetches and truncates articles correctly
- [ ] Verify upstream v3.0.0 features still work

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -3,7 +3,7 @@ name: last30days
 version: "2.9.6"
 description: "Deep research engine covering the last 30 days across 10+ sources - Reddit, X/Twitter, YouTube, TikTok, Instagram, Hacker News, Polymarket, and the web. AI synthesizes findings into grounded, cited reports."
 argument-hint: 'last30 AI video tools, last30 best project management tools'
-allowed-tools: Bash, Read, Write, AskUserQuestion, WebSearch
+allowed-tools: Bash, Read, Write, AskUserQuestion, WebSearch, mcp__exa__web_search_exa, mcp__exa__crawling_exa, mcp__firecrawl__firecrawl_scrape, mcp__firecrawl__firecrawl_crawl
 homepage: https://github.com/mvanhorn/last30days-skill
 repository: https://github.com/mvanhorn/last30days-skill
 author: mvanhorn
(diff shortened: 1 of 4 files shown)


mvanhorn:
@benzntech let me know! curious

benzntech:
**@mvanhorn** — great question! Here's how it works:

The `mcp__firecrawl__*` tools use the Firecrawl MCP server, which bundles its own API credentials internally. From a **user perspective**, there's no `FIRECRAWL_API_KEY` to configure — they just use the tool and it works.

However, the MCP server itself still needs to be installed/configured. It's not a magic built-in — if the Firecrawl MCP server isn't present in Claude Code's environment, those tools won't be available.

So the accurate phrasing is: **"no user-facing API key required"** — rather than "no API key required" which implies zero setup.

Would you like me to update the PR description to clarify this?

mvanhorn:
Hmm okay might not work as well as the .env we have set up right now, but could work..

Could you walk me through use cases?  I signed up for an paid for firecrawl but still am not finding a regular use case.

benzntech:
Thanks for the review. Addressed items 1 and 2 in latest commit (de9eb42):

**1. Fixed - dropped unused tools**
Removed  and  from allowed-tools. Only  remains (used in Step 2.6). Same fix applied to .

**2. Fixed - renamed Step 2b to Step 2.6**
Reading order is now: STEP 2 → STEP 2.6. Monotone.

**3. variants/open/SKILL.md context**
This is a parallel skill for OpenClaw/Gemini/Codex environments that adds watchlist management, briefings, and history queries on top of the base research. Install path:  or similar variant path. It stays in sync by having its own allowed-tools declaration (now also cleaned up to remove unused Exa tools). Primary skill at  is the canonical one;  is for users who want the persistent knowledge features.

**4. PR description stale items**
The INCLUDE_SOURCES/TikTok-5-platforms claims were from an earlier branch. Current PR is scoped to Firecrawl deep fetch only. Exa integration deferred to future PR.

Ready for re-review when you have a chance.

tmchow:
Thanks — closing in favor of your follow-up #269, which combines this Firecrawl deep-fetch work with the Exa MCP piece into a single integration. Will track the merge decision over there.

Issue #166, opened 2026-04-06
Crash on startup: `config.get('INCLUDE_SOURCES', '').split` raises AttributeError when key is None


spikefcz:
## Summary

Fresh install of v2.9.6 on Windows crashes immediately on every run with `AttributeError: 'NoneType' object has no attribute 'split'`.

## Repro

1. `git clone https://github.com/mvanhorn/last30days-skill.git ~/.claude/skills/last30days`
2. `mkdir -p ~/.config/last30days && touch ~/.config/last30days/.env`  (no `INCLUDE_SOURCES` set)
3. `python scripts/last30days.py "anything" --sources reddit --quick`

## Traceback

```
Traceback (most recent call last):
  File ".../scripts/last30days.py", line 2137, in <module>
    main()
  File ".../scripts/last30days.py", line 1740, in main
    _include_sources = {s.strip().lower() for s in config.get('INCLUDE_SOURCES', '').split(',') if s.strip()}
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'split'
```

## Root cause

`scripts/lib/env.py:377` initializes the key with a `None` value:

```python
('INCLUDE_SOURCES', None),
```

So `config['INCLUDE_SOURCES']` exists and equals `None` — the `''` default in `config.get('INCLUDE_SOURCES', '')` never applies, and `None.split(',')` blows up.

Setting `INCLUDE_SOURCES=` in `.env` does **not** fix it (env loader still resolves to `None`).

## Suggested fix

`scripts/last30days.py:1740` — coerce to string before splitting:

```python
_include_sources = {s.strip().lower() for s in (config.get('INCLUDE_SOURCES') or '').split(',') if s.strip()}
```

Or change the default in `env.py:377` from `None` to `''`.

## Environment

- last30days-skill: v2.9.6 (latest main, commit fetched 2026-04-06)
- OS: Windows 11
- Python: 3.14.2
- Node: 22.17.0
- Shell: Git Bash

sdhilip200:
I checked this against current main and could not reproduce the crash anymore.

Evidence:
- skills/last30days/scripts/lib/env.py now defaults INCLUDE_SOURCES to an empty string.
- There is regression coverage in tests/test_env_include_sources_default.py.
- The relevant fix appears to be 23fc6c7 fix(env): default INCLUDE_SOURCES to empty string (#223).
[...]

PR #171, opened 2026-04-08
fix: update INCLUDE_SOURCES default to empty string for better handling

netpoe:
Changed the default value of INCLUDE_SOURCES in the configuration to an empty string. This allows callers to use .split(',') without issues when the key is present but unset, improving the robustness of the configuration handling.

https://github.com/mvanhorn/last30days-skill/issues/166


tmchow:
Thanks — this is resolved on main. `env.py` declares the default as `('INCLUDE_SOURCES', '')` (canonical line 267) and the call sites already use the `config.get('INCLUDE_SOURCES') or ''` idiom that handles both missing-and-set-to-None cases. Closing as already-fixed.

PR #173, opened 2026-04-08
last30days: lazy-load sweet-cookie so vendored Bird works on fresh installs

pejmanjohn:
## Summary
- lazy-load `@steipete/sweet-cookie` instead of importing it at module load time in the vendored Bird runtime
- keep `bird-search.mjs --check` working when `AUTH_TOKEN`/`CT0` are provided or browser cookie lookup is disabled
- add regression tests covering the fresh-install runtime paths

## Why
Fresh installs currently fail with `ERR_MODULE_NOT_FOUND` before Bird can resolve credentials from environment variables or respect `BIRD_DISABLE_BROWSER_COOKIES`.

The Python integration already disables browser-cookie lookup for the normal headless flow, so this follow-up makes the vendored Node runtime behave consistently with that intended path instead of crashing on startup when the optional browser-cookie dependency is absent.

## Verification
- `/Users/pejman/.oss-pr-manager/workspaces/mvanhorn__last30days-skill/.venv/bin/python -m pytest tests/test_bird_x.py -q`
- `AUTH_TOKEN=dummy CT0=dummy node scripts/lib/vendor/bird-search/bird-search.mjs --check`
- `BIRD_DISABLE_BROWSER_COOKIES=1 node scripts/lib/vendor/bird-search/bird-search.mjs --check`

### DIFF
--- a/scripts/lib/vendor/bird-search/lib/cookies.js
+++ b/scripts/lib/vendor/bird-search/lib/cookies.js
@@ -117,14 +117,21 @@ async function readTwitterCookiesFromBrowser(options) {
     }
     return { cookies: out, warnings };
 }
+async function extractCookiesFromBrowser(options) {
+    const { getCookies } = await loadSweetCookie();
+    return readTwitterCookiesFromBrowser({
+        getCookies,
+        ...options,
+    });
+}
 export async function extractCookiesFromSafari() {
-    return readTwitterCookiesFromBrowser({ source: 'safari' });
+    return extractCookiesFromBrowser({ source: 'safari' });
 }
 export async function extractCookiesFromChrome(profile) {
-    return readTwitterCookiesFromBrowser({ source: 'chrome', chromeProfile: profile });
+    return extractCookiesFromBrowser({ source: 'chrome', chromeProfile: profile });
 }
 export async function extractCookiesFromFirefox(profile) {
-    return readTwitterCookiesFromBrowser({ source: 'firefox', firefoxProfile: profile });
+    return extractCookiesFromBrowser({ source: 'firefox', firefoxProfile: profile });
 }
 /**
  * Resolve Twitter credentials from multiple sources.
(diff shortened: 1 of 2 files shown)


mvanhorn (changes requested):
Nice fix - the lazy-load approach is exactly right for fresh installs. One thing to address before merging:

The three convenience exports (`extractCookiesFromSafari`, `extractCookiesFromChrome`, `extractCookiesFromFirefox`) now crash because `readTwitterCookiesFromBrowser` destructures `getCookies` from `options` but the helpers don't pass it. They need the lazy-load too:

```js
export async function extractCookiesFromSafari() {
    const { getCookies } = await loadSweetCookie();
    return readTwitterCookiesFromBrowser({ getCookies, source: 'safari' });
}
```

Same pattern for Chrome and Firefox.

`resolveCredentials` (the only call site in `bird-search.mjs`) works correctly - this is just about keeping the full export surface working.

pejmanjohn:
Addressed in `b7d1a38`.

You were correct that the three convenience exports were still calling `readTwitterCookiesFromBrowser` without passing `getCookies`, so they could throw even though the `resolveCredentials` path was fixed.

What changed:
- added an `extractCookiesFromBrowser()` helper in `scripts/lib/vendor/bird-search/lib/cookies.js` that lazy-loads `@steipete/sweet-cookie` and passes `getCookies` through before calling `readTwitterCookiesFromBrowser`
- updated `extractCookiesFromSafari`, `extractCookiesFromChrome`, and `extractCookiesFromFirefox` to use that helper
- added a regression test in `tests/test_bird_x.py` that stubs `@steipete/sweet-cookie` locally and exercises all three helper exports directly

Verification run after the change:
- `workspaces/mvanhorn__last30days-skill/.venv/bin/python -m pytest tests/test_bird_x.py -q`
- `AUTH_TOKEN=dummy CT0=dummy node scripts/lib/vendor/bird-search/bird-search.mjs --check`
- `BIRD_DISABLE_BROWSER_COOKIES=1 node scripts/lib/vendor/bird-search/bird-search.mjs --check`

That keeps the original fresh-install `--check` fix in place and restores the helper export surface as well.

mvanhorn:
The lazy-load fix addressed exactly what was needed. Thanks for catching the convenience export gap I flagged and adding the extractCookiesFromBrowser wrapper - clean approach. The regression tests are solid too. Merged.

PR #175, opened 2026-04-09
docs: align README with v3 runtime

ziperlee:
This PR aligns the top-level README with the current v3 runtime and makes the v3 skill spec the source of truth for setup/runtime behavior.


mvanhorn:
Good catch on the SKILL.md URLs - those should point to the public repo. The README pointer to the v3 spec is helpful for contributors. Merged, thanks!

PR #182, opened 2026-04-09
fix: refresh expired bluesky sessions

ziperlee:
This refreshes the cached Bluesky session when search hits a 401, so long-running sessions can recover after token expiry instead of failing silently. It adds a regression test that verifies a stale token is retried once and the new session token is used.


mvanhorn:
Clean fix for the expired session issue. The retry-once pattern with cache invalidation on 401 is the right approach. The utf-8 briefing fix and URL corrections are good catches too. Merged, thanks @ziperlee.

PR #186, opened 2026-04-09
fix: rename skills/last30days-v3 directory so plugin install resolves correctly

pejmanjohn:
## Problem

`claude plugin update last30days@last30days-skill` fails with:

> Plugin "last30days" not found in marketplace "last30days-skill"

Uninstalling and reinstalling also fails with the same error. The only workaround is manual cache manipulation.

## Root cause

The Claude Code plugin installer resolves skills by **directory name**, not the `name` field in SKILL.md frontmatter. The skill directory is `skills/last30days-v3/` but the plugin name is `last30days`, so the installer can't find it.

Additionally, `.claude-plugin/marketplace.json` still lists the plugin as `last30days-3` (the pre-rename name), which is a second mismatch against `plugin.json` where the name is `last30days`.

## Changes

1. **`skills/last30days-v3/` → `skills/last30days/`** — directory name now matches the plugin name in SKILL.md and plugin.json
2. **`.claude-plugin/marketplace.json`** — plugin name `"last30days-3"` → `"last30days"` to match plugin.json
3. **`skills/last30days/SKILL.md`** — argument-hint updated from `last30days-v3` to `last30days`

The version is already tracked in SKILL.md frontmatter (`version: "3.0.0-alpha"`), so encoding it in the directory name was redundant.

## Verification

982 tests pass. 14 pre-existing failures (identical on main, unrelated to this change).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
   },
   "plugins": [
     {
-      "name": "last30days-3",
+      "name": "last30days",
       "description": "Research any topic across Reddit, X, YouTube, TikTok, Instagram, HN, Polymarket, GitHub, and 5+ more sources.",
       "version": "3.0.0-alpha",
       "author": {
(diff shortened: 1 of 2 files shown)


mvanhorn:
This conflicts with #194 (version bump) which just merged. Can you rebase against main? The rename is still needed - just needs to pick up the updated SKILL.md content.

pejmanjohn:
Rebased onto current `main` and pushed an update to this branch. The rename is still in place, and `skills/last30days/SKILL.md` now picks up the updated content from the merged version-bump work. Let me know if you want anything else adjusted @mvanhorn

mvanhorn:
Thanks for the quick rebase. This was the last piece of the v3 rename - plugin install should resolve correctly now. Merged.

PR #198, opened 2026-04-09
perf: optimize dedup, parallelize handle searches and enrichment

iliaal:
## Summary

Dedup was the biggest bottleneck. The inner loop called `normalize_text()` four times per comparison and recomputed `item_text()` on every kept item. A `_PreparedText` cache now pre-computes n-gram sets and token sets once per item, cutting dedup time by 6x.

Three I/O phases ran serially and didn't need to:
- Bird handle searches spawned one Node subprocess per handle. Now concurrent via `ThreadPoolExecutor`.
- YouTube comment enrichment looped one video at a time. Now parallel, matching how Reddit enrichment already worked.
- Retry-thin-sources ran one source at a time. Now threaded.

Clustering's O(n²) grouping loop and MMR selection recomputed `_candidate_text()` on every comparison. Both now pull from a pre-built `text_cache` dict and call `prepared_similarity()` instead of `hybrid_similarity()` with raw strings.

`_is_wsl()` is cached with `@lru_cache`. `Bundle.add_items()` uses `extend()` instead of `existing + items` list concatenation.

### Dedup micro-benchmark (worst case: all unique items)

| Input | Before | After | Speedup |
|-------|--------|-------|---------|
| 200 mixed | 0.054s | 0.009s | 6x |
| 300 unique | 2.162s | 0.387s | 5.6x |
| 500 unique | 4.899s | 0.793s | 6.2x |

### End-to-end (4-source query: Reddit, X, HN, Polymarket)

| Metric | Before | After | Speedup |
|--------|--------|-------|---------|
| Wall time | 5.23s | 3.73s | 29% |
| Research phase | 4.5s | 3.2s | 29% |

### What changed

| File | Change |
|------|--------|
| `dedupe.py` | `_PreparedText` caches n-grams + token sets; `dedupe_items()` pre-computes once per item instead of per-comparison |
| `bird_x.py` | `search_handles()` runs handle searches concurrently via `ThreadPoolExecutor` |
| `youtube_yt.py` | Comment enrichment loop replaced with `ThreadPoolExecutor` (matches Reddit pattern) |
| `pipeline.py` | Retry-thin-sources parallelized; list concat replaced with `extend()` |
| `cluster.py` | Pre-computed `text_cache` dict for greedy clustering and MMR; uses `prepared_similarity()` |
| `cookie_extract.py` | `_is_wsl()` cached with `@lru_cache` |
| `schema.py` | `Bundle.add_items()` uses `setdefault().extend()` instead of `get() + concatenation` |

## Test plan

- [x] 122 existing tests pass (2 pre-existing failures from Python version check, unrelated)
- [x] Dedup micro-benchmark: 6x speedup confirmed on 300-item worst case
- [x] End-to-end: 29% wall-time reduction on live 4-source query
- [ ] Verify Bird handle parallelization with 5+ handles (needs X auth with multiple resolved handles)
- [ ] Verify YouTube comment enrichment with ScrapeCreators key

### DIFF
--- a/scripts/lib/bird_x.py
+++ b/scripts/lib/bird_x.py
@@ -312,10 +312,9 @@ def search_handles(
     Returns:
         List of raw item dicts (same format as parse_bird_response output).
     """
-    all_items = []
     core_topic = _extract_core_subject(topic) if topic else None
 
-    for handle in handles:
+    def _search_one_handle(handle: str) -> List[Dict[str, Any]]:
         handle = handle.lstrip("@")
         if core_topic:
             query = f"from:{handle} {core_topic} since:{from_date}"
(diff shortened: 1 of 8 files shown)


mvanhorn:
Nice work on this. The PreparedText cache for dedup is the biggest win here - avoiding repeated normalize_text() calls in the inner loop is exactly the right fix. Parallelizing the handle searches and YouTube enrichment is a clean improvement too. Merged, thanks @iliaal.

PR #199, opened 2026-04-10
fix(reddit): use browser-like headers to fix HTTP 403 from urllib

francocarballar:
## Problem

`reddit_public.py` uses Python's `urllib` with a generic `User-Agent: last30days/3.0 (research tool)`. Reddit's servers block this with **HTTP 403** consistently, resulting in 0 Reddit results for every query.

`curl` with the same URL returns 200 — the issue is specific to `urllib`'s minimal request fingerprint, not Reddit's API availability.

## Root cause

Reddit detects and blocks non-browser HTTP clients based on the combination of `User-Agent` + missing browser headers (`Accept-Language`, `Accept-Encoding`, `Connection`).

## Fix

1. Replace the generic `USER_AGENT` with a standard browser User-Agent string.
2. Add `Accept-Language`, `Accept-Encoding`, and `Connection` headers to match a real browser request.
3. Add `gzip` decompression — Reddit now returns compressed responses when `Accept-Encoding: gzip` is sent, so the response body must be decompressed before JSON parsing.

## Verification

Tested locally: before the fix every request returned `[RedditPublic] 403 forbidden`. After the fix, queries return 9+ results as expected.

```
# Before
[RedditPublic] 403 forbidden: https://www.reddit.com/search.json?q=...

# After
✓ Research complete (13.8s) - Reddit: 9 threads
```

### DIFF
--- a/scripts/lib/reddit_public.py
+++ b/scripts/lib/reddit_public.py
@@ -11,6 +11,7 @@ Handles 429 rate limits with exponential backoff, HTML anti-bot responses,
 network timeouts, and missing subreddits.
 """
 
+import gzip
 import json
 import sys
 import time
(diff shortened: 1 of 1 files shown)


tmchow:
Real bug — Reddit's been progressively tightening anti-bot detection on `urllib`-shaped requests. Closing because `scripts/lib/reddit_public.py` moved during the 2026-04-23 restructure and the patch needed a rebase. Rather than ask you to redo it, I applied the same browser-User-Agent + Accept-Language / Accept-Encoding / Connection headers + gzip decompression directly on main as commit 4bae05e with co-author credit to you. Updated the existing user-agent assertion in `tests/test_reddit_public.py` to match. Thanks for the curl-vs-urllib diagnosis in the PR body — made the fix easy to validate.

PR #206, opened 2026-04-10
perf: batch store_findings, dedup source_items in O(1), remove dead code

iliaal:
### 1. N+1 queries in `store.store_findings()`

The old loop ran one SELECT per finding to check existence, then one INSERT or UPDATE. 100 findings cost 200 serial SQLite roundtrips.

Now: one batch SELECT with `WHERE source_url IN (...)` builds a lookup dict, then `executemany()` handles all inserts and updates. Query count stays constant regardless of batch size.

Benchmark on 500 findings: ~30ms to ~20ms. The gap widens on slower storage or larger batches.

### 2. O(n²) `source_items` dedup in `fusion.weighted_rrf()`

Merging an item into an existing candidate ran `any(existing.source == ... for existing in candidate.source_items)` on every merge, linearly scanning a list that grew with each merge. At 40 candidates with 20 source_items each, fusion went quadratic.

Now tracks `(source, item_id)` tuples in a per-candidate set for O(1) lookup. The `source_items` list itself is unchanged since other code iterates it.

### 3. Dead code removal

- `providers.py`: `GeminiClient.ground_search()` and `.url_context_json()` had zero call sites. Deleted.
- `render.py`: `_top_comment_excerpt()` had zero call sites. Deleted.
- `env.py`: `is_reddit_available()` was a one-line wrapper around `get_reddit_source()`. Callers can check `get_reddit_source(config) is not None` directly.

## Test plan

- [x] All 1022 existing tests pass
- [x] Pre-existing failures unchanged (test_store, test_resolve, test_setup_openclaw, test_watchlist_commands, test_generate_synthesis_inputs_v3, all unrelated to these changes)
- [x] Dead-code grep confirmed zero call sites across scripts/ and tests/
- [x] `store_findings` benchmark shows consistent improvement

### DIFF
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -372,14 +372,6 @@ def config_exists() -> bool:
     return False
 
 
-def is_reddit_available(config: dict[str, Any]) -> bool:
-    """Check if Reddit search is available.
-
-    v3 uses ScrapeCreators only.
-    """
-    return bool(config.get('SCRAPECREATORS_API_KEY'))
-
-
 def get_reddit_source(config: dict[str, Any]) -> str | None:
     """Determine which Reddit backend to use.
 
--- a/skills/last30days/scripts/lib/fusion.py
+++ b/skills/last30days/scripts/lib/fusion.py
@@ -116,6 +116,8 @@ def weighted_rrf(
     """Fuse ranked lists into a single candidate pool."""
     subqueries = {subquery.label: subquery for subquery in plan.subqueries}
     candidates: dict[str, schema.Candidate] = {}
+    # Track (source, item_id) pairs already attached to each candidate for O(1) dedup.
+    seen_source_items: dict[str, set[tuple[str, str]]] = {}
 
     for (label, source), items in streams.items():
         subquery = subqueries[label]
@@ -154,6 +156,7 @@ def weighted_rrf(
                         ]
                     },
                 )
+                seen_source_items[key] = {(item.source, item.item_id)}
                 continue
 
             candidate = candidates[key]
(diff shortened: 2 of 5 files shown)


mvanhorn:
@iliaal sorry about this - restructure under you in 72495c1 (shipped 3.1.1) moved the lib files. All three premises still hold on main:

- `store_findings` N+1: `skills/last30days/scripts/store.py:343`, the original per-finding SELECT/INSERT-or-UPDATE loop is intact.
- `weighted_rrf` O(n²) source-item dedup: `skills/last30days/scripts/lib/fusion.py:182`, the `any(existing.source == ... for existing in candidate.source_items)` is still there.
- Dead code: `is_reddit_available`, `GeminiClient.ground_search`, `GeminiClient.url_context_json`, `_top_comment_excerpt` all still have zero callers across `scripts/` and `tests/`. Confirmed via grep on current main.

One non-blocking note on the batch SELECT: SQLite's default `IN(...)` placeholder limit is 999 on older builds (32766 on 3.32+). Your benchmark covered 500 findings, well under. If a future batch could ever exceed 999 - say a topic with thousands of findings in one run - the `IN`-clause needs chunking. Leaving as-is is fine for current load; flagging so it doesn't get silently exceeded later.

Affected files all moved under `skills/last30days/scripts/`:
- `skills/last30days/scripts/store.py`
- `skills/last30days/scripts/lib/{env,fusion,providers,render}.py`

(Note: `render.py` was deleted then re-added at the new path during restructure, so the rebase will need to re-apply your dead-code removal at the new file. Same diff, new path.)

iliaal:
@mvanhorn clean to merge now as well

PR #207, opened 2026-04-10
refactor(reddit): migrate to http.get(params=...) helper

iliaal:
### 1. `http.get(params=...)` support

Added a `params` kwarg to `http.request()` and `http.get()` that urlencodes a dict into the query string. None values drop, ints and bools stringify, and params append correctly when the URL already has a query string. Six new tests cover the edge cases.

### 2. `reddit.py` migrated to the helper

The three ScrapeCreators call sites (`_global_search`, `_subreddit_search`, `fetch_post_comments`) each had a `try: import requests as _requests / except ImportError` block plus paired `if not _requests: ... else: ...` branches doing the same work twice, once via stdlib urllib and once via the `requests` library.

Deleted the whole fallback pattern. All three functions now call `http.get(url, headers=..., params={...})` directly. The existing `http.py` urllib implementation already has retry logic, 429 handling, and `HTTPError` types that are strictly better than the ad-hoc `requests` branches.

## Diff summary

- `scripts/lib/http.py`: +9 lines (params kwarg + docs)
- `scripts/lib/reddit.py`: +12 / -82 lines (net -70, from ~490 to ~420 lines)
- `tests/test_http_v3.py`: +63 lines (6 new params tests)

## Test plan

- [x] 8 http tests pass (2 existing + 6 new)
- [x] 99 reddit tests pass
- [x] Full suite: 1028 passed, 15 pre-existing failures (unchanged)
- [x] Live smoke test: real ScrapeCreators search returned 12 threads with full engagement data

## Follow-ups (separate PRs)

The same `_requests` fallback pattern lives in five more files (`instagram.py`, `pinterest.py`, `threads.py`, `tiktok.py`, `youtube_yt.py`). Each will get its own PR once this one lands.

### DIFF
--- a/scripts/lib/http.py
+++ b/scripts/lib/http.py
@@ -38,6 +38,7 @@ def request(
     url: str,
     headers: Optional[Dict[str, str]] = None,
     json_data: Optional[Dict[str, Any]] = None,
+    params: Optional[Dict[str, Any]] = None,
     timeout: int = DEFAULT_TIMEOUT,
     retries: int = MAX_RETRIES,
     max_429_retries: int = MAX_429_RETRIES,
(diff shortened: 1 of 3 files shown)


mvanhorn:
Merged, thanks @iliaal. Nice clean refactor - the params= kwarg on http.request() handles None filtering and existing-query-string append correctly, and dropping reddit.py's dual requests/urllib fallback simplifies the error paths. The retries=2 consistency is a bonus fix. Looking forward to the rest of the chain.

PR #208, opened 2026-04-10
fix(github): reject garbage in _parse_date; consolidate date parsing

iliaal:
### 1. github.py silently accepted garbage as "dates"

The old `_parse_date` did a naive string slice:

```python
def _parse_date(iso_str):
    if not iso_str:
        return None
    try:
        return iso_str[:10]
    except (IndexError, TypeError):
        return None
```

Any 10+ character string became a "date." `"hello world"` returned `"hello worl"`, `"2026-99-99"` returned `"2026-99-99"`, `"25 chars of pure junk"` returned `"25 chars o"`.

Now delegates to `dates.parse_date()`, which validates the format and returns `None` for non-date input. GitHub's API always emits real ISO 8601, so this worked in practice, but it's a latent bug waiting for the first unexpected field.

Added 4 new tests: garbage rejection, invalid date values like `2026-99-99`, and both `+00:00` and `+0000` offset variants.

### 2. reddit.py and threads.py consolidated to dates.parse_date()

Both files hand-rolled ISO-offset handling: the `.replace("Z", "+00:00")` dance plus `if clean.endswith("+0000"): clean = clean[:-5] + "+00:00"`. The shared `dates.parse_date()` already handles all of this, including the `+0000` no-colon variant Reddit emits.

I deleted the hand-rolled parsers and called `dates.parse_date()` instead. The `from datetime import datetime, timezone` imports became dead and went with them.

Preserved one original behavior: reddit.py's `if not value: return None` early-return. Without it, a Unix timestamp of `0` parses as `"1970-01-01"` instead of None, which breaks an existing test.

## Diff summary

- `scripts/lib/github.py`: +9 / -8 (delegate to dates.parse_date)
- `scripts/lib/reddit.py`: +9 / -19 (delete hand-rolled parser + dead imports)
- `scripts/lib/threads.py`: +9 / -24 (delete hand-rolled parser + dead imports)
- `tests/test_github.py`: +16 lines (4 new tests covering the fix)

## Test plan

- [x] 119 tests pass across reddit and github modules
- [x] Full suite: 1026 passed, 15 pre-existing failures (unchanged)
- [x] New tests verify garbage rejection: `"hello world"`, `"2026-99-99"`, `"abcdefghij"` all return None
- [x] New tests verify offset handling: both `+00:00` and `+0000` parse correctly

### DIFF
--- a/scripts/lib/github.py
+++ b/scripts/lib/github.py
@@ -17,7 +17,7 @@ import urllib.request
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any, Dict, List, Optional
 
-from . import log
+from . import dates, log
 from .query import extract_core_subject
 from .relevance import token_overlap_relevance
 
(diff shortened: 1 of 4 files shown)


mvanhorn:
Merged, thanks @iliaal. The naive [:10] slicing returning 'hello worl' for 'hello world' was a real bug - nice catch. The garbage-rejection tests capture exactly what the old code got wrong, and the dates.parse_date delegation cleanly removes the +0000 normalization hack in reddit.py.

PR #209, opened 2026-04-10
refactor: consolidate _sc_headers into http.scrapecreators_headers

iliaal:
## Summary

Six source modules each defined the exact same 8-line function:

```python
def _sc_headers(token: str) -> Dict[str, str]:
    """Build ScrapeCreators request headers."""
    return {
        "x-api-key": token,
        "Content-Type": "application/json",
    }
```

Zero per-source variation, zero reason to have six copies. I moved it to `http.scrapecreators_headers()` and migrated all 33 call sites.

## Diff summary

- `scripts/lib/http.py`: +8 lines (new `scrapecreators_headers()` helper)
- `scripts/lib/reddit.py`: +0 / -8 (delete local, use helper)
- `scripts/lib/threads.py`: +0 / -8
- `scripts/lib/tiktok.py`: +0 / -8
- `scripts/lib/instagram.py`: +0 / -8
- `scripts/lib/pinterest.py`: +0 / -8
- `scripts/lib/youtube_yt.py`: +0 / -8

Net: **-40 lines**.

## Test plan

- [x] Full suite: 1022 passed, 15 pre-existing failures (unchanged)
- [x] All six migrated modules import cleanly
- [x] Live smoke test: real reddit search returned 12 threads with full engagement data
- [x] No behavior change. The function body is byte-identical to what each file had.

### DIFF
--- a/scripts/lib/http.py
+++ b/scripts/lib/http.py
@@ -166,6 +166,14 @@ def post_raw(url: str, json_data: Dict[str, Any], headers: Optional[Dict[str, st
     return request("POST", url, headers=headers, json_data=json_data, raw=True, **kwargs)
 
 
+def scrapecreators_headers(token: str) -> Dict[str, str]:
+    """Build ScrapeCreators request headers (x-api-key + JSON content type)."""
+    return {
+        "x-api-key": token,
+        "Content-Type": "application/json",
+    }
+
+
 def get_reddit_json(path: str, timeout: int = DEFAULT_TIMEOUT, retries: int = MAX_RETRIES) -> Dict[str, Any]:
     """Fetch Reddit thread JSON.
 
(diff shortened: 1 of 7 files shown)


mvanhorn:
Thanks @iliaal - the consolidation direction is exactly right. Quick heads up: I just merged #207 (reddit http.get migration) and #208 (date parsing consolidation), which touched some of the same hunks this PR expects. #209 is now in conflict on reddit.py and threads.py.

Could you rebase onto current main? The core intent (single `http.scrapecreators_headers` helper, remove the 6 duplicates) should apply cleanly on top - it's mostly just the surrounding context that shifted.

No rush. Happy to merge once it's rebased.

iliaal:
> Could you rebase onto current main? The core intent (single `http.scrapecreators_headers` helper, remove the 6 duplicates) should apply cleanly on top - it's mostly just the surrounding context that shifted.

Rebased should merge cleanly now

PR #211, opened 2026-04-10
Codex/add openclaw variant files

fanispoulinakisai-boop:
Description
This adds the missing OpenClaw variant files under variants/open that the repo already references in README and scripts/sync.sh.

Why:

README tells OpenClaw users to install last30days-official
scripts/sync.sh already tries to sync variants/open into ~/.openclaw/skills/last30days
main is missing those files, so the OpenClaw sync path is incomplete
What’s included:

variants/open/SKILL.md
variants/open/context.md
variants/open/references/briefing.md
variants/open/references/history.md
variants/open/references/research.md
variants/open/references/watchlist.md
What I tested:

cloned the public repo
reproduced that scripts/sync.sh expected variants/open
added these files on a branch
re-ran scripts/sync.sh in a temp HOME and it completed successfully

### DIFF
--- a/variants/open/SKILL.md
+++ b/variants/open/SKILL.md
@@ -1,0 +1,81 @@
+---
+name: last30days
+version: "3.0.0-open"
+description: "Research any topic across Reddit, X, YouTube, TikTok, Instagram, Hacker News, Polymarket, GitHub, Perplexity, and more. AI agent scores by upvotes, likes, and real money. Also triggered by 'last30'."
+argument-hint: 'last30 AI video tools, last30 watch my competitor every week, last30 give me my briefing'
+allowed-tools: Bash, Read, Write, AskUserQuestion, WebSearch
+disable-model-invocation: true
+---
+
+# last30days (open variant): Research + Watchlist + Briefings
+
+Multi-mode research skill with persistent knowledge accumulation.
+
+## Command Routing
+
+Parse the user's first argument to determine the mode:
+
+| First word | Mode | Reference |
+|---|---|---|
+| `watch` | Watchlist management | `references/watchlist.md` |
+| `briefing` | Morning briefing | `references/briefing.md` |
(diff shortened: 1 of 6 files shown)


mvanhorn:
Thanks @fanispoulinakisai-boop, and also for originally flagging #100 which drove the v3 Reddit timeout resilience work - CONTRIBUTORS.md credits you for that in the v3 Inspiration section.

Going to close this one though, for a couple of reasons:

1. `variants/open/` was intentionally removed from the public repo in v3 (commit 0a9ff16). The OpenClaw variant now lives in my private repo (`last30days-skill-private`) where a separate publish workflow handles it. A cookie-stripping script removes browser-credential code before publishing to ClawhHub. The reason for the split is that ClawhHub's Gemini/VirusTotal scanner reads every shipped file and flags inconsistencies between the SKILL.md security section and actual code capabilities. Keeping the OpenClaw variant in the public repo would regress that cleanup.

2. The content in this PR is v2-era, not v3. The SKILL.md says Reddit requires `OPENAI_API_KEY`, but in v3 Reddit switched to `SCRAPECREATORS_API_KEY` (one key for Reddit, TikTok, Instagram). It also uses the old `watch`/`briefing`/`history` mode router from v2.1, which v3 restructured. Looks like Codex regenerated the variant by looking at older state in git history or changelog.

You DID spot a real internal inconsistency though: `scripts/sync.sh` references `variants/open/` and the README points OpenClaw users at `clawhub install last30days-official`, but the path doesn't exist on the public repo's main. The correct fix is to make the sync.sh OpenClaw branch conditional on `variants/open/` existing in the source tree. I'll open a follow-up PR for that right after this.

Thanks for the careful read and for surfacing this. Also just cut v3.0.0 today (#220, #219, #214) - you're part of the story in the release notes via CONTRIBUTORS.md.

PR #223, opened 2026-04-11
fix(env): default INCLUDE_SOURCES to empty string

hnshah:
## Summary

Fixes a fresh-install crash where `INCLUDE_SOURCES` can be `None`, causing `.split(',')` call sites to raise `AttributeError`.

## Motivation

Issue #166 reports a Windows crash on startup when `INCLUDE_SOURCES` is present in config as `None`.

## Implementation

- Change `scripts/lib/env.py` default for `INCLUDE_SOURCES` from `None` to an empty string.
- Add a unit test asserting `INCLUDE_SOURCES` defaults to `""` when unset.

## Test

- `pytest -q tests/test_env_include_sources_default.py`

## Notes

This is a defensive default that prevents future regressions in any call site that assumes `INCLUDE_SOURCES` is a string.

### DIFF
--- a/scripts/lib/env.py
+++ b/scripts/lib/env.py
@@ -264,7 +264,7 @@ def get_config() -> dict[str, Any]:
         ('XQUIK_API_KEY', None),
         ('FROM_BROWSER', None),
         ('SETUP_COMPLETE', None),
-        ('INCLUDE_SOURCES', None),
+        ('INCLUDE_SOURCES', ''),
     ]
 
     for key, default in keys:
--- a/tests/test_env_include_sources_default.py
+++ b/tests/test_env_include_sources_default.py
@@ -1,16 +1,14 @@
-import os
-
-from scripts.lib.env import get_config
+from scripts.lib import env
 
 
 def test_include_sources_defaults_to_empty_string(monkeypatch, tmp_path):
     # Ensure the env var is not set
     monkeypatch.delenv("INCLUDE_SOURCES", raising=False)
 
-    # Avoid reading any real user config file
-    monkeypatch.setenv("LAST30DAYS_CONFIG_FILE", str(tmp_path / "does-not-exist.env"))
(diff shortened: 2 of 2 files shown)


mvanhorn:
Thanks @hnshah - the env.py change is a good defensive fix. Flagging a test issue before merge.

The test monkeypatches `LAST30DAYS_CONFIG_FILE`, but the actual env var in `env.py:18` is `LAST30DAYS_CONFIG_DIR`. Also, `_config_override = os.environ.get('LAST30DAYS_CONFIG_DIR')` is captured at module import time, so runtime monkeypatching has no effect. I ran the test locally and it fails whenever the developer has `INCLUDE_SOURCES` set in their real `~/.config/last30days/.env`.

Could you either:
1. Set `LAST30DAYS_CONFIG_DIR` to an empty dir before import (tricky given module-level capture),
2. Directly assert the `keys` tuple contains `('INCLUDE_SOURCES', '')` without going through `get_config()`, or
3. Mock `env.CONFIG_FILE` to a tmp path inside the test?

Option 3 is probably the cleanest. Happy to merge once CI-green across environments.

On the issue link: the specific crash at `last30days.py:1740` from issues #166/#151 no longer exists in current main - the consumer was refactored to use `config.get('INCLUDE_SOURCES') or ''`. Your change is still valuable as a defense-in-depth measure.

hnshah:
You're right. The original test was flawed because it used the wrong env var and relied on runtime monkeypatching that doesn't affect the module-level config capture.

I updated the test to patch `env.CONFIG_FILE` directly to a tmp path instead, which keeps it deterministic across environments while preserving the production fix.

Verified locally with:

```bash
python3 -m pytest tests/test_env_include_sources_default.py -q
```

Thanks for the precise catch.

mvanhorn:
Thanks for the quick turnaround on the test fix - patching `env.CONFIG_FILE` directly is exactly the right approach, and it now passes deterministically across environments. Merging.

PR #225, opened 2026-04-11
fix(cli): Write saved output using UTF-8 encoding

Gujiassh:
Summary:\n- save generated output using explicit UTF-8 encoding to avoid Windows locale write errors\n- add regression test for save_output UTF-8 write behavior\n\nValidation:\n- python3 -m pytest tests/test_cli_v3.py -q\n\nFixes #167


mvanhorn:
Merged, thanks @Gujiassh.

PR #226, opened 2026-04-11
fix(cli): keep child cleanup working on Windows

Gujiassh:
## Summary
- make `_cleanup_children()` in `scripts/last30days.py` use process-group termination only when `os.killpg`/`os.getpgid` are available
- add a safe fallback to `os.kill(pid, SIGTERM)` for platforms (notably Windows) without process-group APIs
- add targeted regression tests for both process-group and fallback code paths

## Testing
- `pytest -q tests/test_last30days_cleanup.py`

Refs #110.

### DIFF
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -62,7 +62,12 @@ def _cleanup_children() -> None:
         pids = list(_child_pids)
     for pid in pids:
         try:
-            os.killpg(os.getpgid(pid), signal.SIGTERM)
+            killpg = getattr(os, "killpg", None)
+            getpgid = getattr(os, "getpgid", None)
+            if callable(killpg) and callable(getpgid):
+                killpg(getpgid(pid), signal.SIGTERM)
+            else:
+                os.kill(pid, signal.SIGTERM)
         except (ProcessLookupError, PermissionError, OSError):
             continue
 
--- a/tests/test_last30days_cleanup.py
+++ b/tests/test_last30days_cleanup.py
@@ -1,0 +1,50 @@
+import sys
+import unittest
+from pathlib import Path
+from unittest import mock
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "skills" / "last30days" / "scripts"))
+
+import last30days
+
+
+class TestCleanupChildren(unittest.TestCase):
+    def setUp(self) -> None:
+        with last30days._child_pids_lock:
+            last30days._child_pids.clear()
+
+    def tearDown(self) -> None:
+        with last30days._child_pids_lock:
+            last30days._child_pids.clear()
+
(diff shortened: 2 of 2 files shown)


mvanhorn:
Thanks @Gujiassh - the atexit fix and tests are clean. One thing before merge: this same `os.killpg(os.getpgid(...))` pattern appears at 4 other sites that will still crash on Windows:

- `scripts/lib/bird_x.py:196` (Bird X timeout)
- `scripts/lib/bird_x.py:347` (handle search timeout)
- `scripts/lib/youtube_yt.py:244` (YouTube search timeout)
- `scripts/lib/youtube_yt.py:469` (YouTube transcript timeout)

Each of those catches `(ProcessLookupError, PermissionError, OSError)` but on Windows `os.killpg` raises `AttributeError`, so the crash still propagates. Issue #156 is actually about `youtube_yt.py:244`, not the atexit path.

Could you extend the same guard pattern to those four sites? Happy to merge once they're covered.

Gujiassh:
Extended the Windows-safe timeout cleanup to the remaining four call sites mentioned in review.

Updated in `24373ff`:
- `scripts/lib/bird_x.py:196`
- `scripts/lib/bird_x.py:347`
- `scripts/lib/youtube_yt.py:244`
- `scripts/lib/youtube_yt.py:469`

Validation:
- `python3 -m unittest tests.test_bird_x tests.test_youtube_yt`

Gujiassh:
Refresh complete on current `main`.

The repo structure has moved since this PR was opened: the runtime now lives under `skills/last30days/scripts/`, and the subprocess timeout handling in `bird_x.py` / `youtube_yt.py` is already centralized through `lib/subproc.py` on current base.

So I narrowed this PR to the remaining current-base gap only:
- updated `skills/last30days/scripts/last30days.py` so `_cleanup_children()` falls back to `os.kill(pid, SIGTERM)` when `killpg/getpgid` are unavailable on Windows
- added `tests/test_last30days_cleanup.py` against the current plugin layout

Validation:
- `python3 -m unittest tests.test_last30days_cleanup tests.test_bird_x tests.test_youtube_yt tests.test_subproc -v`

Force-pushed refreshed head: `53d2f3a`. This should clear the dirty state while keeping only the remaining Windows-specific cleanup fix.

tmchow:
Thanks for catching this — the `os.killpg` call in `_cleanup_children` did indeed crash on Windows since the `(ProcessLookupError, PermissionError, OSError)` handler doesn't catch `AttributeError`. Your PR was already on the canonical path, so it's the cleanest of the batch; closing only because the rest of the multi-week-old batch was getting reapplied directly, and I wanted consistency. The same `hasattr(os, "killpg")` guard you wrote landed as commit 16ce073 with co-author credit to you. Appreciate the targeted regression tests on the PR — useful as I confirmed the fix.

Issue #91, opened 2026-03-25
Bug: bird_x.py os.killpg/os.getpgid crashes on Windows


Eshwar02:
## Description
In `scripts/lib/bird_x.py` (~line 195), the timeout handler calls `os.killpg(os.getpgid(proc.pid), signal.SIGTERM)` which doesn't exist on Windows.

The `preexec_fn=os.setsid` is guarded with `hasattr(os, 'setsid')`, but the kill path is not similarly guarded. The except clause catches `ProcessLookupError, PermissionError, OSError` but **not** `AttributeError`.

## Impact
Raises unhandled `AttributeError` on Windows when a subprocess times out.

## Suggested Fix
Guard with `hasattr` or catch `AttributeError`:
```python
try:
    os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
except (ProcessLookupError, PermissionError, OSError, AttributeError):
    proc.kill()
```

## Severity
Low

mvanhorn:
Speaking as Matt's firstmate:

**Triage (FM-CLEAR-L30-2):** bird_x Windows killpg — needs PR/check

PR #227, opened 2026-04-11
fix(windows): stabilize bundled Bird X search

### LINKED ISSUE #81: Embedded bird-search module crashes, causing X/Twitter searches to be skipped (OpenClaw)
## Description
I installed the skill according to the instructions on the skill's GitHub repository, using the command: `clawhub install last30days-official`. 

When running the `last30days` skill via OpenClaw, X/Twitter searches are silently skipped with the following warning in the status panel:
`⏭️ X: skipped — Bird installed but not authenticated — log into x.com in browser`

This happens even if valid `AUTH_TOKEN` and `CT0` are provided via environment variables (or `.env` file). 

Upon investigating `scripts/last30days.py` and the `bird-search` execution, the Python script calls the embedded Node script:
`node bird-search.mjs --whoami`

This embedded app crashes immediately due to three packaging issues in `scripts/lib/vendor/bird-search/`:

### 1. Missing `package.json` (ES Modules conflict)
The entire `package.json` is missing from the published `vendor/bird-search` folder. Because there is no `package.json` to define `"type": "module"`, Node treats the `.js` files inside `lib/` as CommonJS. When `bird-search.mjs` tries to import them, Node throws an error because the `.js` files use ESM `export`/`import` syntax.

### 2. Missing dependency (@steipete/sweet-cookie)
Because there is no `package.json` and no `node_modules` folder, the heavily relied upon `@steipete/sweet-cookie` dependency is missing, throwing `ERR_MODULE_NOT_FOUND` if you try to hotfix the module type.

### 3. Missing JSON config files
In `lib/twitter-client-constants.js`, it attempts to import JSON files:
`import queryIds from './query-ids.json' with { type: 'json' };`
However, `query-ids.json` and `features.json` were also stripped from the published `lib/` directory, resulting in another `ERR_MODULE_NOT_FOUND`.

### Proposed Fixes
1. Ensure the `package.json` (containing `"type": "module"` and the dependencies) is included in the published artifact.
2. Include the missing `.json` files (`query-ids.json` and `features.json`) in the artifact.
3. EITHER bundle the `node_modules` directly so the script can run purely offline, OR automatically trigger an `npm install` on the first run of the script.
4. **Alternative:** Fallback to the global `bird` CLI command if it's already installed on the system, rather than strictly enforcing the vendored script.

Chelebii:
## Summary
- avoid abrupt `process.exit(...)` shutdowns in the bundled Bird wrapper on Windows
- force UTF-8 decoding for Bird subprocess output to avoid Windows console/codepage crashes
- make X diagnose respect AUTH_TOKEN/CT0 values loaded from config
- write saved markdown output as UTF-8 and reconfigure Windows stdout/stderr to UTF-8

## Why
On Windows, free X search could authenticate but still fail during the bundled Bird runtime with a libuv assertion (`UV_HANDLE_CLOSING` in `src\\win\\async.c`).

This patch keeps the Bird wrapper alive long enough to flush handles cleanly, fixes Windows text decoding, and makes `/last30days --diagnose` report Bird auth correctly when cookies come from config.

## Testing
Tested on Windows with local AUTH_TOKEN/CT0 configured:
- `python scripts/last30days.py --diagnose`
- direct Bird smoke test via `bird_x.search_x('OpenClaw', '2026-04-04', '2026-04-11', 'quick')` -> returned 12 X items
- end-to-end run:
  - `python scripts/last30days.py OpenClaw --auto-resolve --emit=compact --save-dir=... --save-suffix=smoke --store`

## Related
- closes #81
- refs #91
- refs #110

### DIFF
--- a/scripts/last30days.py
+++ b/scripts/last30days.py
@@ -33,6 +33,11 @@ def ensure_supported_python(version_info: tuple[int, int, int] | object | None =
 
 ensure_supported_python()
 
+if os.name == "nt":
+    for stream in (sys.stdout, sys.stderr):
+        if hasattr(stream, "reconfigure"):
+            stream.reconfigure(encoding="utf-8", errors="replace")
+
 SCRIPT_DIR = Path(__file__).parent.resolve()
 sys.path.insert(0, str(SCRIPT_DIR))
 
@@ -103,7 +108,7 @@ def save_output(report: schema.Report, emit: str, save_dir: str, suffix: str = "
         content = emit_output(report, emit)
     else:
         content = render.render_full(report)
-    out_path.write_text(content)
+    out_path.write_text(content, encoding="utf-8")
     return out_path
 
 
--- a/scripts/lib/bird_x.py
+++ b/scripts/lib/bird_x.py
@@ -177,6 +177,8 @@ def _run_bird_search(query: str, count: int, timeout: int) -> Dict[str, Any]:
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             text=True,
+            encoding="utf-8",
+            errors="replace",
             preexec_fn=preexec,
             env=_subprocess_env(),
         )
@@ -336,6 +338,8 @@ def search_handles(
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
                 text=True,
+                encoding="utf-8",
+                errors="replace",
                 preexec_fn=preexec,
                 env=_subprocess_env(),
             )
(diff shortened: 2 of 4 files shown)


mvanhorn:
Merged, thanks @Chelebii. Solid Windows fix - the `process.exitCode` swap in bird-search.mjs is exactly right for preventing stdout truncation, and the subprocess encoding parameters plus stdout reconfigure cover the three common Windows encoding failure modes. Appreciate the credential propagation fix in env.py too.

PR #230, opened 2026-04-13
fix(cli): restore --days alias compatibility

BryanTegomoh:
## Summary
- restore `--days` as a backward-compatible alias for `--lookback-days`
- preserve existing `lookback_days` destination
- add a CLI test covering the alias and topic token preservation

## Why
Using the older `--days` flag currently causes argparse to treat the numeric value as topic input, which leads to nonsense runs like researching `30` instead of the requested topic.

## Repro
```bash
python3 scripts/last30days.py --quick --days 30 --emit compact "biosecurity AI agents"
```
Before this change, the run mis-parses and researches `30`.

## Validation
```bash
python3 -m unittest tests.test_cli_v3.CliV3Tests.test_build_parser_accepts_days_alias_and_preserves_topic_tokens
python3 scripts/last30days.py --quick --days 30 --emit compact "biosecurity AI agents"
```


mvanhorn:
Clean implementation, thanks @BryanTegomoh. The argparse alias approach is the right way to handle this.

PR #234, opened 2026-04-13
fix(bird_x): skip all-None engagement dicts

j-sperling:
## Summary

When a tweet has no engagement metrics at all (all fields missing from API
response), `_first_of()` returns `None` for each key, producing:
```python
{"likes": None, "reposts": None, "replies": None, "quotes": None}
```

This all-None dict propagates downstream to `signals.py` and `fusion.py`,
where engagement scoring treats it as "data exists but is zero" rather than
"no engagement data available."

## Fix

Single-line guard: return `None` instead of an all-None dict.

```python
# Before:
"engagement": engagement,

# After:
"engagement": engagement if any(v is not None for v in engagement.values()) else None,
```

## Test plan

- [x] `test_engagement_none_when_all_fields_missing` -- all-None dict becomes None
- [x] `test_engagement_preserved_when_any_field_present` -- dict kept when any metric exists
- [x] Updated `test_none_likes_when_missing` to assert `engagement` is None (not subscript into it)
- [x] Full test suite: 1024 passed, 15 pre-existing failures (store, watchlist, resolve, setup -- unrelated)

### DIFF
--- a/scripts/lib/bird_x.py
+++ b/scripts/lib/bird_x.py
@@ -460,7 +460,7 @@ def parse_bird_response(response: Dict[str, Any], query: str = "") -> List[Dict[
             "url": url,
             "author_handle": author_handle.lstrip("@"),
             "date": date,
-            "engagement": engagement,
+            "engagement": engagement if any(v is not None for v in engagement.values()) else None,
             "why_relevant": "",  # Bird doesn't provide relevance explanations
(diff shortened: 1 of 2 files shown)


mvanhorn:
Good catch, thanks @j-sperling. The all-None guard and test coverage are exactly right.

PR #238, opened 2026-04-14
Fix Codex skill routing when WebSearch is available

pejmanjohn:
## What this fixes

The skill currently tells agents:

- Codex is part of the "no WebSearch" path in Step 0
- sessions that skip WebSearch planning should add `--auto-resolve`

That is stale for modern Codex runs. Codex can run with native web search enabled, so routing by product name is wrong. In that case the skill should take the WebSearch-capable path and keep using agent planning via `--plan`.

## Why it matters

This is not just a wording cleanup. `SKILL.md` is executable behavior for the agent.

Before this change, the skill could steer Codex toward the weaker fallback path that was meant for non-WebSearch environments. That means:

- worse first-run routing
- weaker pre-research and entity resolution
- more reliance on engine-side `--auto-resolve`
- less clear guidance for Codex users and maintainers

After this change, the skill routes by actual capability:

- if WebSearch is available in the current session, use the standard/WebSearch path
- if WebSearch is not available, use the OpenClaw / non-WebSearch fallback path and add `--auto-resolve`

## Changes

- update both shipped skill files to branch on actual `WebSearch` capability instead of hard-coding Codex into the non-WebSearch path
- clarify that WebSearch-capable runs should keep using `--plan` and should not add `--auto-resolve`
- add a regression test covering both shipped skill copies
- align the internal `--auto-resolve` comment in `scripts/last30days.py`

## Verification

```bash
python3 -m pytest -q tests/test_skill_routing.py tests/test_version_consistency.py
```

## Before / After

Before, a Codex probe had to work around the shipped wording:

```text
I think WebSearch is available in this session. The environment exposes the `web` search tool, so functionally this is a WebSearch-capable runtime even though the skill text lists “Codex” as a non-WebSearch example. I’d branch on actual capability, not that example label.
```

After this change, the routing is explicit:

```text
Follow the standard setup flow. In a Codex session with web search enabled, you do have `WebSearch` capability, so the Step 0 router sends you to the `DO have WebSearch capability` branch rather than the OpenClaw fallback path.
```

### DIFF
--- a/.agents/skills/last30days/SKILL.md
+++ b/.agents/skills/last30days/SKILL.md
@@ -91,8 +91,8 @@ To detect first run: check if `~/.config/last30days/.env` exists. If it does NOT
 
 **When first run is detected, detect your platform first:**
 
-**If you do NOT have WebSearch capability (OpenClaw, Codex, raw CLI):** Run the OpenClaw setup flow below.
-**If you DO have WebSearch (Claude Code):** Run the standard setup flow below.
+**If you do NOT have WebSearch capability in this session (for example, OpenClaw or a raw CLI session):** Run the OpenClaw setup flow below.
+**If you DO have WebSearch capability in this session (for example, Claude Code or Codex with search enabled):** Run the standard setup flow below.
 
 ---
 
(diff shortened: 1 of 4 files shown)


tmchow:
Thanks for the careful read on this. The SKILL.md routing has been substantially rewritten since this PR, and the gating is now capability-based rather than platform-based: Step 0.55's Platform Gate reads "If your platform does NOT support WebSearch (e.g., OpenClaw, raw CLI), skip Steps 0.55 and 0.75" — Codex isn't hard-coded into the non-WebSearch path anymore, the WebSearch capability itself is the trigger. The PRECONDITION GATE in Research Execution lists "OpenClaw, raw CLI, Codex without web tools" as examples of non-WebSearch platforms, which is the qualifier you wanted. Closing as substantively resolved by the intervening SKILL.md rewrites — also the old paths this targets (root `SKILL.md` and `.agents/skills/last30days/SKILL.md`) no longer exist after the 3.1.1 restructure.

PR #240, opened 2026-04-14
Work around Claude plugin loader rejecting root skill paths

### LINKED ISSUE #239: Plugin fails to load in Claude Code: "Path escapes plugin directory: ./ (skills)"
## Problem
Installing `last30days@last30days-skill` in Claude Code (Windows, v2.9.5) produces:

```
last30days@last30days-skill [last30days]: Path escapes plugin directory: ./ (skills)
```

## Cause
In `.claude-plugin/plugin.json`:

```json
"skills": ["./"]
```

Claude Code rejects `"./"` because it resolves to the plugin root, which it treats as a path-escape. The `skills` field is expected to list subdirectories that each contain a `SKILL.md`, but this plugin places `SKILL.md` at the plugin root.

## Suggested fix
Move `SKILL.md` (and related skill assets) into a subdirectory, e.g. `skills/last30days/SKILL.md`, and update the manifest:

```json
"skills": ["skills/last30days"]
```

## Environment
- OS: Windows 11
- Claude Code: latest
- Plugin version: 2.9.5

pejmanjohn:
Fixes [#239](https://github.com/mvanhorn/last30days-skill/issues/239).

## Summary

Claude's marketplace loader currently fails to load this plugin when `.claude-plugin/plugin.json` points `skills` at the plugin root (`"./"`), even though the manifest validates and Anthropic's docs suggest that layout should be allowed.

This PR works around the loader behavior by shipping the runtime skill from a real plugin-local subdirectory and pointing the manifest there.

## Changes

- point the Claude plugin manifest at `./skills/last30days-plugin`
- add `skills/last30days-plugin/SKILL.md` as a mirrored copy of the root runtime skill
- add a regression test that keeps the plugin-local runtime skill identical to root `SKILL.md`
- clarify in the README that root `SKILL.md` remains the source of truth

## Verification

```bash
python3 -m pytest -q tests/test_claude_plugin_manifest.py tests/test_version_consistency.py
claude plugins validate .claude-plugin/plugin.json
```

### Before

In a disposable Claude local-scope install using the unpatched repo:

```text
last30days@last30days-skill
Status: ✘ failed to load
Error: Path escapes plugin directory: ./ (skills)
```

### After

In the same install flow using the patched worktree:

```text
last30days@last30days-skill
Status: ✔ enabled
```

Slash-command smoke test:

```text
/last30days
```

## Notes

This is a compatibility workaround for Claude's current marketplace loader behavior. The root `SKILL.md` remains the canonical runtime skill, and the plugin-local copy is kept in sync by test coverage.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -11,6 +11,6 @@
   "repository": "https://github.com/mvanhorn/last30days-skill",
   "license": "MIT",
   "keywords": ["research", "reddit", "twitter", "youtube", "tiktok", "instagram", "trends", "prompts", "polymarket", "github", "perplexity", "threads", "pinterest", "eli5", "hacker-news"],
-  "skills": ["./"],
+  "skills": ["./skills/last30days-plugin"],
   "hooks": {}
 }
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 **An AI agent-led search engine scored by upvotes, likes, and real money - not editors.**
 
-This README tracks the current v3 pipeline. The runtime skill spec lives in [skills/last30days/SKILL.md](skills/last30days/SKILL.md), which is the source of truth for the latest command and setup behavior.
+This README tracks the current v3 pipeline. The runtime skill spec lives in [SKILL.md](SKILL.md), which is the source of truth for the latest command and setup behavior. The Claude plugin ships a mirrored runtime copy at [skills/last30days-plugin/SKILL.md](skills/last30days-plugin/SKILL.md) because Claude's marketplace loader rejects a root-level `skills: ["./"]` manifest path in practice.
 
 Claude Code:
 ```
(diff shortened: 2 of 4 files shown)


maayan-t8:
Confirmed this fixes the issue. I hit the same `Path escapes plugin directory: ./ (skills)` error after a Claude Code update (v2.1.107). Applied the same change locally (patched both cache and marketplace `plugin.json` + copied `SKILL.md` into `skills/last30days-plugin/`) and the plugin loads cleanly now.

Clean fix, minimal change. 👍

tmchow:
Thanks for the workaround — the `skills` path-escape was a real bug on v3.0.5. v3.1.0 shipped the actual fix by dropping the `skills` key from `.claude-plugin/plugin.json` entirely (CHANGELOG closes #306), and the v3.1.1 restructure moved the runtime under `skills/last30days/` natively. So the loader-rejection workaround isn't needed anymore. Closing as resolved.

PR #257, opened 2026-04-15
fix: remove trailing slash from plugin paths to resolve path escape error

MaTriXy:
## Problem

Installing the plugin via Claude Code marketplace fails with:

```
✘ Plugin errors
└ 1 plugin error(s) detected:
  └ last30days@last30days-skill [last30days]: Path escapes plugin directory: ./ (skills)
```

## Root Cause

Both `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` use `"./"` (with trailing slash) as the path value. Claude Code's plugin path resolver treats the trailing slash in `"./"` as a path that escapes the plugin directory boundary.

## Fix

Removed the trailing slash from both path values:

| File | Before | After |
|------|--------|-------|
| `.claude-plugin/plugin.json` | `"skills": ["./"]` | `"skills": ["."]` |
| `.claude-plugin/marketplace.json` | `"source": "./"` | `"source": "."` |

`"."` correctly resolves to the plugin root directory where `SKILL.md` lives, without triggering the path escape check.

Simple fix for https://github.com/mvanhorn/last30days-skill/issues/239

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -15,7 +15,7 @@
         "name": "Matt Van Horn",
         "url": "https://github.com/mvanhorn"
       },
-      "source": "./",
+      "source": ".",
       "category": "productivity",
       "homepage": "https://github.com/mvanhorn/last30days-skill"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -11,6 +11,6 @@
   "repository": "https://github.com/mvanhorn/last30days-skill",
   "license": "MIT",
   "keywords": ["research", "reddit", "twitter", "youtube", "tiktok", "instagram", "trends", "prompts", "polymarket", "github", "perplexity", "threads", "pinterest", "eli5", "hacker-news"],
-  "skills": ["./"],
+  "skills": ["."],
   "hooks": {}
 }


mvanhorn:
@MaTriXy thank you so much for jumping in on this — really appreciate you taking the time to diagnose the root cause and ship a clean two-line fix. That kind of contribution is exactly what makes maintaining an OSS plugin actually sustainable, and I want to make sure you feel seen for it.

I ended up going a slightly different direction in v3.0.4 / v3.0.5 — dropping the `"skills"` key from `plugin.json` entirely instead of changing `"./"` to `"."`. After auditing every other plugin in the Claude Code marketplace ecosystem (compound-engineering, coding-tutor, codex, esper, all the Anthropic official plugins), none of them declare a `"skills"` key at all — Claude Code's plugin loader auto-discovers `skills/*/SKILL.md` when the key is absent. Removing it matches that ecosystem standard and avoids relying on path-resolution behavior that could shift in a future Claude Code release.

Your fix would have absolutely worked too — and honestly it's a more conservative change than what I shipped. Both paths solve the same bug. I'm closing this PR as superseded only because v3.0.4 already shipped a different cut of the same fix, not because the contribution wasn't valuable.

Hope you'll consider sending more PRs in the future — and thanks again for the help getting this in front of users faster.

PR #269, opened 2026-04-16
feat: add Firecrawl deep fetch + Exa MCP integration

benzntech:
## Summary
- Add Firecrawl MCP tools to allowed-tools (scrape + crawl)
- Add Exa MCP tools to allowed-tools (web search + crawling)
- Add Step 2b: deep fetch top web results with Firecrawl after Exa/WebSearch
- Step 2.5: append WebSearch results to saved raw file
- Security & Permissions: document Firecrawl + Exa credential handling
- INCLUDE_SOURCES config + TikTok/Instagram opt-in in NUX
- ScrapeCreators modal: 5 platforms instead of Reddit-only

## Test plan
- [ ] Run `/last30days` with Firecrawl MCP configured
- [ ] Verify Step 2b fetches and truncates articles correctly
- [ ] Verify Exa search results included in output
- [ ] Verify upstream v3.0.0 features still work


mvanhorn:
Sorry to be a broken record here but can you talk me through this from value to end user perspective? I believe there's one, I just want to hear your perspective.

mvanhorn:
@benzntech thoughts on above?

benzntech:
**Value to end user:**

Before this PR, `/last30days` research hits a wall when Exa/WebSearch returns short snippets — users still have to manually find and read the full articles themselves.

This PR fixes that:

1. **Complete answers, not links** — Firecrawl deep-fetches top results and truncates to ~8K tokens. Users get synthesized findings, not just URLs to click.

2. **Better topic coverage** — Exa's web search fills gaps where X/Twitter has no conversation (e.g., niche topics, non-English sources, older trends).

3. **Same workflow, better output** — No new UX. Just run `/last30days [topic]` and get richer, more grounded results.

**Example:** Before, researching "AI coding assistants" might return 3 tweets + 2 Reddit posts. After, it returns those + full summaries from top articles, HN threads, and product pages that weren't on social media.

mvanhorn:
Thanks @benzntech, the value framing makes sense and the Firecrawl integration itself looks well-scoped (depth-aware fetch counts, parallel scraping, fail-open on errors).

A few items I'd like to address before merge:

1. allowed-tools currently adds `mcp__exa__web_search_exa`, `mcp__exa__crawling_exa`, `mcp__firecrawl__firecrawl_scrape`, and `mcp__firecrawl__firecrawl_crawl`. Step 2b only calls `firecrawl_scrape`. Can we drop the other three (or wire them into an explicit step)? Keeping the surface as small as we use.

2. Step 2b is inserted after Step 2.5, which makes the reading order Step 2 -> Step 2.5 -> Step 2b. Could we renumber to Step 2.6 (or move Step 2b ahead of Step 2.5) so the order is monotone?

3. The variants/open/SKILL.md (version 2.1-open) is a parallel skill structure I don't have context on. Could you say a bit about the intent (target install path, who it serves, how it stays in sync with the v3 main skill)? Happy to discuss it in this PR or split it into a separate one if you prefer to land Firecrawl first.

4. The PR description mentions INCLUDE_SOURCES + TikTok/Instagram opt-in in NUX and ScrapeCreators modal: 5 platforms instead of Reddit-only. I don't see those changes in the diff. Is the description carrying claims from an earlier branch, or did some commits get lost in a rebase?

Once those are sorted, this looks like a solid integration. Thanks for the back-and-forth on the value framing.

mvanhorn:
@benzntech lmk! you've got me excited now haha

mvanhorn:
@benzntech ping!

tmchow:
Thanks for the thorough integration here, and for combining the Firecrawl and Exa work into a single PR. Closing — passing on this direction for now. The skill's web-search and deep-fetch story is intentionally lean (WebSearch + optional Exa via API key in the Python engine), and adding MCP-routed Firecrawl + Exa as always-on Step 2b would expand the credential surface and SKILL.md contract more than I want to take on right now. Not closed because of any defect in the PR — it's a scope/direction call. Happy to revisit if there's stronger user demand for MCP-based web sources down the road.

PR #270, opened 2026-04-17
fix: restore first-run wizard by moving skill to canonical path

giuseppebisemi:
## Summary

Restores the first-run wizard for Claude Code marketplace installs.

Since v3.0.5, `/last30days:last30days <topic>` has been skipping the Step 0 welcome flow on fresh installs. The setup wizard itself was still present in `scripts/lib/setup_wizard.py`, and the Step 0 instructions were still present in the user-facing skill body, but that body was no longer what Claude Code loaded for the slash command.

This does not change the wizard flow itself. It fixes which file Claude Code loads for `/last30days:last30days`, so the existing Step 0 logic becomes reachable again.

## Root cause

PR #267 added `commands/last30days.md` as a minimal slash-command stub and removed `skills/last30days-nux/SKILL.md`, which had been the canonical user-invocable skill path.

In local verification on current Claude Code, the slash-command loader used the stub as the model prompt and did not inject the full `SKILL.md` body via the `Skill` tool. As a result, Step 0 never reached the model, so fresh installs skipped the welcome text, setup modal, and topic picker.

## Fix

Adopt the skill-as-command layout for the Claude Code plugin while preserving the claude.ai `.skill` bundle layout.

This mirrors the skill-as-command layout used by [compound-engineering](https://github.com/EveryInc/compound-engineering-plugin) and avoids relying on a separate slash-command stub.

- Move the user-facing skill body from `SKILL.md` at repo root to `skills/last30days/SKILL.md`, which is the canonical Claude Code plugin path.
- Move the internal v3 spec from `skills/last30days/SKILL.md` to `skills/last30days-v3-spec/SKILL.md`.
- Delete `commands/last30days.md`, so command resolution comes from plugin slug + skill slug rather than a separate stub file.
- Update `scripts/sync.sh` to copy from `$SRC/skills/last30days/SKILL.md`.
- Update `scripts/build-skill.sh` to promote the canonical skill file back to bundle-root `SKILL.md` before stripping `skills/*` and `.claude-plugin/*`, so the claude.ai bundle still ships a single root `SKILL.md`.
- Update `tests/test_version_consistency.py` and `CLAUDE.md` to reference the new canonical path.

The runtime scripts at the plugin root are unchanged, so `${CLAUDE_PLUGIN_ROOT}/scripts/...` resolution continues to work.

## Compatibility

| Install path | Impact |
|--------------|--------|
| Claude Code marketplace | Fresh installs get the first-run wizard again; repeat runs still skip Step 0 when `SETUP_COMPLETE=true` is present |
| `bash scripts/sync.sh` | Continues to sync the skill into the Claude plugin cache plus `~/.agents/skills/last30days/` and `~/.codex/skills/last30days/`; Hermes remains supported when present; OpenClaw remains conditional on `variants/open` |
| claude.ai `.skill` bundle | Still builds as a single-root-`SKILL.md` bundle suitable for upload |
| Existing users with `SETUP_COMPLETE=true` | No visible behavior change |

## Verification

**Fresh install flow**

1. Removed `~/.config/last30days/.env` to simulate first run.
2. Applied the layout change to the local Claude plugin cache install.
3. Reloaded plugins.
4. Ran `/last30days:last30days nvidia`.
5. Verified that Claude resolved the skill base directory to `.../skills/last30days`, loaded the full `SKILL.md` body, showed the Step 0 welcome flow, and preserved the original topic through setup.

**Repeat run**

1. Re-ran `/last30days:last30days nvidia` with `SETUP_COMPLETE=true` present in `.env`.
2. Verified that Step 0 was skipped silently and the normal pipeline ran with the expected resolved flags.

**Packaging and sync**

- `bash scripts/sync.sh` succeeds with the new source path and still updates the intended downstream targets.
- `bash scripts/build-skill.sh` produces `dist/last30days.skill` successfully.
- The built `.skill` bundle stays below the 200-file cap and contains a single root `SKILL.md`.

## Testing checklist

- [x] Verified the canonical user-facing skill body matches the previous root `SKILL.md`
- [x] Verified the internal v3 spec matches the previous `skills/last30days/SKILL.md`
- [x] Verified `scripts/sync.sh` deploys from the canonical skill path
- [x] Ran `python3 -m unittest tests.test_version_consistency`
- [x] Ran `bash scripts/build-skill.sh`
- [x] Verified the built `.skill` bundle has a single root `SKILL.md` and stays under the 200-file cap
- [x] Verified fresh-install behavior restores the Step 0 wizard
- [x] Verified repeat-run behavior still skips Step 0 when `SETUP_COMPLETE=true` is present

Not tested: re-uploading the built `.skill` bundle into claude.ai and validating that path end-to-end.

## Recovery for existing v3.0.5 users

```bash
claude plugin update last30days@last30days-skill
/reload-plugins
```

No `.env` migration is required. Users who already completed setup keep `SETUP_COMPLETE=true` and continue to bypass Step 0. Users who installed v3.0.5 but never completed first run will see the wizard on their next invocation.


tmchow:
Thanks for tracking this down — the wizard issue was real on v3.0.5. The canonical-path migration ended up shipping a few days later in v3.1.1 (commit 72495c1, "Restructure as Codex plugin"), which moved the runtime under `skills/last30days/` along with the rest of the harness-specific reorg. So the underlying premise here is resolved on main; closing as superseded. Appreciate the detailed verification protocol on this one.

PR #282, opened 2026-04-18
perf: cache PreparedQuery per stream, skip double-normalize in dedupe

iliaal:
## What

Two small perf cleanups in the scoring hot path. No behavior change, no API surface change.

**1. Cache PreparedQuery once per stream**

`pipeline._normalize_score_dedupe` runs the scoring loop on ~40 items per stream. Each item calls `signals.local_relevance(item, ranking_query)` and then `snippet.extract_best_snippet(item, ranking_query)` across ~5 body windows. Both paths bottom out in `relevance.token_overlap_relevance`, which re-ran `tokenize(query)` and `_normalize_phrase(query)` every call. That's ~240 redundant tokenizations per stream on a query that never changes.

Fix: `relevance.PreparedQuery` holds `q_tokens`, `informative_q_tokens`, and `normalized_phrase`. The pipeline builds it once per stream and passes it through `annotate_stream` and `extract_best_snippet`. `token_overlap_relevance` accepts either a string or a prepared query, so nothing downstream had to change its signature.

**2. Stop normalizing twice in `_PreparedText`**

`dedupe._PreparedText.__init__` computed `norm = normalize_text(raw)`, then handed `norm` to `get_ngrams`, which started with `text = normalize_text(text)` again. I added a private `_ngrams_of_normalized` helper; `get_ngrams` keeps its public signature and `_PreparedText` skips the second pass.

## Measured

Micro-benchmarks on this machine:

- `tokenize(query)`: 1.4μs. `_normalize_phrase(query)`: 1.0μs. ~240 calls avoided per stream → ~0.6ms/stream, ~8ms per 15-stream run.
- `_PreparedText` build: 10.2μs → 7.3μs (1.39x). ~1000 builds per run → ~3ms saved.

Combined: ~10ms per run. Network I/O dominates wall clock, so treat this as a cleanup with a small incidental perf win.

## Verification

Full suite: **1088 pass, 15 pre-existing fail** (matches main baseline exactly). No new tests; the existing 170 tests across `test_relevance`, `test_signals_v3`, `test_snippet_v3`, `test_dedupe_v3`, `test_cluster_v3`, `test_pipeline_v3`, `test_youtube_relevance` cover the touched code paths.

## Files

~~~
scripts/lib/dedupe.py    | 13 ++++++++-----
scripts/lib/pipeline.py  |  6 ++++--
scripts/lib/relevance.py | 32 ++++++++++++++++++++++++++------
scripts/lib/signals.py   | 10 +++++++---
scripts/lib/snippet.py   |  5 +++--
5 files changed, 48 insertions(+), 18 deletions(-)
~~~

### DIFF
--- a/skills/last30days/scripts/lib/dedupe.py
+++ b/skills/last30days/scripts/lib/dedupe.py
@@ -39,11 +39,14 @@ def normalize_text(text: str) -> str:
     return re.sub(r"\s+", " ", text).strip()
 
 
+def _ngrams_of_normalized(norm: str, n: int = 3) -> set[str]:
+    if len(norm) < n:
+        return {norm} if norm else set()
+    return {norm[index:index + n] for index in range(len(norm) - n + 1)}
+
+
 def get_ngrams(text: str, n: int = 3) -> set[str]:
-    text = normalize_text(text)
-    if len(text) < n:
-        return {text} if text else set()
-    return {text[index:index + n] for index in range(len(text) - n + 1)}
+    return _ngrams_of_normalized(normalize_text(text), n)
 
 
 def jaccard_similarity(left: set[str], right: set[str]) -> float:
@@ -90,7 +93,7 @@ class _PreparedText:
 
     def __init__(self, raw: str) -> None:
         norm = normalize_text(raw)
-        self.ngrams = get_ngrams(norm) if norm else set()
+        self.ngrams = _ngrams_of_normalized(norm)
         self.tokens = _tokenize(norm)
 
 
--- a/skills/last30days/scripts/lib/pipeline.py
+++ b/skills/last30days/scripts/lib/pipeline.py
@@ -30,6 +30,7 @@ from . import (
     query,
     reddit,
     reddit_public,
+    relevance,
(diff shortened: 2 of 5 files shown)


mvanhorn:
@iliaal sorry about this - restructure under you in 72495c1 (shipped 3.1.1) moved the lib files. Both perf premises still hold on main:

- `tokenize(query)` is still recomputed per-call inside `token_overlap_relevance` at `skills/last30days/scripts/lib/relevance.py:98`. `PreparedQuery` isn't there.
- `_PreparedText.__init__` at `skills/last30days/scripts/lib/dedupe.py:92-93` still does `normalize_text(raw)` then hands `norm` to `get_ngrams`, which re-normalizes.

The 5 affected files all moved together under `skills/last30days/scripts/lib/`: relevance.py, signals.py, snippet.py, pipeline.py, dedupe.py. The restructure was a pure file-move with no edits to function bodies, so the diff applies line-for-line at the new paths.

iliaal:
@mvanhorn Rebased as well

PR #283, opened 2026-04-18
refactor(normalize): extract _join_comment_excerpts helper

iliaal:
## What

`_normalize_reddit`, `_normalize_hackernews`, and `_normalize_github` each inlined the same 5-line comprehension to stringify and space-join the first 3 top_comments' excerpt field. Extract one helper, call it from all three.

```python
def _join_comment_excerpts(top_comments, key, limit=3) -> str:
    return " ".join(
        str(comment.get(key) or "").strip()
        for comment in top_comments[:limit]
        if isinstance(comment, dict)
    )
```

The comment field name varies per source: Reddit and GitHub use `"excerpt"`, HN uses `"text"`. Passing it as a parameter keeps each caller's intent visible at the call site.

## Why

Small dedup. Any future change to comment-excerpt handling (truncation, separator choice) now happens in one place instead of three. Behavior unchanged.

## Verification

Full suite: **1088 pass, 15 pre-existing fail** (matches main baseline).

## Files

~~~
scripts/lib/normalize.py | 31 ++++++++++++++++---------------
1 file changed, 16 insertions(+), 15 deletions(-)
~~~

### DIFF
--- a/skills/last30days/scripts/lib/normalize.py
+++ b/skills/last30days/scripts/lib/normalize.py
@@ -110,6 +110,19 @@ def _first_present(d: dict[str, Any], keys: tuple[str, ...], default: Any) -> An
     return default
 
 
+def _join_comment_excerpts(
+    top_comments: list[Any],
+    key: str,
+    limit: int = 3,
+) -> str:
+    """Space-join the `key` field from the first `limit` dict-shaped comments."""
+    return " ".join(
+        str(comment.get(key) or "").strip()
+        for comment in top_comments[:limit]
+        if isinstance(comment, dict)
+    )
+
+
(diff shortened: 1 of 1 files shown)


mvanhorn:
@iliaal sorry about this - restructure under you in 72495c1 (shipped 3.1.1) moved the file. The change still applies cleanly; only the path is different.

New location: `skills/last30days/scripts/lib/normalize.py`. The three duplicated join patterns are still there at lines 172, 341, 444. Confirmed via grep on current main. The restructure was a pure file-move with no edits to the function bodies, so your diff applies line-for-line at the new path.

iliaal:
@mvanhorn Rebased, should merge cleanly now

PR #285, opened 2026-04-19
fix: output contract + planner breadth + entity grounding (Hermes Agent Use Cases)

mvanhorn:
## Summary

Fixes the 2026-04-19 `/last30days Hermes Agent (Actual) Use Cases` failure where two consecutive runs dumped the engine's internal `## Ranked Evidence Clusters` scratchpad as user output instead of the canonical `What I learned:` prose synthesis.

Five root causes, five units — all landing in this PR.

## What's fixed

**Unit 1 — SKILL.md LAW 6 + render.py envelope comments.** The prior canonical-boundary text read "Pass through the lines ABOVE this boundary verbatim," which the model scoped broadly to include the scratchpad. `render_compact` now wraps evidence in `<!-- EVIDENCE FOR SYNTHESIS -->` comments and the emoji footer in `<!-- PASS-THROUGH FOOTER -->` comments. LAW 6 names the failure mode with a concrete self-check string (`### 1.` followed by a score tuple) and includes a worked transformation example.

**Unit 2 — planner intent-modifier handling.** Topics ending in "use cases", "workflows", "review", "examples", etc. now get their modifier stripped from `search_query` and paraphrased across 4-5 subqueries. `_max_subqueries` cap raised 3 → 5 for how_to/opinion/product. Deterministic fallback gets its own 3-subquery paraphrase fanout when an intent modifier is present.

**Unit 3 — deterministic fallback defaults.** `_infer_intent` default changed `breaking_news` → `concept` (safer freshness posture for unclassified topics). Recency regexes added above the default so genuinely time-sensitive topics still hit `strict_recent`. `_keyword_query` now quotes only title-cased multi-word proper nouns, not the user's full topic. New stderr warning when running without `--plan` and without an LLM provider.

**Unit 4 — reranker entity grounding.** `_primary_entity(topic)` strips intent-modifier suffixes. Fallback rerank subtracts `ENTITY_MISS_PENALTY` (25 pts) when primary entity absent from title + snippet. LLM rerank path gets a "Primary entity grounding" hint. Skips empty-text candidates (image-only TikToks etc.) to avoid false negatives. Kills the "Nate Herk Managed Agents video scoring #2 on a Hermes Agent query" failure.

**Unit 5 — always-on planner trace.** `pipeline.py` now emits a unified stderr block with intent, freshness, cluster_mode, and one line per subquery after plan finalization. No more `--debug` gating. `source=` annotation distinguishes external (`--plan`), llm, and deterministic paths.

## Verification

[...]

PR #286, opened 2026-04-20
feat: clickable source links in synthesis (no raw URLs)

mvanhorn:
## Summary

Every cited source in the `/last30days` synthesis — X handles, subreddits, publications, YouTube channels, Polymarket markets — now renders as a blue CMD-clickable markdown link in the terminal. Raw URL strings are still forbidden everywhere in the output.

The old rule "NEVER paste raw URLs" blocked both raw URLs and clickable links. But Claude Code renders CommonMark `[text](url)` as a blue hyperlink with the URL hidden — exactly what users want. Matt Mullenweg surfaced this in a reply to a thread where Jay-K asked about link-backs:

> on my Claude Code, you can CMD-click a link, and it opens. Spread the credit, it costs nothing.
> — @photomatt

Flipping the rule is a pure prompt change. No Python engine changes — the raw research dump already carries URLs per item.

## What changed

- **`SKILL.md` — invert the URL formatting rule.** New rule: "every link must be `[text](url)`, never a raw URL string." Plain text is allowed only as a fallback when the raw research has no URL for a source.
- **`SKILL.md` — update every citation template.** Narrative paragraphs, KEY PATTERNS, RECOMMENDATIONS per-item "Sources:" lines, and the `🌐 Web` + `🗣️ Top voices` stats-block lines all show markdown-link examples now.
- **`SKILL.md` — retire the "I have all the links... Just ask." closer.** Links are now inline, so the closer is redundant. Output ends at the QUERY_TYPE-specific invitation.
- **`SKILL.md` — two new PRE-PRESENT SELF-CHECK items:** (7) no raw `http://`/`https://` strings, (8) no broken empty links like `[name]()`.
- **`README.md` — one-line note** under "How it works" calling out the clickable links.
- **`release-notes.md` — dated entry** crediting @jay_k and @photomatt.

## Test plan

- [ ] Run `bash scripts/sync.sh` to deploy the updated `SKILL.md`.
- [ ] In Claude Code on Ghostty, run `/last30days kanye west` (GENERAL). Verify citations in "What I learned" show as blue underlined text; CMD-click opens the source in the browser.
- [ ] Verify the `🌐 Web:` stats line shows blue source names with no visible URL strings.
- [ ] Verify the `🗣️ Top voices:` line shows blue @handles and r/subreddits.
- [ ] Run `/last30days nano banana pro prompts for Gemini` (PROMPTING). Verify output ends on the "describe your vision" prompt with no "I have all the links..." closer.
- [ ] Scan the full output of one run for `http://` or `https://` substrings outside of `[text](url)` — should be zero.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,7 @@ These platforms don't have relationships with each other. X doesn't know what Re
 2. **The agent resolves who matters.** Finds X handles (including founders), GitHub repos, subreddits, TikTok hashtags, YouTube channels. For "Kanye West" it knows r/hiphopheads, @kanyewest, and "bully review" on YouTube. For "OpenClaw" it resolves openclaw/openclaw on GitHub and fetches live star counts.
 3. **All sources searched in parallel.** Multi-query expansion. Results scored by engagement, relevance, freshness.
 4. **The depth nobody else has.** Full YouTube transcripts from reaction videos. Top Reddit comments with upvote counts. TikTok captions. Polymarket odds. Not just titles and links.
+   Every cited source is a blue CMD-clickable link in the terminal (X handles, subreddits, publications). No raw URL strings — clean text, one click to the source.
 5. **Same story, merged.** Wireless Festival announced on Reddit, discussed on X, ticket prices on TikTok = one cluster, not three separate items.
 6. **Synthesized into one brief.** Grounded in specific data. Cited by source. Ranked by what people actually engage with. Not "here's what I found." It's "here's what matters."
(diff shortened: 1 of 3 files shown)


mvanhorn:
Closing — this branch was built on a stale main and diverged from the output-contract hardening that landed in PR #285 (and subsequent commits a0d61b0, 4d9f29d, 4388fed, etc.). Live test confirmed my three enforcement commits broke the working format rather than fixing it. Restarting with a minimal link-rule change on top of current main.

PR #288, opened 2026-04-20
fix: use removeprefix("r/") for subreddit names, not lstrip("r/")

alexkeyme:
## Problem

`str.lstrip("r/")` treats its argument as a character set, not a literal prefix — it strips leading `r` and `/` repeatedly. For subs starting with `r` the result is wrong:

```python
>>> "robotics".lstrip("r/")
'obotics'
>>> "r/ruby".lstrip("r/")
'uby'
```

The result is that subs starting with "r" can't be searched by the agent. 

## Fix

Replace with `str.removeprefix("r/")` (Python 3.9+, project already requires 3.12) at three sites:

- `scripts/last30days.py:306` — CLI arg parsing
- `scripts/lib/reddit_public.py:201` — URL building
- `scripts/lib/entity_extract.py:109` — mention counting

## Test

One regression test in `tests/test_reddit_public.py` that fails on the bug and passes on the fix. Existing tests happened not to cover this case — they used names like `ClaudeAI` / `MachineLearning` whose first post-prefix character isn't in `{r, /}`.

### DIFF
--- a/scripts/last30days.py
+++ b/scripts/last30days.py
@@ -303,7 +303,7 @@ def main() -> int:
     depth = "deep" if args.deep else "quick" if args.quick else "default"
     try:
         x_related = [h.strip() for h in args.x_related.split(",") if h.strip()] if args.x_related else None
-        subreddits = [s.strip().lstrip("r/") for s in args.subreddits.split(",") if s.strip()] if args.subreddits else None
+        subreddits = [s.strip().removeprefix("r/") for s in args.subreddits.split(",") if s.strip()] if args.subreddits else None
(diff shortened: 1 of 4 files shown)


jenniferscott11718-byte (approved):

alexkeyme:
@jenniferscott11718-byte how's your policy - will you merge and close this when you want to include it, or am I supposed to do anything?

tmchow:
Excellent catch — this is a real bug that's been silently mangling any subreddit starting with `r` (so `r/robotics` became `obotics`, `r/ruby` became `uby`). Closing because the four call sites moved during the 2026-04-23 layout restructure (under `skills/last30days/scripts/`), so the patch wouldn't apply cleanly. Rather than ask you to rebase, I applied the same `removeprefix("r/")` fix directly on main as commit 5994b4f with co-author credit to you. Great pickup.

PR #290, opened 2026-04-21
[codex] Make memory directory configurable

davemorin:
## Summary

- Add a single `LAST30DAYS_MEMORY_DIR` default in `SKILL.md` for saved research files.
- Route skill `--save-dir` examples and related saved-file instructions through that variable.
- Update the comparison helper to use the same default and add a regression test that guards against reintroducing the hard-coded save-dir command.

## Why

The skill hard-coded `~/Documents/Last30Days` in several places. Centralizing the default makes it easy for users to choose a different memory file directory by changing or exporting one variable.

## Validation

- `/opt/homebrew/bin/python3.14 -m unittest tests.test_version_consistency`
- `/opt/homebrew/bin/python3.14 -m unittest tests.test_cli_v3`
- `/opt/homebrew/bin/python3.14 scripts/last30days.py "memory dir smoke" --mock --emit=compact --save-dir=/tmp/last30days-memory-smoke --save-suffix=test`
- `bash -n scripts/compare.sh`
- `git diff --check`
- `PATH="/opt/homebrew/bin:$PATH" bash scripts/sync.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -202,15 +202,15 @@ Intelligent search, fun judge, cross-source cluster merging, single-pass compari
 
 ### Highlights
 
-Auto-save research briefings to `~/Documents/Last30Days/` as topic-named .md files. Every run now builds a personal research library automatically - no more manual copy-paste.
+Auto-save research briefings to the default memory directory as topic-named .md files. Every run now builds a personal research library automatically - no more manual copy-paste.
 
 ### Added
 
-- Auto-save complete research briefings (synthesis, stats, follow-up suggestions) to `~/Documents/Last30Days/{topic-slug}.md` after every run
+- Auto-save complete research briefings (synthesis, stats, follow-up suggestions) to the default memory directory after every run
 - Kebab-case filename generation from topic (e.g., "Claude Code skills" -> `claude-code-skills.md`)
 - Duplicate topic handling: appends date suffix instead of overwriting (e.g., `claude-code-skills-2026-03-05.md`)
 - Agent mode (`--agent`) also saves research files
-- Brief confirmation after save: "Saved to ~/Documents/Last30Days/{slug}.md"
+- Brief confirmation after save with the saved file path
 
 ### Credits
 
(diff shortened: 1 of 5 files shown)


mvanhorn:
Thanks for picking this up, Dave. Direction is right and the regression test is a nice touch. A few things I'd tweak before this goes out of draft:

1. Prose that references the bare variable name reads a bit rough. Lines like "saves research briefings to LAST30DAYS_MEMORY_DIR" used to show users a concrete path. Worth keeping the default inline, e.g. "saves research briefings to LAST30DAYS_MEMORY_DIR (defaults to ~/Documents/Last30Days)".

2. The scripts/last30days.py hunk is docstring-only. compute_save_path_display still only ~-collapses $HOME paths, so if someone sets the memory dir to /mnt/storage/... the footer shows the absolute path anyway. Either wire up display handling for non-HOME paths or drop the hunk.

3. Worth adding a short Configuration section or env var note in SKILL.md so new users discover LAST30DAYS_MEMORY_DIR without reading the source.

4. The test pins two files. A Path.rglob sweep for stray ~/Documents/Last30Days occurrences would guard the whole repo (README, docs, etc.) against reintroduction.

Happy to merge once you mark it ready for review.

mvanhorn (approved):
Nicely iterated, Dave. Diff is clean, backward compat is preserved (default resolves to ~/Documents/Last30Days when the env var is unset), your smoke test creates the file, and compare.sh passes `bash -n`.

One non-blocking heads-up: your new `test_no_stray_hardcoded_memory_dir_paths` walks the filesystem via `Path.rglob`, so on any dev machine with a local `docs/plans/` directory (gitignored, used for internal planning) it trips over historical planning docs that reference the old path. Adding `docs` to `skip_dirs` or switching to `git ls-files` would make it green locally too. I'll push a follow-up for that so you don't have to bounce another commit.

Thanks for the contribution. Merging.

davemorin:
Thanks for the review and quick merge, @mvanhorn! Appreciate the feedback on the prose — much cleaner now.

mvanhorn:
Thanks Dave - glad that landed. The regression test was the right call; makes future refactors of that env-var handling a lot safer.

Issue #254, opened 2026-04-15
Add pytest-cov coverage gate to CI (60% baseline, measured on first run)


dgenio:
## Category
`code-quality` | **Priority: P2** | **Effort: S (1–2 days)**

## Why This Matters
Coverage tooling is fully configured in [`pyproject.toml`](pyproject.toml) but never runs in CI. Without a gate, coverage can drop to zero on core modules with no signal. With issue #249 in place, adding a coverage gate is a small, safe extension.

**Evidence:**
- [`pyproject.toml`](pyproject.toml) has correct `[tool.coverage.run]` and `[tool.coverage.report]` sections but nothing invokes them
- The test suite is large (1,000+ tests per [`README.md`](README.md)) but actual coverage on core modules is unknown without a CI run

## Definition of Done
- [ ] `.github/workflows/ci.yml` (issue #249) extended to run `pytest` with coverage:
  ```bash
  python -m pytest -q --tb=short --cov=scripts --cov-report=term-missing
  ```
- [ ] `[tool.coverage.run]` in [`pyproject.toml`](pyproject.toml) updated to add:
  ```toml
  fail_under = 60
  ```
  _(60% gate — achievable with existing tests; intended to rise over time, not as a ceiling)_
- [ ] If overall coverage is already above 60% on first run, raise the gate to the actual measured baseline (round down to nearest 5%)
- [ ] Coverage baseline (overall %, per-module breakdown for [`scripts/lib/pipeline.py`](scripts/lib/pipeline.py), [`scripts/lib/schema.py`](scripts/lib/schema.py), [`scripts/lib/render.py`](scripts/lib/render.py)) recorded in a comment in `pyproject.toml` above `fail_under`
- [ ] Coverage badge added to [`README.md`](README.md)
- [ ] [`AGENTS.md`](AGENTS.md) (issue #250) updated to note: do not reduce `fail_under` without documenting why in a PR comment

## Dependencies
- Depends on #249 (CI workflow must exist first)
- Coordinate with #250 (AGENTS.md note)

Issue #253, opened 2026-04-15
Add ruff linting and format checking to CI


dgenio:
## Category
`code-quality` | **Priority: P1** | **Effort: S (1–2 days)**

## Why This Matters
No formatting or style enforcement exists. [`pyproject.toml`](pyproject.toml) has no `[tool.ruff]` section. Style inconsistencies accumulate across 40+ library modules with no gate. With CI now in place (issue #249), this is the natural next code-quality addition.

**Evidence:**
- [`pyproject.toml`](pyproject.toml) contains only `pytest` and `pytest-cov` in dev tools; no linter config anywhere in the repo
- Note at top of [`scripts/last30days.py`](scripts/last30days.py): `# ruff: noqa: E402` — ruff is clearly the intended linter but is not yet configured repo-wide
- 40+ modules under [`scripts/lib/`](scripts/lib/) with no enforced style

## Definition of Done
- [ ] `[tool.ruff]` section added to [`pyproject.toml`](pyproject.toml):
  ```toml
  [tool.ruff]
  target-version = "py312"
  line-length = 120

  [tool.ruff.lint]
  select = ["E", "F", "I"]      # pycodestyle errors, pyflakes, isort
  exclude = ["scripts/lib/vendor/*"]
  ```
- [ ] `ruff check .` run locally; all violations resolved or suppressed with inline `# noqa: <code>` comments as a one-time cleanup (do not silence wholesale with `per-file-ignores = ["*": ...]`)
- [ ] `ruff format --check .` run locally; any format differences resolved
- [ ] `.github/workflows/ci.yml` (issue #249) extended with a `ruff` job:
  ```yaml
  - run: pip install ruff
  - run: ruff check .
  - run: ruff format --check .
  ```
- [ ] CI fails on new violations from this point forward
- [ ] [`AGENTS.md`](AGENTS.md) (issue #250) updated to note: run `ruff check .` before committing

## Dependencies
- Depends on #249 (CI workflow must exist before adding ruff to it)
- Coordinate with #250 (AGENTS.md note)

tmchow:
Thanks for suggestion! this isn't being considered at this time but appreciate the suggestion.

PR #295, opened 2026-04-21
feat: add CI workflow to run pytest on push and PR

### LINKED ISSUE #249: Add CI workflow to run pytest on every push and PR
## Category
`code-quality` | **Priority: P0** | **Effort: S (1–2 days)**

## Why This Matters
The test suite has 1,000+ tests that never run in CI. Every PR merges without any pass/fail signal. This is the single most critical infrastructure gap in the repo.

**Evidence:** [`.github/workflows/release.yml`](.github/workflows/release.yml) only triggers on version tags and only builds/releases the artifact. There is no workflow that runs `pytest` on push or PR.

## Definition of Done
- [ ] New workflow file `.github/workflows/ci.yml` triggers on `push` to `main` and on all pull requests
- [ ] Installs dependencies via `uv sync --group dev` (or `pip install -e ".[dev]"` as fallback)
- [ ] Runs `python -m pytest -q --tb=short` (matching the config in [`pyproject.toml`](pyproject.toml))
- [ ] Job fails if any test fails
- [ ] Workflow CI badge added to the top of [`README.md`](README.md)
- [ ] Verified to pass on current `main` before merging

## Notes
- Exclude `scripts/lib/vendor/` from test discovery (already handled by `pyproject.toml` coverage settings)
- Use Python 3.12 to match `requires-python = ">=3.12"` in [`pyproject.toml`](pyproject.toml)
- Keep the existing [`release.yml`](.github/workflows/release.yml) unchanged; this is a new parallel workflow

## Dependencies
None — fully parallelizable with other issues in this wave.

davemorin:
## Summary

- Adds `.github/workflows/ci.yml` that runs pytest on every push to `main` and on all pull requests
- Uses `uv sync --group dev` for dependency installation (matching `pyproject.toml` dev group)
- Runs on Python 3.12 (matching `requires-python = ">=3.12"`)
- Intentionally minimal — no lint or coverage gates (those are #253 and #254)

The test suite has 1,000+ tests that currently have no CI signal. Every PR merges without pass/fail feedback.

Fixes #249

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,22 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - uses: astral-sh/setup-uv@v6
+
+      - run: uv sync --group dev
+
+      - run: uv run python -m pytest -q --tb=short


davemorin:
Note: CI is failing, but all 16 test failures are **pre-existing on main** — this PR only adds the workflow file, no code changes. These failures were previously invisible because no CI existed.

The failing tests are in `test_store.py` (7), `test_setup_openclaw.py` (3), `test_watchlist_commands.py` (2), `test_resolve.py` (2), `test_generate_synthesis_inputs_v3.py` (1, missing script file). This is exactly the kind of visibility this CI workflow is designed to provide.

mvanhorn:
Thanks for writing this. Closing as redundant: #317 (just merged) added validate.yml running the focused plugin-contract + version-consistency suite on every push/PR, which stays green. A full pytest workflow would be yellow indefinitely until the 14 pre-existing failures get worked through. If those get unblocked later, worth revisiting a full-suite job then.

PR #298, opened 2026-04-21
fix: expand sync.sh import verification to cover all library modules

### LINKED ISSUE #96: Bug: sync.sh import verification only checks 4 of 12+ modules
## Description
In `scripts/sync.sh` (~line 32), import verification runs:
```bash
python3 -c "from lib import youtube_yt, bird_x, render, ui"
```

This only checks 4 modules but misses ~8 newer ones: `brave_search`, `truthsocial`, `bluesky`, `parallel_search`, `xiaohongshu_api`, `scrapecreators_x`, `polymarket`, `hackernews`.

## Impact
Broken imports in newer modules would only be caught at runtime.

## Suggested Fix
Add all modules to the import verification.

## Severity
Low

davemorin:
## Summary

- Expands the import check in `sync.sh` from 7 modules to all 46 library modules
- Skips browser-dependent modules (`chrome_cookies`, `cookie_extract`, `safari_cookies`) that require browser binaries
- Both sync targets (common + Hermes) are updated

**Before:** `sync.sh` only checked `briefing, store, watchlist, youtube_yt, bird_x, render, ui` — broken imports in 39 other modules would only surface at runtime.

**After:** all core modules are verified at sync time. Verified that the full import list passes locally.

Fixes #96

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/scripts/sync.sh
+++ b/scripts/sync.sh
@@ -57,7 +57,7 @@ sync_target() {
 
   if (
     cd "$target/scripts" &&
-    python3 -c "import briefing, store, watchlist; from lib import youtube_yt, bird_x, render, ui; print('  Import check: OK')"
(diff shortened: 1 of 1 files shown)


mvanhorn:
#317 landed while this was open and reorganized the runtime. The file is now at skills/last30days/scripts/sync.sh. The 7-to-46 module expansion is still valuable; could you rebase against current main? If you'd rather I carry it forward in a follow-up, no problem, just say the word.

tmchow:
Thanks @davemorin — this one got fully overtaken by main. `scripts/sync.sh` itself was deleted in commit `9fb19ea` (May 15), so the post-sync import smoke-test the PR expanded no longer has a home. The dev install path is now `npx skills add . -g -y` and that path doesn't run a custom import check. The expanded module-list intent is moot; closing as superseded. If a post-install verification is still desirable in the new workflow, that'd be a separate PR with a different shape.

PR #318, opened 2026-04-24
fix: make hooks.json polyglot for Gemini CLI and Claude Code compatibility

flyingnobita:
Gemini CLI reads the `timeout` field in `hooks.json` as milliseconds, causing a 5ms timeout that fails on execution. This PR removes the timeout so both CLIs can fall back to their generous defaults (600s for Claude, 60s for Gemini).

Additionally, Gemini CLI uses `${extensionPath}` instead of `${CLAUDE_PLUGIN_ROOT}`. This PR uses a short-circuit bash command to gracefully try both directories, ensuring a single `hooks.json` file is perfectly compatible with both environments without any build steps.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/scripts/check-config.sh\"",
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh\"",
             "timeout": 5
           }
         ]
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,8 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh\"",
-            "timeout": 5
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh\""
           }
         ]
       }


mvanhorn:
Thanks @flyingnobita, the diagnosis is clear (Gemini reading `timeout: 5` as 5ms) and the dual-path idea is nice. Two asks before this lands:

1. Guard against silent fallthrough when both env vars are unset. With a plain `;` the bash step runs unconditionally even if both `cd` calls fail, which executes `bash hooks/scripts/check-config.sh` from whatever the caller's CWD happens to be. Can we change it to either:

   ```json
   "command": "cd \"${CLAUDE_PLUGIN_ROOT:-}\" 2>/dev/null || cd \"${extensionPath:-}\" 2>/dev/null || exit 0; bash hooks/scripts/check-config.sh"
   ```

   (exit 0 on both-unset, so no false-failure notification) or:

   ```json
   "command": "{ cd \"${CLAUDE_PLUGIN_ROOT:-}\" 2>/dev/null || cd \"${extensionPath:-}\" 2>/dev/null; } && bash hooks/scripts/check-config.sh"
   ```

   (`&&` so the script only runs when a cd succeeded). Either is fine, the goal is just that we never run `bash hooks/scripts/check-config.sh` from an arbitrary CWD.

2. Confirm with a live Claude Code run. Could you install this branch as a plugin in Claude Code (`/plugin install` or the equivalent local path) and confirm the SessionStart hook still fires the `/last30days: Ready - N sources active.` banner? I saw the Gemini side in the PR description; I'd like to make sure the Claude Code path is still working end-to-end after the shape change, not just compiling. A screenshot or pasted banner output in the PR thread is plenty.

If you also have OpenClaw handy, same check there would be a bonus, it inherits Claude Code's plugin schema so I expect it works, but confirming is cheap.

Once both are in, happy to merge.

flyingnobita:
**Item 1 fixed** - switched to Option B (`&&` gating):

```json
"command": "{ cd \"${CLAUDE_PLUGIN_ROOT:-}\" 2>/dev/null || cd \"${extensionPath}\" 2>/dev/null; } && bash hooks/scripts/check-config.sh"
```

Note: `:-` is kept on `${CLAUDE_PLUGIN_ROOT:-}` since it's a real shell env var in Claude Code (defensive against `set -u`). It is intentionally not applied to `${extensionPath}` because Gemini CLI substitutes that as a JSON template token at config-parse time and only recognizes the bare `${varname}` form - adding `:-` breaks the template and causes the Gemini hook to fail.

**Item 2 confirmed** - tested in both runtimes:

- **Claude Code**: installed via `/plugin marketplace add /path/to/last30days-skill` then `/plugin install last30days`
- **Gemini CLI**: installed via `gemini extensions install /path/to/last30days-skill`
- **Openclaw**: installed via `openclaw plugins install last30days --marketplace /home/rpi5/Projects/last30days-skill --dangerously-force-unsafe-install`

The SessionStart hook fires and the plugin loads correctly with 9 sources active in Gemini and Openclaw runtimes (see screenshot below).

One note: the latest version of Claude Code no longer surfaces hook output directly to the user in the terminal. The `SessionStart:startup hook success: /last30days: Ready — 9 sources active.` message no longer appears as visible terminal output - instead it is injected into the model's context as a system message. As shown in the screenshot, the plugin still loads properly and the hook output is present in context, it's just no longer displayed to the user.  

#### Claude Code
image  

#### Gemini CLI
image

#### Openclaw
image

mvanhorn:
Two correct catches in here: the Gemini-side timeout-as-ms bug, and that Gemini exposes the plugin path as `${extensionPath}` not `${CLAUDE_PLUGIN_ROOT}`. The cd approach has two bugs though.

Manual test (clean HOME with no global config, project dir with `.claude/last30days.env` containing 5 API keys):

```
Test 1: cd command, CLAUDE_PLUGIN_ROOT set
  Output: 'Ready to use. Run /last30days to get started — setup takes 30 seconds.'
  (welcome message; script found NO config)

Test 2: parameter expansion, CLAUDE_PLUGIN_ROOT set
  Output: 'Ready — 10 sources active.'
  (script found project config)

Test 3: parameter expansion, only extensionPath set (Gemini scenario)
  Output: 'Ready — 10 sources active.'
  (works for Gemini AND finds project config)

Test 4: cd command, only extensionPath set (Gemini scenario)
  Output: 'bash: hooks/scripts/check-config.sh: No such file or directory'
  (Gemini path doesn't even fire; cd "" succeeds going to $HOME, OR clause is skipped)
```

What's happening:

1. The cd resets bash's CWD to the plugin root, but `hooks/scripts/check-config.sh` reads `.claude/last30days.env` as a relative path that needs to be the user's project root. Test 1 shows the script falls through to the new-user welcome because it can't find the project config any more.
2. `cd "${CLAUDE_PLUGIN_ROOT:-}"` with an empty-but-set var succeeds (cd to $HOME), so the OR clause never fires the extensionPath fallback. Test 4 shows the script fails in the exact Gemini-only scenario this PR is meant to fix.

Parameter expansion sidesteps both:

```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh\""
```

Same fallback chain (Claude -> Gemini -> cwd as last resort), still drops the bad timeout, and the script keeps running from the user's project root so PROJECT_ENV resolves correctly. Tests 2 and 3 above are this form running.

flyingnobita (commented):
Good catch.

Test 1: I was using ~/.config/last30days/.env for both claude and gemini and didn't know about `.claude/last30days.env`

Test 4: I see what you mean. For some reasons I can't reproduce it on my end (as you can see in my screenshot) so I never caught this.

P.S. The slash command doesn't work in openclaw and I was about to submit a fix but I noticed the latest commits actually causes the skill to not work for me in openclaw.

tmchow:
@flyingnobita Greptile flagged a description-vs-diff mismatch here — your PR description says the timeout was removed so both CLIs fall back to their generous defaults, but `"timeout": 5` was still present in `hooks/hooks.json`. I've pushed a follow-up commit that actually removes that field so Claude (600s) and Gemini (60s) both use their CLI defaults as intended.

PR #317, opened 2026-04-24
Restructure as Codex plugin

clairernovotny:
## Summary

This restructures the repository so it can be installed and discovered as a Codex plugin while preserving the existing Claude/OpenClaw-facing skill behavior.

The important layout change is moving the runtime payload from root-level files into the Codex plugin skill payload:

- `SKILL.md` -> `skills/last30days/SKILL.md`
- `scripts/` -> `skills/last30days/scripts/`
- `agents/` -> `skills/last30days/agents/`
- `assets/` -> `skills/last30days/assets/`

It also adds/updates the Codex plugin metadata:

- expands `.codex-plugin/plugin.json`
- adds `.agents/plugins/marketplace.json`
- updates Claude/Hermes/docs/tests/build scripts to point at the moved payload

Fixes #315.

## Why this is needed

Codex plugin discovery needs plugin metadata and a skill payload layout that can be loaded from a marketplace/plugin root. The prior root-level `SKILL.md` + root `scripts/` layout works as a plain skill repository, but it is not sufficient for a Codex plugin install because Codex needs the plugin manifest/marketplace metadata and a discoverable skill under `skills/`.

Authoritative Codex sources:

- The Codex CLI marketplace command supports installing from local marketplace root directories, so this repo can be made directly installable as a local marketplace/plugin root: https://github.com/openai/codex/blob/a9f75e5cda2d6ff469a859baf8d2f50b9b04944a/codex-rs/cli/src/marketplace_cmd.rs#L32-L37
- OpenAI's bundled `plugin-creator` skill defines Codex plugin scaffolding around a required `.codex-plugin/plugin.json`: https://github.com/openai/codex/blob/a9f75e5cda2d6ff469a859baf8d2f50b9b04944a/codex-rs/skills/src/assets/samples/plugin-creator/SKILL.md#L1-L3
- The same source says plugin creation always creates `.codex-plugin/plugin.json`, fills the manifest/interface section, and creates/updates `.agents/plugins/marketplace.json` when marketplace metadata is needed: https://github.com/openai/codex/blob/a9f75e5cda2d6ff469a859baf8d2f50b9b04944a/codex-rs/skills/src/assets/samples/plugin-creator/SKILL.md#L49-L54
[...]

PR #320, opened 2026-04-24
Add Brave browser support for X/Twitter cookie extraction

kaushikgopal:
## Summary

- Adds Brave browser as a supported cookie source for X/Twitter auth on macOS
- Refactors `chrome_cookies.py` to extract a shared `_extract_chromium_cookies_macos` helper — Chrome and Brave both call it with their own DB path and Keychain service name
- Adds `_find_brave_cookies_db()` for profile discovery: tries `Default/` first, then scans numbered `Profile N/` directories
- Registers `"brave"` in the browser extractors dict and inserts it into the macOS `auto` order (`chrome → brave → firefox → safari`)

## Why

Brave is a Chromium-based browser that stores cookies using the same v10 AES-128-CBC encryption as Chrome. The only differences are the DB path (`BraveSoftware/Brave-Browser/`) and the macOS Keychain service name (`"Brave Safe Storage"`). Users who use Brave as their primary browser and are logged into X there were silently skipped by the auto-detection path, with no fallback and no diagnostic.

## Before / After

Before, `auto` mode on macOS tried `chrome → firefox → safari`. A Brave user logged into X would get no X results without manually configuring `XAI_API_KEY` or `xurl`.

After, `auto` mode tries `chrome → brave → firefox → safari`. The setup wizard reports `X cookies found in brave` on success, and the skill works without any manual key configuration for Brave users.

## Commit Notes

- `feat(cookies): add Brave browser cookie extraction for macOS`
  - Why: Brave users were excluded from zero-config X/Twitter auth
  - What changed: refactored Chrome extraction into a shared `_extract_chromium_cookies_macos(db_path, keychain_service, ...)` helper; added `extract_brave_cookies_macos` and `_find_brave_cookies_db`; registered `"brave"` in `cookie_extract.py`
  - Implication: `FROM_BROWSER=auto` and the setup wizard now work for Brave with no config changes

## Testing

- Verified on macOS with Brave as primary browser:
  ```
  python3 scripts/last30days.py setup
  # → "X cookies found in brave"
  ```
- `extract_cookies_with_source("brave", ".x.com", [...])` returns tokens from `Default/Cookies`
- `extract_cookies_with_source("auto", ...)` falls through Chrome (not installed) and succeeds on Brave

### DIFF
--- a/skills/last30days/scripts/lib/chrome_cookies.py
+++ b/skills/last30days/scripts/lib/chrome_cookies.py
@@ -285,19 +285,24 @@ def _find_brave_cookies_db() -> Optional[Path]:
     """Find Brave's Cookies database on macOS.
 
     Tries the Default profile first, then scans numbered Profile directories
-    in creation order. Brave creates extra profiles as "Profile 1", "Profile 2",
-    etc. alongside Default.
+    by most-recently-modified. Brave creates extra profiles as "Profile 1",
+    "Profile 2", etc. alongside Default; the most recently used one is the
+    likeliest to hold current cookies. Lexicographic sort would visit
+    "Profile 10" before "Profile 2", which can return the wrong profile.
     """
     default = BRAVE_BASE_DIR / "Default" / "Cookies"
     if default.exists():
         return default
 
     try:
-        for child in sorted(BRAVE_BASE_DIR.iterdir()):
-            if child.is_dir() and child.name.startswith("Profile "):
-                candidate = child / "Cookies"
-                if candidate.exists():
-                    return candidate
+        candidates = [
+            child for child in BRAVE_BASE_DIR.iterdir()
+            if child.is_dir() and child.name.startswith("Profile ")
+        ]
+        for child in sorted(candidates, key=lambda p: p.stat().st_mtime, reverse=True):
+            candidate = child / "Cookies"
+            if candidate.exists():
+                return candidate
     except OSError:
         pass
 
--- a/skills/last30days/scripts/lib/chrome_cookies.py
+++ b/skills/last30days/scripts/lib/chrome_cookies.py
@@ -1,9 +1,12 @@
-"""Chrome cookie extraction for macOS.
+"""Chrome and Brave cookie extraction for macOS.
 
-Extracts cookies from Chrome's encrypted SQLite database using only stdlib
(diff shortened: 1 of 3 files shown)


mvanhorn:
Thanks for this. Two asks before merge:

1. Rebase: #317 landed while you were writing this and reorganized the runtime. The files are now at skills/last30days/scripts/lib/chrome_cookies.py and skills/last30days/scripts/lib/cookie_extract.py.

2. Chrome regression verification: the _extract_chromium_cookies_macos refactor changes the Chrome path, not just adds Brave alongside it. I want to be sure Chrome doesn't silently regress for existing users. Can you confirm you've tested on a clean Chrome install (fresh Keychain prompt, cookies extracted successfully) AND a clean Brave install?

Let me know when rebased and retested. If you'd rather I carry it forward in a follow-up, just say the word.

tmchow:
Thanks for this — clean refactor. The shared `_extract_chromium_cookies_macos` helper is exactly the right shape for the Chromium family (Brave, Edge, Arc, Vivaldi all share the v10 AES-128-CBC + Keychain pattern with only a path + service-name swap), so this opens the door to those without further churn.

Rebased onto current main: the layout migration (`scripts/lib/` → `skills/last30days/scripts/lib/`) applied cleanly via Git's rename detection, no manual cherry-pick needed. Folded in a tiny mechanical fix to `tests/test_chrome_cookies.py` to retarget two mocks at the new `_get_chromium_encryption_key` symbol (the old `_get_chrome_encryption_key` wrapper is no longer in the decryption path). Authorship preserved on your commit.

`pytest tests/test_chrome_cookies.py tests/test_safari_cookies.py tests/test_cookie_extract.py` — 49 passed. Auto-order on macOS is `chrome → brave → firefox → safari` as expected. Marking ready-for-maintainer.

tmchow:
@kaushikgopal pushed a follow-up to your branch addressing Greptile's P1 on `_find_brave_cookies_db`: lexicographic sort visits profiles as `Profile 1 → Profile 10 → Profile 2`, which can return the wrong profile (e.g. `Profile 10` before `Profile 2` when X cookies live in `Profile 2`). Switched to mtime-descending sort so the most-recently-used profile wins, which is the likeliest to hold current X cookies.

PR #333, opened 2026-04-29
Preserve clean mode for last-run state

iamitp:
## Summary\n- skip last-run writes when LAST30DAYS_CONFIG_DIR is set to an empty string\n- show last-run summary from the configured state directory in the hook\n- add regression coverage for empty and custom config directories\n\n## Tests\n- python3 -m unittest tests.test_last_run_state


tmchow:
Thanks @iamitp — closing this in favor of #334, which is the same fix on the canonical post-plugin-layout path (`skills/last30days/scripts/last30days.py`) with regression coverage for the no-config, custom-config-write, and custom-config-hook-read cases. Same author, same intent, same files; #334 just supersedes this one. Will triage #334 next.

PR #334, opened 2026-04-29
Preserve clean mode for last run state

iamitp:
## Summary
- skip writing last-run.json when LAST30DAYS_CONFIG_DIR is explicitly set to an empty string
- read last-run.json from LAST30DAYS_CONFIG_DIR in the session hook when a custom config directory is configured
- apply the CLI writer fix to the rebased canonical path: skills/last30days/scripts/last30days.py
- add regression coverage for no-config mode, custom config writes, and custom config hook reads

## Validation
- uv run pytest tests/test_last_run_state.py
- full uv run pytest currently reports 1327 passed / 13 failed on upstream main-adjacent code; failures are outside this patch area, covering footer nudge suppression, setup device-auth timing mocks, and store/watchlist expectations.

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -74,15 +74,16 @@ else
   LAST_RUN_FILE="$HOME/.config/last30days/last-run.json"
 fi
 LAST_RUN_LINE=""
-if [[ -n "$LAST_RUN_FILE" && -f "$LAST_RUN_FILE" ]]; then
-  LAST_RUN_LINE=$(LAST_RUN_FILE="$LAST_RUN_FILE" python3 - <<'PY' 2>/dev/null
+if [[ -n "$LAST_RUN_FILE" && -f "$LAST_RUN_FILE" ]] && command -v python3 &>/dev/null; then
+  LAST_RUN_LINE=$(LAST_RUN_FILE="$LAST_RUN_FILE" python3 - <<'PY' 2>/dev/null || true
 import datetime
 import json
 import os
 
 path = os.environ["LAST_RUN_FILE"]
 try:
-    d = json.load(open(path))
+    with open(path) as fh:
+        d = json.load(fh)
     topic = (d.get("topic") or "?")[:60]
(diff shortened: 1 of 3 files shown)


tmchow:
Thanks @iamitp — rebased onto current `main` (clean, no conflicts) and pushed as `a017a5c`. The fix is on the canonical post-plugin-layout path: `_write_last_run` now bails when `env.CONFIG_DIR is None` (set by `LAST30DAYS_CONFIG_DIR=""`), and `hooks/scripts/check-config.sh` reads `last-run.json` from a custom `LAST30DAYS_CONFIG_DIR` when configured. Closed #333 in favor of this one; all 3 tests in `tests/test_last_run_state.py` pass locally. Flagging for maintainer review.

tmchow:
Pushed a fixup (b296a65) addressing the three Greptile findings. The meaningful catch is P1: with `set -euo pipefail` and a `last-run.json` present, the python3 command substitution exited 127 when python3 was absent — the SessionStart hook failed silently with empty output. I gated the block on `command -v python3 &>/dev/null` and added `|| true` as belt-and-suspenders against unexpected Python crashes. The two P2s are stylistic: hoisted `import datetime` to the module's top-level imports in `last30days.py`, and switched to a `with open(...)` context manager in the embedded Python heredoc.

Heads-up: prior auto-applied commits (bc8cf0d / 9586e83 / 323576e) left the branch in a broken state — the bash script had a duplicated `if [[ ... ]]; then` (syntax error), and `import datetime` was removed from inside `_write_last_run` without being added at module top (NameError silently swallowed by `except Exception: pass`, meaning `last-run.json` would never be written). The fixup supersedes those.

Empirically verified: `PATH=/tmp/nopy_bin /bin/bash hooks/scripts/check-config.sh` (a coreutils-only PATH with no python3) now produces normal output instead of exiting 127 with empty stdout; full test suite green (1483 passed, 4 skipped); targeted suite `test_last_run_state.py` + `test_plugin_contract.py` green (7 passed).

PR #338, opened 2026-05-02
fix(render): use forward slashes in save-path footer for Windows

dzivkovi:
## Summary

The `📎 Raw results saved to …` footer line prints a mangled path on Windows: it mixes a Unix tilde with backslashes (`~/Documents\Last30Days\foo-raw-v3.md`), which neither File Explorer, PowerShell, nor a `file://` URI can resolve. One-line fix: use `Path.as_posix()` so the home-relative segment always uses forward slashes.

## Changes

- `skills/last30days/scripts/last30days.py:190` — change `f\"~/{relative}\"` to `f\"~/{relative.as_posix()}\"` in `compute_save_path_display`.

The internal pathlib resolution was already correct; only the display string was mangled. macOS and Linux output is unchanged because their separator is already `/`.

## Testing

Repro on Windows (PowerShell, fresh `/last30days` run):

```
python3 last30days.py \"british airways middle east\" --emit=compact --save-dir=\"\$HOME/Documents/Last30Days\"

# before this fix
📎 Raw results saved to ~/Documents\Last30Days\british-airways-middle-east-raw.md
                                  ^^^                                         ^
                                  Windows backslashes — path is unresolvable

# after this fix
📎 Raw results saved to ~/Documents/Last30Days/british-airways-middle-east-raw.md
                                  ^^^                                         ^
                                  Forward slashes — opens in Explorer & shells
```

Verified by hitting the change with the same engine invocation that surfaced the bug. The user-visible footer now matches the Unix-style path the docs already use elsewhere (e.g. SKILL.md line 264).

- [ ] Ran \`uv run python -m pytest -q --tb=short\` — not run; change is one line in a string-formatting helper with no behavior change for non-Windows hosts and no public API change. Happy to add a unit test if you'd like one (would assert that `compute_save_path_display` never contains a backslash on any platform).
- [x] No `scripts/` directory structure changes; `sync.sh` not required.

## Related Issues

None tracked. Reproduced live on Windows 11 / Python 3.12.10 / Git Bash, on the v3.1.1 plugin cache.

### DIFF
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -197,7 +197,7 @@ def compute_save_path_display(save_dir: str, topic: str, suffix: str, emit: str)
         relative = raw.relative_to(home)
         return f"~/{relative.as_posix()}"
     except ValueError:
-        return str(raw)
+        return raw.as_posix()
 
 
 def read_synthesis_file(path: str) -> str:
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -195,7 +195,7 @@ def compute_save_path_display(save_dir: str, topic: str, suffix: str, emit: str)
     try:
         home = _Path.home().resolve()
         relative = raw.relative_to(home)
-        return f"~/{relative}"
+        return f"~/{relative.as_posix()}"
     except ValueError:
         return str(raw)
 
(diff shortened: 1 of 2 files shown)


tmchow:
Pushed `0e353ae` to address both Greptile P2 catches: the `except ValueError` fallback in `compute_save_path_display` now returns `raw.as_posix()` so paths outside `$HOME` also get forward slashes on Windows (previously `str(raw)` would still emit `C:\Users\...`), and the `import shutil` inside `test_compute_save_path_display_uses_posix_slashes_under_home` is hoisted to the module-level import block alongside `tempfile`/`subprocess`/etc. All 42 tests in `tests/test_cli_v3.py` + `tests/test_render_v3.py` pass under Python 3.12.

PR #340, opened 2026-05-03
fix(youtube): surface transcript-fetch ratio + add degraded nudge for stale yt-dlp

dzivkovi:
## Summary

When \`yt-dlp\` is installed but stale (or otherwise unable to fetch transcripts for any returned videos), runs previously reported YouTube as fully successful in two user-facing surfaces - the footer and the quality nudge. The engine logs the failure correctly at default stderr level (\`[YouTube] Got transcripts for 0/N videos (N failed)\`), but that line gets buried in 100+ lines of parallel-source progress output and is contradicted by the success-shaped conclusion surfaces that follow. This PR makes both surfaces honest about transcript-fetch state.

## Changes

- **\`scripts/lib/render.py\`**: footer always renders \`M/N with transcripts\` instead of suppressing the segment when \`with_transcripts == 0\`. The very case that needs to be loud (zero transcripts captured) was the case the footer hid. Format is \`M/N\` (not bare \`M\`) so the denominator is in the message and the user does not have to cross-reference the videos count.

- **\`scripts/lib/quality_nudge.py\`**: adds a third tier between \"active\" and \"missing\": **degraded**. Triggered when \`yt-dlp\` is installed AND videos were returned AND transcript-fetch ratio is below threshold (default 50%, tunable via \`DEGRADED_TRANSCRIPT_THRESHOLD\` env var). Emits an actionable nudge: \"YouTube returned N videos but only M transcripts captured. The most common cause is a stale yt-dlp binary - YouTube's caption format changes frequently and old binaries silently fail every transcript. Update via your package manager: scoop update yt-dlp (Windows), brew upgrade yt-dlp (macOS), or pip install -U yt-dlp.\" Threshold rationale: 50% accommodates a few legitimate caption-disabled videos in a multi-video result, but a stale-binary run that fails every transcript trips the nudge cleanly. Score impact: degradation is informational, not score-affecting; YouTube still counts as active in \`score_pct\` so users do not see their score drop for a fixable client-side issue.

- **\`scripts/last30days.py\`**: populates \`youtube_videos_count\` and \`youtube_transcripts_count\` in the \`research_results\` dict it passes to \`compute_quality_score\`, enabling the new degraded check at the call site.

- **\`tests/test_quality_nudge.py\`**: 6 new \`TestYouTubeDegraded\` cases cover zero-transcripts-flags-degraded, partial-above-threshold-does-not-flag, zero-videos-does-not-flag (no false positives on absence), one-of-three-flags-degraded, threshold-tunable-via-config, and degraded-does-not-affect-score.

- **\`tests/test_render_v3.py\`**: 4 new \`YoutubeFooterTranscriptRatioTests\` cases cover zero-transcripts-with-videos-renders-zero-over-total (the regression repro), partial-renders-ratio, full-renders-ratio, and no-videos-suppresses-entire-segment.

## Testing

- [x] All 29 new test cases verified GREEN with the fix and RED without it (temp-reverted both files separately to confirm each test catches the specific regression it asserts).
- [x] Full \`test_quality_nudge.py\` (25 tests) and \`test_render_v3.py\` (33 tests) suites pass.
- [x] Existing pre-PR failures in \`test_store.py\`, \`test_subproc.py\`, \`test_watchlist_commands.py\`, and \`test_version_consistency.py\` reproduce on pristine \`upstream/main\` without these changes - not regressions from this PR.
- [x] Integration-validated against an intentionally stale \`yt-dlp 2025.03.31\` binary placed first on PATH:

  Pre-fix footer:
  \`\`\`
  🔴 YouTube: 3 videos │ 386,815 views
  \`\`\`

  Post-fix footer:
  \`\`\`
  🔴 YouTube: 3 videos │ 386,815 views │ 0/3 with transcripts
  \`\`\`

  Post-fix stderr nudge:
  \`\`\`
  Research quality: 4/5 core sources.
  Missing: X/Twitter.
  Degraded: YouTube.

  Free fixes:
    - YouTube returned 3 videos but only 0 transcripts captured. The most
      common cause is a stale yt-dlp binary - YouTube's caption format
      changes frequently and old binaries silently fail every transcript.
      Update via your package manager: scoop update yt-dlp (Windows),
      brew upgrade yt-dlp (macOS), or pip install -U yt-dlp.
  \`\`\`

- [x] No \`scripts/\` directory structure changes; \`sync.sh\` not required.

## Out of scope

Deliberately not in this PR (each its own future PR):

- **Schema change to expose \`transcripts_captured\` in the EVIDENCE FOR SYNTHESIS block** so synthesizing models can flag degradation in prose. Larger change touching \`schema.py\` and the EVIDENCE block; deserves separate review.
- **Auto-update of yt-dlp.** Out of scope; the engine should never silently mutate a user binary. The nudge directs the user to their own package manager.
- **Generalizing \"degraded\" to other sources.** Reddit, X, etc. could have analogous \"returned but degraded\" states. Not generalizing now; this PR fixes the specific high-impact case where one user spent six weeks of research with silently-empty YouTube transcripts before noticing.

## Related Issues

None tracked. Reproduced live on Windows 11 / Python 3.12.10 / Git Bash with yt-dlp 2025.03.31 (year-stale) before discovery.

### DIFF
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -890,6 +890,8 @@ def main() -> int:
                 1 for it in youtube_items
                 if (it.metadata.get("transcript_highlights") or it.metadata.get("transcript_snippet"))
             ),
+            "youtube_error": report.errors_by_source.get("youtube"),
+            "x_error": report.errors_by_source.get("x"),
         }
         quality = quality_nudge.compute_quality_score(config, research_results)
         if quality.get("nudge_text"):
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -880,7 +880,18 @@ def main() -> int:
     # Show quality nudge if applicable
     try:
         from lib import quality_nudge
-        quality = quality_nudge.compute_quality_score(config, {})
+        # Populate transcript-fetch ratio so quality_nudge can detect the
+        # degraded-YouTube failure mode (videos returned but transcripts
+        # silently failed - typically a stale yt-dlp binary).
+        youtube_items = report.items_by_source.get("youtube") or []
+        research_results = {
+            "youtube_videos_count": len(youtube_items),
+            "youtube_transcripts_count": sum(
+                1 for it in youtube_items
+                if (it.metadata.get("transcript_highlights") or it.metadata.get("transcript_snippet"))
+            ),
+        }
+        quality = quality_nudge.compute_quality_score(config, research_results)
         if quality.get("nudge_text"):
             sys.stderr.write(f"\n{quality['nudge_text']}\n")
             sys.stderr.flush()
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -865,7 +865,18 @@ def main() -> int:
     # Show quality nudge if applicable
     try:
         from lib import quality_nudge
-        quality = quality_nudge.compute_quality_score(config, {})
+        # Populate transcript-fetch ratio so quality_nudge can detect the
+        # degraded-YouTube failure mode (videos returned but transcripts
+        # silently failed - typically a stale yt-dlp binary).
+        youtube_items = report.items_by_source.get("youtube") or []
+        research_results = {
+            "youtube_videos_count": len(youtube_items),
+            "youtube_transcripts_count": sum(
+                1 for it in youtube_items
+                if (it.metadata.get("transcript_highlights") or it.metadata.get("transcript_snippet"))
+            ),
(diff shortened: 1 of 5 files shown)


tmchow:
Thanks - good catch. Footers that hide the zero case turn the failure mode into silence, which is exactly the bug we needed surfaced. The 50% degraded threshold (tunable via DEGRADED_TRANSCRIPT_THRESHOLD) is the right hygiene call - tolerates one legitimately caption-disabled video in a multi-video result without false-flagging, while a stale-binary run that fails every transcript still trips loudly. Rebased onto main, tests pass, marking ready-for-maintainer.

PR #343, opened 2026-05-05
fix: prefer sandboxed Safari cookie path

### LINKED ISSUE #342: Safari cookie path is the legacy ~/Library/Cookies (sandboxed Safari moved cookies to ~/Library/Containers years ago)
## Summary

`safari_cookies.py` reads `~/Library/Cookies/Cookies.binarycookies` (line 110), but on every modern macOS (since the App Sandbox shipped for Safari, ~10.14) Safari stores cookies inside its sandbox container at:

```
~/Library/Containers/com.apple.Safari/Data/Library/Cookies/Cookies.binarycookies
```

The legacy `~/Library/Cookies/` path is usually an empty leftover directory. So even when a user is logged into x.com in Safari, the engine reports "no AUTH_TOKEN/CT0" and the X source returns 0 results. The first-run setup wizard's "log into x.com in Safari and re-run" advice silently fails for the same reason.

## Repro

1. Log into x.com in Safari (modern macOS — anything ≥10.14, so basically all users).
2. Run `/last30days <named-entity-topic>` on Claude Code.
3. Observe: stderr says `💡 Unlock X: log into x.com in Firefox or Safari, then re-run.`, and the engine footer shows X with 0 items.
4. `ls -la ~/Library/Cookies/Cookies.binarycookies` → `No such file or directory`.
5. `ls -la ~/Library/Containers/com.apple.Safari/Data/Library/Cookies/Cookies.binarycookies` → file exists and is live (mtime updates whenever Safari is open).

## Verification that the fix works

A user-side workaround that worked on macOS 25.4.0 / Safari (May 2026):

```sh
ln -s ~/Library/Containers/com.apple.Safari/Data/Library/Cookies/Cookies.binarycookies \
      ~/Library/Cookies/Cookies.binarycookies
```

After the symlink, `extract_safari_cookies_macos('x.com', ['auth_token','ct0'])` returns both cookies (auth_token len=40, ct0 len=160) and the X source starts populating on the next run.

## Suggested fix

`safari_cookies.py:110` should try the sandboxed-container path first, then fall back to the legacy path. Sketch:

```python
candidates = [
    Path.home() / "Library" / "Containers" / "com.apple.Safari" / "Data" / "Library" / "Cookies" / "Cookies.binarycookies",
    Path.home() / "Library" / "Cookies" / "Cookies.binarycookies",
]
cookie_path = next((p for p in candidates if p.exists()), candidates[0])
```

Reading the sandboxed file does require Full Disk Access for the calling process (Terminal / Claude Code / Codex CLI / etc.) on macOS. The `[safari] Permission denied` message at line 118 is still useful, but the path probe should land on the right file before that check fires — otherwise users get a misleading "no cookies found" instead of "permission denied," and they reach for AUTH_TOKEN/CT0 manually instead of granting Full Disk Access.

A second small UX nit: when the legacy path doesn't exist, the current code reports nothing actionable. Consider logging the candidate paths tried and pointing users to System Settings → Privacy & Security → Full Disk Access if the sandboxed path exists but isn't readable.

## Environment

- macOS Darwin 25.4.0
- Safari (default user-installed, not Technology Preview)
- Skill version: last30days v3.1.1 (cache: `~/.claude/plugins/cache/last30days-skill/last30days/3.1.1/`)
- Host: Claude Code (CLI)
- File: `skills/last30days/scripts/lib/safari_cookies.py:110`

Bortlesboat:
## Summary
- prefer Safari's modern sandboxed Cookies.binarycookies path before the legacy ~/Library/Cookies path
- keep the legacy path as fallback
- make the permission-denied regression deterministic across platforms

## Validation
- `uv run --no-sync pytest tests/test_safari_cookies.py tests/test_env_include_sources_default.py -q`
- `git diff --check`

Fixes #342

### DIFF
--- a/skills/last30days/scripts/lib/safari_cookies.py
+++ b/skills/last30days/scripts/lib/safari_cookies.py
@@ -107,7 +107,18 @@ def extract_safari_cookies_macos(
     if sys.platform != "darwin":
         return None
 
-    cookie_path = Path.home() / "Library" / "Cookies" / "Cookies.binarycookies"
+    cookie_paths = [
+        Path.home()
+        / "Library"
+        / "Containers"
+        / "com.apple.Safari"
+        / "Data"
+        / "Library"
+        / "Cookies"
+        / "Cookies.binarycookies",
+        Path.home() / "Library" / "Cookies" / "Cookies.binarycookies",
+    ]
+    cookie_path = next((path for path in cookie_paths if path.exists()), cookie_paths[0])
 
     try:
         raw = cookie_path.read_bytes()
--- a/tests/test_safari_cookies.py
+++ b/tests/test_safari_cookies.py
@@ -193,6 +193,37 @@ class TestErrorPaths:
         assert result["auth_token"] == "test_auth_abc123"
         assert result["ct0"] == "test_ct0_xyz789"
 
+    def test_falls_back_to_legacy_safari_cookie_path(self, tmp_path: Path):
+        # Sandboxed path is intentionally NOT created — only the legacy path exists.
+        legacy_dir = tmp_path / "Library" / "Cookies"
+        legacy_dir.mkdir(parents=True)
+        legacy_data = _build_binary_cookies_file(
+            [_build_page([_build_cookie_record(".x.com", "auth_token", "legacy_auth")])]
+        )
+        (legacy_dir / "Cookies.binarycookies").write_bytes(legacy_data)
+
+        sandbox_path = (
+            tmp_path
+            / "Library"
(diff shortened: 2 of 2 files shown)


tmchow:
Pushed a follow-up commit to address Greptile's P2 catch — the legacy fallback path (`~/Library/Cookies/Cookies.binarycookies`) wasn't exercised by a test, only the sandbox-wins case was. Added `test_falls_back_to_legacy_safari_cookie_path` so both branches of the resolver are now covered.

PR #339, opened 2026-05-03
docs: add CONFIGURATION.md + README pointers

dzivkovi:
## Summary

Adds CONFIGURATION.md at repo root - a focused configuration reference covering save paths, the per-source API-key matrix, reasoning and web-search backend priority, the trend-monitoring stack (\`--store\` + \`watchlist.py\` + \`briefing.py\`), and per-client patterns. Surfaces the project-scoped \`.claude/last30days.env\` config file (currently only referenced in \`hooks/scripts/check-config.sh\`) and the existing trend-monitoring scripts that the README did not previously surface for users. Updates README with a short \"Configuration\" section pointing to the new file, plus a one-line callout linking a 6-min community architecture walkthrough.

## Changes

- New \`CONFIGURATION.md\` (~256 lines), organized by how often each layer is touched: per-run flags → env vars / .env → optional trend-monitoring stack → per-client patterns.
- Documents \`.claude/last30days.env\` project-scoped config file priority over the global \`~/.config/last30days/.env\`.
- Documents the existing \`--store\` / \`watchlist.py\` / \`briefing.py\` trend-monitoring stack.
- README: short \"Configuration\" section after \"Bring your own keys\"; one-line \"Going deeper\" callout
- All CLI surface claims (watchlist subcommands, briefing modes, source dedupe key, env file priority chain) fact-checked against the live \`scripts/\` source before commit.

## Testing

- [x] Docs-only change; no engine code modified.
- [x] No \`scripts/\` directory structure changes; \`sync.sh\` not required.
- [x] CONFIGURATION.md renders cleanly on GitHub (verified via fork branch preview).

## Related Issues

None.

PR #344, opened 2026-05-05
feat: configuration enablement — env-var defaults + source resilience

dzivkovi:
Six small additive changes that make the skill correctly understand its
configured sources, plus tests + docs.

User-visible benefits

- LAST30DAYS_STORE=1 in .env turns persistence default-on without
  remembering --store on every invocation. Mirrors LAST30DAYS_DEBUG /
  LAST30DAYS_SKIP_PREFLIGHT convention.
- SCRAPE_CREATORS_API_KEY (with underscore) accepted as alias for the
  canonical name. Matches the spelling used in the vendor's own example
  code (Adrian Horning's repo); saves the next user the same diagnostic
  rabbit hole.
- Bluesky search now hits api.bsky.app (canonical AppView) instead of
  public.api.bsky.app (BunnyCDN-blocked public mirror as of 2026-05-04).
  BSKY_SEARCH_HOST env var lets users self-rescue future host migrations
  without a code release. Pre-fix: silent 0 Bluesky posts on every run.
- App-password format validator emits a one-shot stderr warning when
  BSKY_APP_PASSWORD doesn't match xxxx-xxxx-xxxx-xxxx form. Detect-don't-
  gate: createSession still accepts main passwords; the warning helps
  users identify a hygiene issue without breaking existing setups.
- Instagram retry on multi-token 500. SC's v2 reels endpoint wraps
  Google Search and 500's frequently on multi-word queries; a hashtag-
  form retry runs once before bubbling up. Documented vendor instability.
- LAST30DAYS_TRANSCRIPT_TIMEOUT env var (default 30s, was hardcoded 15s).
  SC's transcript endpoint regularly takes >15s; the old default was
  clipping legitimate responses.
- Silent-failure visibility: new bonus_errored field in the quality
  nudge fires when SC is configured but Instagram returned 0 items.
  Users see "Bonus source silent: Instagram" instead of unexplained
  absence.
- YouTube degraded-ratio false-positive fixed. Captions-disabled videos
  can never produce a transcript regardless of yt-dlp version; they're
  now subtracted from the denominator so a single uploader-disabled
  video doesn't false-trigger the "stale yt-dlp" nudge.
- urllib retry path: status_code attribute typo fix. The Instagram
  500-retry was dead code on the urllib branch (getattr(e, 'status', ...)
  while http.HTTPError exposes status_code).

Docs

- README.md: added /plugin install last30days step after marketplace add
  in three places (the install was previously omitted in the docs).
- CONFIGURATION.md: documented LAST30DAYS_STORE env var, added
  BSKY_SEARCH_HOST + app-password format section, mentioned
  LAST30DAYS_TRANSCRIPT_TIMEOUT in the Instagram source row.

Test plan

- 43 new unit tests across test_bluesky.py, test_instagram_sc.py,
  test_quality_nudge.py, test_youtube_yt.py
- 141 total tests passing in target suite
- Verified end-to-end: /last30days "Toronto resale condo market" with
  all 11+ sources active stored 35 new + 5 updated findings, all builder-
  PR-style accounts absent (organic agent voice in Instagram + TikTok
  results)

Backward compatibility

All changes are strictly additive. Optional kwargs default to None.
New env vars are opt-in. Existing CLI flags untouched. Existing callers
of public functions unaffected.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@ htmlcov/
 
 # Internal planning docs (ce:plan output) — keep local, don't publish
 docs/plans/
+.context/
 
 /work
 /print
-
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,7 @@ htmlcov/
 
 # Internal planning docs (ce:plan output) — keep local, don't publish
 docs/plans/
+
+/work
+/print
+
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -1,0 +1,256 @@
+# Configuration
+
+Everything you can tune in `/last30days` without editing the engine source.
+Three layers, in order of how often you'll touch them:
+
+1. **Per-run flags** - what you pass on the command line.
+2. **Environment variables and `.env`** - what's enabled across all runs.
+3. **Optional trend-monitoring stack** - SQLite store, watchlist, briefings.
+
+Per-client patterns and the experimental beta channel are at the bottom.
+
+> Skip ahead: [Where output is saved](#where-output-is-saved) - [API keys](#api-keys-env) - [Reasoning provider](#reasoning-provider-priority) - [Web search backend](#web-search-backend-priority) - [Trend monitoring](#trend-monitoring-store--watchlist--briefings) - [Per-client patterns](#per-client-patterns) - [Beta channel](#beta-channel)
+
+## Why this document exists
+
+This is a focused **configuration reference** maintained alongside the engine. The runtime contract (the voice rules, the planner protocol, the LAWs the synthesizing model follows) lives in [`skills/last30days/SKILL.md`](skills/last30days/SKILL.md) - that file is authoritative when the two ever differ. This file's job is narrower: surface every knob a user or operator can turn, in one place, kept current with the code so client-facing setups stay reliable. New configuration knobs added to the engine should be reflected here in the same PR.
+
+---
+
+## Where output is saved
+
+| Platform | Default path | Override |
+|---|---|---|
+| Linux / macOS | `~/Documents/Last30Days/` | `LAST30DAYS_MEMORY_DIR=/path` |
+| Windows | `C:\Users\<you>\Documents\Last30Days\` | `LAST30DAYS_MEMORY_DIR=C:\path` |
+
+Each run produces one file per topic, slug-named:
+`<slug>-raw[-suffix].md`. Same topic + same suffix on the same day overwrites; same topic + same suffix on different days appends a date stamp.
+
+**Per-run overrides:**
+- `--save-dir <path>` - one-off output location.
+- `--save-suffix <name>` - distinguish runs of the same topic (e.g. per client: `--save-suffix=acme`).
+
+The footer line `📎 Raw results saved to ~/Documents/Last30Days/<slug>-raw.md` is the canonical pointer; if it shows backslashes on Windows update past v3.1.1.
+
+---
+
+## API keys (`.env`)
+
+The skill reads keys from a `.env` file. Two locations are supported, in priority order:
+
+1. **`.claude/last30days.env`** in the current project directory (project-scoped) - takes precedence when present.
+2. **`~/.config/last30days/.env`** at the user level (global default) - the fallback.
+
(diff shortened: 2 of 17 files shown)


dzivkovi:
**Heads-up on PR stacking, Matt**

This PR's branch (`feat/config-enablement`) was built on top of my local `daniel/personal` which already includes the content of #338, #339, and #340. The diff against `main` therefore shows their changes too. Three options depending on your preference:

1. **Merge this first** — supersedes #338/#339/#340 (you'd close them). Single review, single merge, biggest PR.
2. **Merge #338/#339/#340 in your preferred order first** — I'll rebase this on top, the diff will then show only tonight's ~873 lines (configuration enablement + observability + tests).
3. **Cherry-pick** the tonight-only commits onto a fresh branch — if you'd prefer I redo the PR cleanly off `main`. Likely some conflicts with #339 (CONFIGURATION.md) which I'd resolve.

I lean (2) since you may already be planning specific merge order for the earlier PRs. Happy to do (3) if it makes the review easier — let me know.

The tonight-only delta in this PR (separable from the prior PRs):
- `LAST30DAYS_STORE` env var, `SCRAPE_CREATORS_API_KEY` alias, `BSKY_SEARCH_HOST` override
- Bluesky search host fix (canonical AppView, was BunnyCDN-blocked public mirror)
- Instagram retry on multi-token 500, transcript timeout config
- `bonus_errored` quality-nudge field for silent-failure visibility
- YouTube degraded-ratio denominator fix (captions-disabled videos)
- App-password format validator + hygiene warning
- README install-step doc fix (`/plugin install last30days` was missing)
- 43 new unit tests, 141/141 passing in target suite

tmchow:
Rebased onto main and pushed. Coherent bundle — the env-var defaults follow the existing `LAST30DAYS_DEBUG` / `SKIP_PREFLIGHT` convention so they slot in cleanly. The Bluesky host swap is the time-sensitive bit here: the CDN block of `public.api.bsky.app` is a hard breakage of the prior default, so that part can't wait. Conflict resolution: README install table and the new `Configuration` section both kept; instagram tests adapted to the unified `http.get` style on main; quality_nudge tests switched to the `lib.*` import path with `sys.path` setup; README/CONFIGURATION.md memory-dir lines reworded to "defaults to" form to satisfy `test_no_stray_hardcoded_memory_dir_paths`. Full suite is 1522 passed, 4 skipped.

tmchow:
Pushed a mechanical fix for both Greptile catches as 5c29613.

The first one was a false-positive Instagram silent-failure nudge: `last30days.py` always writes `instagram_items_count` into `research_results` (it pulls from `report.items_by_source.get("instagram") or []`), so a user who set `EXCLUDE_SOURCES=instagram` would intentionally skip the source and still get the "Bonus source silent" nudge. Fixed by gating `_is_instagram_silent_failure` on the canonical EXCLUDE_SOURCES parsing pattern from `pipeline.available_sources()` — if `instagram` is excluded, the silent-failure check returns False before looking at the count.

The second was the unused `BSKY_SEARCH_URL` module constant: `search_bluesky` only ever calls `_resolve_search_url(config)`, so the two regression-guard tests that asserted on the constant were guarding a value with no runtime effect. Removed the constant and rewrote the two tests to assert against `_resolve_search_url()`'s default return, which keeps the regression guard but anchors it at the code path the runtime actually uses.

86 tests pass (`pytest tests/test_quality_nudge.py tests/test_bluesky.py`).

tmchow:
Pushed mechanical follow-up for the symmetric case Greptile flagged: when `INCLUDE_SOURCES` is set to an opt-in allowlist that omits `instagram`, the silent-failure gate now returns False too — same parsing pattern as the `EXCLUDE_SOURCES=instagram` guard, just inverted. Tests added for allowlist-without-instagram (suppressed), allowlist-with-instagram (still fires), and empty/unset allowlist (still fires).

dzivkovi:
Thanks for both Greptile fixes and the rebase. Anchoring `_is_instagram_silent_failure` on `pipeline.available_sources()` keeps the source-name parsing in one place instead of duplicating it into the nudge layer; the `INCLUDE_SOURCES` symmetric follow-up was a good catch on top of Greptile's `EXCLUDE_SOURCES` one.

PR #347, opened 2026-05-06
docs: update search key file paths

dinakars777:
## Summary

Updates the `docs/how-search-works.md` Key Files table to match the current `skills/last30days/scripts/` layout.

## Changes

- Replace stale root `scripts/` paths with current skill paths.
- Swap retired module names for existing v3 modules such as `pipeline.py`, `providers.py`, and `relevance.py`.

## Testing

## Related Issues

None

### DIFF
--- a/docs/how-search-works.md
+++ b/docs/how-search-works.md
@@ -142,7 +142,7 @@ The repo vendors a search-only subset of Bird's Twitter GraphQL client and shell
 | Likes/reposts | Real (X API) | Real (x_search tool) |
 | Replies/quotes | Real | Real |
 | Author handle | Real | Real |
-| Relevance score | Default 0.7 (re-ranked by score.py) | AI-assessed 0.0-1.0 |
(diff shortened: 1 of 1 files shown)


tmchow:
Pushed a small mechanical fix to your branch (force-with-lease, rebased onto current `main`): swapped the leftover `score.py` reference on line 145 of `docs/how-search-works.md` to `relevance.py` so it lines up with the renamed module in your Key Files table.

Heads-up for a separate follow-up: the Reddit Search section's prose and the architecture diagram (lines ~12-13 and ~30-61) still describe the retired OpenAI Responses API path (`POST /v1/responses`, `gpt-5.2` fallback chain, etc.), even though Key Files now correctly points to `reddit_public.py` (public Reddit JSON, no API key). I left that for a follow-up PR since it's more than a terminology swap — it needs a real section rewrite (and a diagram update) to match the current `reddit_public.py` flow, and I didn't want to expand the scope of this docs PR without your sign-off.

Thanks for the cleanup!

PR #349, opened 2026-05-06
chore: sync gemini extension version

dinakars777:
## Summary

Updates `gemini-extension.json` so its version matches the current project and plugin manifests, and adds it to the existing version contract test.

## Changes

- Change Gemini extension metadata from `3.0.5` to `3.2.4`. (Retargeted during rebase — PR was originally opened against `3.1.1` but main moved to `3.2.4` before merge. The contract test below is the durable fix; it'll catch future drift regardless of literal value.)
- Assert the Gemini extension version matches the project version in `tests/test_plugin_contract.py`.

## Testing

- [ ] Ran `bash skills/last30days/scripts/sync.sh` (not needed: metadata/test-only change)

## Related Issues

None

### DIFF
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -1,6 +1,6 @@
 {
   "name": "last30days-skill",
-  "version": "3.0.5",
+  "version": "3.2.4",
   "description": "Research a topic from the last 30 days across Reddit, X, YouTube, TikTok, Instagram, Hacker News, Polymarket, and the web.",
   "settings": [
     {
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks for catching this drift, Dinakar — and nice call adding the contract test, that's the part that earns its keep long-term. I retargeted the version bump from 3.1.1 to 3.2.4 on rebase since main moved on while this was open, so the new `test_versions_match_across_manifests` passes against current main. The test itself is unchanged and will catch any future drift between `plugin.json` and `gemini-extension.json` automatically.

PR #352, opened 2026-05-06
docs: mark root spec as historical

dinakars777:
## Summary

Adds a note to `SPEC.md` that it is historical and points readers to the maintained runtime contract.

## Changes

- Mark `SPEC.md` as historical.
- Point to `skills/last30days/SKILL.md` for current runtime behavior.

## Testing

## Related Issues

None

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,5 +1,8 @@
 # last30days Skill Specification
 
+> Historical implementation spec. The maintained runtime contract lives in
+> `skills/last30days/SKILL.md`.
+
 ## Overview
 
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks @dinakars777 — going to take a different path here. `SPEC.md` has no inbound references anywhere on current main (checked README, SKILL.md, AGENTS.md, CHANGELOG, and docs/), so the cleaner move is to delete it outright rather than annotate it as historical. Git history preserves the content if anyone needs to dig it up. Landing the deletion in a small follow-up PR alongside `TASKS.md` (same reasoning, see #353). Appreciate the diligence in surfacing the drift — that's how this gets resolved either way.

PR #353, opened 2026-05-06
docs: mark root tasks as historical

dinakars777:
## Summary

Adds a note to `TASKS.md` that it is a historical checklist and points readers to the current runtime code location.

## Changes

- Mark `TASKS.md` as historical.
- Point to `skills/last30days/scripts/` for current runtime code.

## Testing

## Related Issues

None

### DIFF
--- a/TASKS.md
+++ b/TASKS.md
@@ -1,5 +1,8 @@
 # last30days Implementation Tasks
 
+> Historical checklist from the original implementation. Current runtime code
+> lives under `skills/last30days/scripts/`.
+
 ## Setup & Configuration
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks @dinakars777 — same reasoning as #352. `TASKS.md` is also fully orphaned on current main (no inbound references in README, SKILL.md, AGENTS.md, CHANGELOG, or docs/), so the cleaner move is to delete it outright rather than annotate it as historical. Git history preserves the checklist content. Both deletions are landing together in a small follow-up PR.

PR #354, opened 2026-05-06
fix: route parallel web backend through grounding

dinakars777:
## Summary

Ensures explicit `--web-backend=parallel` enables the `grounding` source like the other explicit web backends.

## Changes

- Include `parallel` in the explicit web-backend source activation path.
- Add a regression test that verifies the Parallel backend reaches grounding and reports the missing `PARALLEL_API_KEY` from the dispatcher instead of failing early with no sources.

## Testing

- [x] Ran `uv run pytest tests/test_pipeline_v3.py tests/test_plugin_contract.py tests/test_version_consistency.py`
- [ ] Ran `bash skills/last30days/scripts/sync.sh` (not run; source change prepared for PR review)

## Related Issues

None

### DIFF
--- a/skills/last30days/scripts/lib/pipeline.py
+++ b/skills/last30days/scripts/lib/pipeline.py
@@ -203,7 +203,7 @@ def run(
             available = [source for source in available if source in requested_sources]
     if web_backend == "none":
         available = [s for s in available if s != "grounding"]
-    elif web_backend in ("brave", "exa", "serper") and "grounding" not in available:
+    elif web_backend in ("brave", "exa", "serper", "parallel") and "grounding" not in available:
         available.append("grounding")
     if not available:
         raise RuntimeError("No sources are available for this run.")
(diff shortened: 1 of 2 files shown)


tmchow:
@dinakars777 mechanical follow-up: I relaxed the second `assertIn` on `test_parallel_web_backend_enables_grounding_source` to drop the exact `PARALLEL_API_KEY` substring match and keep only the stable `assertIn("grounding", report.errors_by_source)` check — that way the test stays green if `grounding.py`'s missing-key error message ever gets reworded.

I deliberately did **not** add `mock=True` to that test (a separate review suggestion): with `mock=True`, `_retrieve_stream` short-circuits to `_mock_stream_results` and the grounding error path never fires, which would break the very contract this test exists to assert. The non-mock path is safe here because `LAST30DAYS_REASONING_PROVIDER=auto` with no keys falls through to the local/deterministic runtime in `providers.resolve_runtime` rather than raising, so CI won't blow up on missing creds. Pushed as `ec0b126` on your branch.

PR #355, opened 2026-05-06
test: cover parallel grounding backend

dinakars777:
## Summary

Adds test coverage for the Parallel web-search grounding backend.

## Changes

- Cover `parallel_search` response parsing and date filtering.
- Cover auto-dispatch when only `PARALLEL_API_KEY` is configured.
- Cover explicit `backend=parallel` missing-key validation.

## Testing

- [x] Ran `uv run pytest tests/test_grounding_v3.py tests/test_plugin_contract.py tests/test_version_consistency.py`
- [ ] Ran `bash skills/last30days/scripts/sync.sh` (not needed: test-only change)

## Related Issues

None

### DIFF
--- a/tests/test_grounding_v3.py
+++ b/tests/test_grounding_v3.py
@@ -130,13 +130,13 @@ class ParallelSearchTests(unittest.TestCase):
                     "title": "Parallel Result",
                     "url": "https://example.com/parallel",
                     "snippet": "A parallel snippet",
-                    "published_date": "2026-03-15T00:00:00Z",
+                    "publish_date": "2026-03-15T00:00:00Z",
                 },
                 {
                     "title": "Old Parallel Result",
                     "url": "https://example.com/old-parallel",
                     "snippet": "Should be filtered",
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks @dinakars777 - good catches from Greptile. I went ahead and pushed both mirror tests (`test_auto_prefers_serper_over_parallel` and `test_parallel_search_returns_empty_for_no_results`) to your branch following the existing patterns from the brave/exa siblings.

tmchow:
Pushed f2737fc to fix the mock key mismatch Greptile flagged: the test had `published_date` but `parallel_search` in `grounding.py:156` reads `publish_date`, so every row's `pub_date` was `None` and `_in_date_range` filtered them all out — the assertion would have failed. With the key corrected, `test_parallel_search_filters_to_in_range_dated_items` now actually exercises the date-filter path (in-range row kept, old + undated rows dropped). Full `tests/test_grounding_v3.py` suite: 26 passed.

PR #356, opened 2026-05-06
fix: honor explicit optional source requests

dinakars777:
## Summary

Allows explicitly requested optional sources to pass validation and availability checks when their credentials are configured.

## Changes

- Add `threads` and `pinterest` to the accepted `--search` source list.
- Allow explicit `requested_sources=["threads"]` when `SCRAPECREATORS_API_KEY` is configured, even without `INCLUDE_SOURCES=threads`.
- Allow explicit `requested_sources=["perplexity"]` when `OPENROUTER_API_KEY` is configured, even without `INCLUDE_SOURCES=perplexity`.
- Add parser and availability coverage for these optional sources.

## Testing

- [x] Ran `uv run pytest tests/test_cli_v3.py tests/test_plugin_contract.py tests/test_version_consistency.py`
- [ ] Ran `bash skills/last30days/scripts/sync.sh` (not run; source change prepared for PR review)

## Related Issues

None

### DIFF
--- a/skills/last30days/scripts/lib/pipeline.py
+++ b/skills/last30days/scripts/lib/pipeline.py
@@ -126,9 +126,7 @@ def available_sources(config: dict[str, Any], requested_sources: list[str] | Non
         available.append("perplexity")
     if requested_sources and "xiaohongshu" in requested_sources and env.is_xiaohongshu_available(config):
         available.append("xiaohongshu")
-    if env.is_threads_available(config) or (
-        requested_sources and "threads" in requested_sources and config.get("SCRAPECREATORS_API_KEY")
-    ):
+    if env.is_threads_available(config):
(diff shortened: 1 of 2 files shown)


tmchow:
Pushed a small refactor on top (`261ea58`) cleaning up dead code in the threads branch: both operands of the `or` in `available_sources` were checking `SCRAPECREATORS_API_KEY` (`is_threads_available(config)` is exactly `bool(config.get('SCRAPECREATORS_API_KEY'))`), so the second clause was unreachable. Your real fixes — adding `pinterest`/`threads` to `MOCK_AVAILABLE_SOURCES` and the explicit-perplexity branch — are intact, and `test_explicit_threads_search_uses_scrapecreators_key_without_include_sources` still passes.

PR #357, opened 2026-05-06
fix: skip POSIX secret warning on Windows

### LINKED ISSUE #325: Windows: `_check_file_permissions` in `scripts/lib/env.py` reports false-positive warning regardless of NTFS ACL
Hi Matt — first, thanks for this skill. It's been a high-signal addition to my workflow.

## Bug

`skills/last30days/scripts/lib/env.py`, function `_check_file_permissions` (lines 54-67 in v3.1.1, identical on `main` at time of writing) checks `.env` permissions using POSIX mode bits:

```python
mode = path.stat().st_mode
if mode & 0o044:
    sys.stderr.write(f"[last30days] WARNING: {path} is readable by other users. Run: chmod 600 {path}\n")
```

On Windows NTFS, `path.stat().st_mode` returns synthesized POSIX bits derived only from the read-only attribute — it does **not** reflect actual NTFS ACLs. As a result, even after hardening the file with `icacls /inheritance:r /grant:r "%USERNAME%:F"` (and `icacls` confirming only the owner has access), the warning still fires on every invocation. There is no ACL change a Windows user can make to silence it.

## Reproduction (Windows 11, NTFS)

1. Create `%USERPROFILE%\.config\last30days\.env` with an `OPENAI_API_KEY`.
2. Harden: `icacls "%USERPROFILE%\.config\last30days\.env" /inheritance:r /grant:r "%USERNAME%:F"`. Verify `icacls` output shows only the owner.
3. Run any `last30days` command. The warning still fires.

## Suggested fix

Two options:

1. **Pragmatic:** early-return on Windows — `if os.name == 'nt': return` at the top of `_check_file_permissions`. Trade-off: no permission warning on Windows at all.
2. **Proper:** Windows-aware check via `subprocess.run(["icacls", str(path)], capture_output=True, text=True)`, parsing for non-owner principals (`Everyone`, `BUILTIN\Users`, `Authenticated Users`). More work but actually checks real permissions.

Option 1 is the minimum viable fix and removes the spurious noise; option 2 is the right long-term answer.

Happy to provide more diagnostic output if helpful — let me know.

— Victor Johansson, Trinordic AB

dinakars777:
Windows no longer gets a repeated `.env` permission warning from synthesized POSIX mode bits. `_check_file_permissions` now leaves Windows ACL validation alone instead of suggesting `chmod 600`, and the regression test verifies the Windows path does not stat or warn.

Fixes mvanhorn/last30days-skill#325.

Tests: `uv run pytest tests/test_env_v3.py tests/test_plugin_contract.py tests/test_version_consistency.py`

---

### DIFF
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -70,6 +70,10 @@ class OpenAIAuth:
 
 def _check_file_permissions(path: Path) -> None:
     """Warn to stderr if a secrets file has overly permissive permissions."""
+    if os.name == "nt":
+        # Windows reports synthesized POSIX mode bits that do not reflect NTFS ACLs.
+        return
+
     try:
         mode = path.stat().st_mode
         # Check if group or other can read (bits 0o044)
--- a/tests/test_env_v3.py
+++ b/tests/test_env_v3.py
@@ -41,6 +41,14 @@ class EnvV3Tests(unittest.TestCase):
             with mock.patch.dict(os.environ, {}, clear=False):
                 self.assertIsNone(bird_x.is_bird_authenticated())
 
+    def test_file_permission_check_skips_windows_posix_mode_bits(self):
+        path = mock.Mock(spec=Path)
+        with mock.patch.object(env.os, "name", "nt"), mock.patch.object(env.sys.stderr, "write") as write:
+            env._check_file_permissions(path)
+
+        path.stat.assert_not_called()
+        write.assert_not_called()
+
 
 if __name__ == "__main__":
     unittest.main()


tmchow:
Thanks @dinakars777 — clean small fix targeting a real Windows wart. Python's \`Path.stat().st_mode\` returns synthesized POSIX bits on NTFS that don't actually reflect file ACLs, so the "chmod 600" suggestion was both wrong advice and noisy on every Windows SessionStart. Early-return on \`os.name == "nt"\` is the right shape — leaves the existing POSIX check intact, no behavior change for Linux/macOS.

The test mocks \`os.name\` to \`"nt"\` and asserts both \`stat\` and \`stderr.write\` are never called, which is the right contract — it would catch any future refactor that accidentally re-enables the check.

Rebased onto current \`main\` (was ~46 commits behind, no conflicts).

PR #358, opened 2026-05-06
fix: initialize OpenClaw poll timing once

dinakars777:
OpenClaw device-auth polling now initializes its deadline and reminder timer from the same timestamp. This keeps the polling clock model simpler and fixes the brittle timeout/expired-token test paths that expected one start time followed by loop checks.

Tests: `uv run pytest tests/test_setup_openclaw.py tests/test_plugin_contract.py tests/test_version_consistency.py`

---

### DIFF
--- a/skills/last30days/scripts/lib/setup_wizard.py
+++ b/skills/last30days/scripts/lib/setup_wizard.py
@@ -335,8 +335,9 @@ def poll_device_auth(
     """
     import sys
 
-    deadline = time.time() + timeout
-    last_reminder = time.time()
+    started_at = time.time()
+    deadline = started_at + timeout
+    last_reminder = started_at
     reminder_count = 0
     max_reminders = 4
     reminder_interval = 30  # seconds between reminders
--- a/tests/test_setup_openclaw.py
+++ b/tests/test_setup_openclaw.py
@@ -213,10 +213,10 @@ class TestPollDeviceAuth:
     @patch("lib.setup_wizard.urlopen")
     def test_timeout_returns_none(self, mock_urlopen, mock_time):
         """Returns None when timeout is exceeded."""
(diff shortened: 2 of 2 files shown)


tmchow:
Thanks @dinakars777 — clean small fix. Capturing \`started_at\` once and deriving both \`deadline\` and \`last_reminder\` from it is correct, and it sidesteps the subtle "two \`time.time()\` calls one line apart can return different microsecond values" race in the original code.

One thing worth flagging because it overlaps with PR #416 (also currently labeled ready-for-maintainer): #416 ran into the same \`TestPollDeviceAuth\` failure and fixed it by padding the \`mock_time.time\` \`side_effect\` lists to handle the extra \`time.time()\` call the impl had picked up after the tests were written. That fix is the test-side patch; this PR is the impl-side cleanup. They're complementary rather than conflicting:

- If both land, the padded mocks in #416 will have one extra \`side_effect\` entry that's no longer consumed - harmless, just unused.
- The cleaner impl in this PR is the better long-term state, so #358-then-#416 or #416-then-#358 both work.

37/37 \`tests/test_setup_openclaw.py\` tests pass on the rebased branch. Rebased onto current \`main\` (was ~46 commits behind, no conflicts).

tmchow:
Quick follow-up @dinakars777 — added a one-line companion commit to `tests/test_setup_openclaw.py`. Once #416's test-side padding (`side_effect=[0, 0, 301]`) landed on main, the rebased branch needed one less mock value to match this PR's collapsed `time.time()` call pattern: `started_at` is captured once, while-check reads the second value past the deadline. Dropped to `side_effect=[0, 301]` and updated the comment block to describe the new sequence. 37/37 setup_openclaw tests now pass and the PR flips to MERGEABLE.

PR #359, opened 2026-05-06
fix: persist raw items when rankings are absent

dinakars777:
Watchlist persistence now keeps raw `items_by_source` findings when a report has no ranked candidates. Ranked reports keep the existing high-quality candidate path plus the HN/Polymarket supplement, while mock/headless reports no longer drop Reddit, X, or other raw-source findings.

Tests: `uv run pytest tests/test_store.py tests/test_watchlist_commands.py tests/test_plugin_contract.py tests/test_version_consistency.py`

---

### DIFF
--- a/skills/last30days/scripts/store.py
+++ b/skills/last30days/scripts/store.py
@@ -687,11 +687,16 @@ def findings_from_report(
         findings.append(finding)
         seen_urls.add(candidate.url)
     
-    # Phase 2: Add HN/PM items not already captured in ranked candidates
-    for source_name in ["hackernews", "polymarket"]:
-        if source_name not in report.items_by_source:
-            continue
-        for item in report.items_by_source[source_name]:
+    # Phase 2: Add raw items not already captured in ranked candidates.
+    # When rankings are present, keep the historical HN/PM supplement behavior.
+    # When rankings are absent, preserve watchlist/mock runs from every source.
+    raw_source_names = (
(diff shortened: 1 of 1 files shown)


tmchow:
Thanks @dinakars777 — your diagnosis was exactly right. The bug is real: \`store.findings_from_report\` was hard-gated to "supplement HN/PM only," so when \`ranked_candidates\` came back empty (mock runs, headless runs, runs where rerank was skipped), every other source got silently dropped from watchlist persistence.

Closing this one as superseded rather than merging, because PR #416 ([\"fix(ci): run full pytest suite, repair 13 rotted tests\"](https://github.com/mvanhorn/last30days-skill/pull/416)) lands the equivalent fix - same \`not report.ranked_candidates\` gate, same fallback to all \`items_by_source\` sources when rankings are empty, same preservation of HN/PM supplement when rankings are present. The formulation differs in surface detail (\`list(items_by_source)\` vs \`items_by_source.keys()\`, etc.) but the semantics are identical, and #416 bundles this fix alongside three other rotted-test repairs and a CI workflow change that broadens default coverage past the prior 2-test sweep. Keeping that bundle together preserves the story of why these failures were hiding (the workflow only ran a narrow slice) rather than landing each repair as its own one-line change.

Two independent identifications of the same bug is the strongest evidence the bug needed fixing, though — appreciated. Keep them coming.

PR #364, opened 2026-05-08
fix(hooks): replace unsafe eval with declare in check-config.sh

### LINKED ISSUE #361: SessionStart hook executes shell commands from .env values via unsafe eval
## Summary

`hooks/scripts/check-config.sh` loads `~/.config/last30days/.env` (or `.claude/last30days.env`) using `eval`, which executes any shell command substitution (`` `...` `` or `$()`) present in a value — even when that substitution appears inside what looks like an inline comment.

In my case, the `.env` file generated by the setup wizard included this line (from the project's own example/template):

```
OPENAI_API_KEY=sk-...      # optional - legacy Reddit fallback if using `codex login`
```

The plugin's `SessionStart` hook ran on every Claude Code / Claude Desktop session start and silently invoked `codex login`, which opened my browser to the OpenAI OAuth flow. Because the OAuth callback to `localhost:1455` raced with the brief hook process, login never completed, and the prompt re-fired forever.

## Root cause

`hooks/scripts/check-config.sh:37` (3.1.1):

```bash
eval "ENV_${key}=\"${value}\""
```

`bash`'s `read -r key value` does not treat `#` as a comment delimiter, so the entire remainder of the line — including the backticks — is captured into `$value`. `eval` then re-parses the whole assignment, and command substitution inside the double-quoted value fires.

Reproduction:

```bash
cat > /tmp/repro.env <<'EOF'
FOO=bar # this comment has `touch /tmp/pwned` in it
EOF

# Mimic the hook's loader
while IFS='=' read -r key value; do
  [[ "$key" =~ ^[[:space:]]*# ]] && continue
  [[ -z "$key" ]] && continue
  key=$(echo "$key" | xargs)
  value=$(echo "$value" | xargs | sed 's/^["'\''']//;s/["'\''']$//')
  eval "ENV_${key}=\"${value}\""
done < /tmp/repro.env

ls -la /tmp/pwned   # exists — backticks executed
```

## Impact

This is a security issue, not just a UX bug. `eval` of an arbitrary string read from disk means any command substitution in a values file (or in a comment that ends up captured as part of a value) executes with the privileges of the user running Claude. A malicious or accidental backtick or `$(...)` anywhere downstream of the first `=` runs.

The repository's own example/template appears to ship with backticks-in-comments syntax, so this bug is reachable without any user action beyond running the setup wizard.

## Suggested fix

Drop the `eval` entirely. The values are only needed for inspection (counting active sources, etc.), so a non-eval parse is straightforward. Two safer options:

1. Source the file in a subshell and read variables back out (still risky for arbitrary content but avoids reading inline-comment text into values):

   ```bash
   while IFS= read -r line; do
     [[ "$line" =~ ^[[:space:]]*# ]] && continue
     [[ -z "$line" ]] && continue
     # strip inline comments and whitespace before assignment
     line="${line%%#*}"
     [[ "$line" =~ ^([A-Za-z_][A-Za-z0-9_]*)=(.*)$ ]] || continue
     key="${BASH_REMATCH[1]}"
     value="${BASH_REMATCH[2]}"
     # trim whitespace and surrounding quotes
     value="${value#"${value%%[![:space:]]*}"}"
     value="${value%"${value##*[![:space:]]}"}"
     value="${value#\"}"; value="${value%\"}"
     value="${value#\'}"; value="${value%\'}"
     printf -v "ENV_${key}" '%s' "$value"
   done < "$file"
   ```

2. Or, since the hook only checks presence/absence of a handful of known keys, replace the loop with `grep -E '^KEY=' file | head -1` per key. Simpler, no parser needed.

## Environment

- Plugin version: 3.1.1
- Claude Code: 2.1.131
- Claude Desktop: latest as of 2026-05-06
- macOS 15 (Darwin 25.3.0)

## Workaround for affected users

Edit `~/.config/last30days/.env` and remove any inline comments containing backticks or `$(...)`. Placeholder lines like `OPENAI_API_KEY=sk-... # ... \`codex login\`` should be deleted entirely if you don't have a real value for that key.

davemorin:
## Summary

Replace `eval` with `declare` in `load_env_vars()` and strip inline comments before variable assignment. This prevents command substitution from backtick-containing comments in `.env` files.

## Changes

- `hooks/scripts/check-config.sh`: Strip inline comments (`# ...`) from values before assignment using `${value%%[[:space:]]#*}`
- Replace `eval "ENV_${key}=\"${value}\""` with `declare "ENV_${key}=${value}"` to prevent shell expansion

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short` (pre-existing failures only, no regressions)

Tested with a `.env` containing:
```
FOO=bar      # this comment has `touch /tmp/pwned` in it
```
Before: `eval` executes `touch /tmp/pwned` via backtick expansion on every SessionStart.
After: `declare` assigns `bar` without command substitution, inline comment stripped.

## Related Issues

Fixes #361

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -37,7 +37,9 @@ load_env_vars() {
       # command substitution in backtick-containing comments
       value="${value%%[[:space:]]#*}"
       if [[ -n "$key" && -n "$value" ]]; then
-        declare -g "ENV_${key}=${value}"
+        # printf -v writes via assignment semantics (global from inside a
+        # function), works on macOS's /bin/bash 3.2 — `declare -g` is 4.2+.
+        printf -v "ENV_${key}" '%s' "$value"
       fi
     done < "$file"
   fi
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -33,8 +33,11 @@ load_env_vars() {
       [[ -z "$key" ]] && continue
       key=$(echo "$key" | xargs)
       value=$(echo "$value" | xargs | sed 's/^["'\''"]//;s/["'\''"]$//')
+      # Strip inline comments (# preceded by whitespace) to prevent
+      # command substitution in backtick-containing comments
+      value="${value%%[[:space:]]#*}"
       if [[ -n "$key" && -n "$value" ]]; then
-        eval "ENV_${key}=\"${value}\""
+        declare "ENV_${key}=${value}"
       fi
     done < "$file"
   fi
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -37,7 +37,7 @@ load_env_vars() {
       # command substitution in backtick-containing comments
       value="${value%%[[:space:]]#*}"
       if [[ -n "$key" && -n "$value" ]]; then
-        declare "ENV_${key}=${value}"
+        declare -g "ENV_${key}=${value}"
       fi
     done < "$file"
   fi


tmchow:
Thanks Dave — the \`eval\` was a real foot-gun, and the empirical repro you included (\`FOO=bar # comment with \\\`touch /tmp/pwned\\\`\`) is exactly the kind of "looks harmless, runs arbitrary commands" surface that lives in hook code until someone notices.

Verified the fix empirically by running the patched \`load_env_vars\` against a malicious \`.env\` containing both an in-comment backtick payload and an in-value backtick payload — neither pwn-marker file was created. The \`bash -n\` syntax check passes and the hook continues to drive the source-count message normally end-to-end.

Two-part fix is correctly scoped: the comment-strip (\`\${value%%[[:space:]]#*}\`) keeps user-friendly \`.env\` ergonomics (people do write \`KEY=value # what this is for\`), and \`declare\` is the right replacement for \`eval\` because it builds the variable assignment without re-parsing the value as shell. Together they close both the comment-payload and value-payload vectors.

Rebased onto current \`main\` (the branch was ~46 commits behind, no conflicts).

tmchow:
Follow-up @davemorin — Greptile caught a real scope issue the earlier empirical eval-fix test missed. Without `-g`, `declare` inside `load_env_vars` creates function-local variables, so every `ENV_*` value gets discarded on return and `.env`-based config loading silently breaks for the caller. Pushed a fixup commit (`726d232`) that switches the assignment to `declare -g`; reran the malicious-comment .env test and confirmed both the security fix (no backtick execution) and the scope fix (${ENV_FOO} reaches the caller as expected) are now passing empirically.

tmchow:
Greptile's follow-up flagged a real macOS portability issue: `declare -g` is bash 4.2+, but macOS ships `/bin/bash` at 3.2 (still — held back since the GPLv3 licence change). With `#!/bin/bash` + `set -euo pipefail` at the top of this script, the previous fixup would have aborted every SessionStart on a Mac with `declare: -g: invalid option` — silently disabling the .env loader the patch was meant to repair.

Switched to `printf -v "ENV_${key}" '%s' "$value"`. `printf -v` exists since bash 3.1 and writes via plain assignment semantics — global from inside a function (unlike `declare`, which goes function-local without `-g`). Empirically verified on `/bin/bash 3.2.57`: ENV_FOO/ENV_BAZ resolve at caller scope, and the original repro (`FOO=bar # comment with \`touch /tmp/pwned\``) still produces no `/tmp/pwned`. Three concerns now closed: command-substitution risk, function-local scope, and bash-version dependency.

PR #365, opened 2026-05-08
fix(version): replace hardcoded v3.0.0 with dynamic _skill_version()

### LINKED ISSUE #284: Version metadata drift across SKILL.md, plugin.json, and rendered output (Hermes/Codex installs show ?, 3.0.1, and 3.0.0 on v3.0.9)
## Summary

On the current `v3.0.9` checkout, version metadata appears to drift across multiple sources:

- `.claude-plugin/plugin.json` says `3.0.9`
- root `SKILL.md` still says `3.0.1`
- rendered compact output still prints `# last30days v3.0.0: ...`
- non-Claude installs (Hermes at least) show badge `🌐 last30days v? · synced ...`

This looks like three separate version sources falling out of sync.

Related older issue: #190. That one was about the earlier v2→v3 rename/metadata transition. This report is about the current `v3.0.9` state still exposing drift at runtime.

## What I observed

Using current upstream `origin/main` / tag `v3.0.9`:

### 1) Plugin metadata is 3.0.9

`.claude-plugin/plugin.json`
```json
{
  "name": "last30days",
  "version": "3.0.9"
}
```

### 2) Root SKILL.md is still 3.0.1

`SKILL.md`
```yaml
version: "3.0.1"
```

and later in the file:

```md
# last30days v3.0.1: Research Any Topic from the Last 30 Days
```

### 3) Rendered compact output title is hardcoded to 3.0.0

`scripts/lib/render.py`
```py
lines = [
    *_render_badge(),
    f"# last30days v3.0.0: {report.topic}",
```

There are other stale `v3.0.0` literals too, e.g.:

- `scripts/last30days.py` docstring
- `scripts/lib/ui.py`

### 4) Hermes/non-plugin installs show `v?` in the badge

`scripts/lib/render.py` currently resolves the badge version by walking upward looking only for:

```py
.claude-plugin/plugin.json
```

If that file is not present, it returns `"?"`.

That means non-Claude installs (Hermes, likely some Codex/manual installs too) get:

```md
🌐 last30days v? · synced YYYY-MM-DD
```

even though the code is actually from `v3.0.9`.

## Why this matters

For Hermes/manual installs specifically, `scripts/sync.sh` copies the root `SKILL.md` into the target:

```bash
cp "$SRC/SKILL.md" "$HERMES_TARGET/SKILL.md"
```

So Hermes ends up with:

- `SKILL.md` version = `3.0.1`
- badge version = `?`
- rendered report title = `3.0.0`
- actual runtime/features = `3.0.9`

The runtime works, but the version reporting is confusing and makes it hard to tell what’s actually installed.

## Repro

1. Checkout current `v3.0.9`
2. Run:
   ```bash
   bash scripts/sync.sh
   python3.12 ~/.hermes/skills/research/last30days/scripts/last30days.py "Kanye West" --quick --emit=compact
   ```
3. Observe output beginning with:
   ```md
   🌐 last30days v? · synced ...
   # last30days v3.0.0: Kanye West
   ```

## Likely fix direction

Seems like there should be one canonical runtime version source, and all of these should derive from it:

- root `SKILL.md`
- `.claude-plugin/plugin.json`
- rendered badge/title
- sync target/versioned cache paths
- tests

At minimum, these look out of sync today:

- `SKILL.md` still at `3.0.1`
- `render.py` still hardcoded to `v3.0.0`
- badge version lookup too Claude-specific for Hermes/manual installs

## Note on tests

Current tests seem to bless some of the stale literals, e.g. expecting `v3.0.0` in rendered output / CLI docstrings, so this drift may not fail CI right now.

davemorin:
## Summary

Replace hardcoded `v3.0.0` version strings in render.py, ui.py, and last30days.py with dynamic calls to `_skill_version()`, which reads the version from plugin.json at runtime.

## Changes

- `render.py`: Replace 3 hardcoded `v3.0.0` title strings (lines 83, 589, 777) with `_skill_version()` calls
- `ui.py`: Import `_skill_version` from render module; replace 2 hardcoded `v3.0.0` strings in the source status box with dynamic version + computed padding
- `last30days.py`: Remove version from module docstring (was `v3.0.0`, now version-free)
- `test_render_v3.py`, `test_cli_v3.py`: Update assertions to check for version prefix without a hardcoded number

**Before:** plugin.json says 3.1.1, but compact output title says `v3.0.0`, source status box says `v3.0.0`
**After:** all runtime version displays use `_skill_version()` which reads plugin.json

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short` (41 affected tests pass, pre-existing failures unrelated)

## Related Issues

Fixes #284

### DIFF
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # ruff: noqa: E402
-"""last30days v3.0.0 CLI."""
+"""last30days CLI."""
 
 from __future__ import annotations
 
--- a/skills/last30days/scripts/lib/render.py
+++ b/skills/last30days/scripts/lib/render.py
@@ -96,7 +96,7 @@ def render_compact(report: schema.Report, cluster_limit: int = 8, fun_level: str
     non_empty = [s for s, items in sorted(report.items_by_source.items()) if items]
     lines = [
         *_render_badge(),
-        f"# last30days v3.0.0: {report.topic}",
+        f"# last30days v{_skill_version()}: {report.topic}",
         "",
         *_assistant_safety_lines(),
         f"- Date range: {report.range_from} to {report.range_to}",
@@ -602,7 +602,7 @@ def render_comparison_multi(
 
     lines: list[str] = [
         *_render_badge(),
-        f"# last30days v3.0.0: {synthesized_topic}",
+        f"# last30days v{_skill_version()}: {synthesized_topic}",
         "",
         *_assistant_safety_lines(),
         f"- Comparison mode: {len(entities)} entities ({', '.join(entities)})",
@@ -790,7 +790,7 @@ def render_full(report: schema.Report) -> str:
     # Start with the same header as compact
     non_empty = [s for s, items in sorted(report.items_by_source.items()) if items]
     lines = [
-        f"# last30days v3.0.0: {report.topic}",
+        f"# last30days v{_skill_version()}: {report.topic}",
         "",
         *_assistant_safety_lines(),
         f"- Date range: {report.range_from} to {report.range_to}",
--- a/skills/last30days/scripts/lib/ui.py
+++ b/skills/last30days/scripts/lib/ui.py
@@ -6,6 +6,8 @@ import threading
 import random
 from typing import Optional
 
+from .render import _skill_version
+
 # Check if we're in a real terminal (not captured by Claude Code)
 IS_TTY = sys.stderr.isatty()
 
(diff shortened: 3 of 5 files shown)


tmchow:
Thanks Dave — good catch. The hardcoded \`v3.0.0\` strings were drift waiting to happen and showing the wrong number to users on every install past 3.0.x.

Verified empirically that \`_skill_version()\` now returns the right value (3.2.4 on current main) and 41/41 affected tests pass. One note: since this PR was opened, the version-source helper was refactored — \`_skill_version()\` now reads SKILL.md frontmatter (via \`lib/skill_meta.read_skill_version()\`) rather than plugin.json directly. Behavior is the same (single source of truth, dynamic at runtime), the helper just moved one indirection deeper. The rebase picked that up cleanly with no conflicts.

Rebased onto current \`main\` (the branch was ~46 commits behind).

PR #366, opened 2026-05-08
feat(web): auto-enrich Reddit URLs from web search via JSON API

### LINKED ISSUE #324: Reddit body content blocked when WebFetch runs inside Claude Code
When `/last30days` runs inside Claude Code (Anthropic's CLI), the host's `WebFetch` tool refuses Reddit URLs (`Claude Code is unable to fetch from www.reddit.com`). This silently degrades retrieval quality on any topic with strong Reddit presence — the planner emits Reddit URLs, but body content + comments never make it into the synthesis stage.

Ran on topic: *"claude code agentic OS"*, 2026-04-26. 5 of the top-15 retrieval URLs were Reddit; all 5 failed. Fell back to using mgrep titles + match-scores as a sentiment proxy, which loses comment-level quotes and full thread context.

## Suggestions

1. Add a Reddit fallback path — Reddit's public JSON endpoints (`reddit.com/r/<sub>/comments/<id>.json`) work without auth and aren't subject to the WebFetch host filter. Could route Reddit URLs through a small adapter at the retrieve stage.
2. Or use Firecrawl scrape via Actions if the skill already depends on Firecrawl elsewhere.
3. At minimum, document the limitation in `SKILL.md` so users know to expect it and can supplement Reddit retrieval manually.

## Environment

- Claude Code v2.1.119
- macOS (darwin-arm64)
- last30days-skill v3.0.10

davemorin:
## Summary

Auto-enrich Reddit URLs that appear in web search results (Brave, Exa, Serper, Parallel) by fetching thread body and top comments via Reddit's public JSON API. This bypasses Claude Code's WebFetch restriction on reddit.com.

## Changes

- `grounding.py`: Refactored `web_search()` to capture results and pass them through `_enrich_reddit_items()` before returning
- `grounding.py`: Added `_enrich_reddit_items()` which detects Reddit thread URLs in web search results and fetches body text (up to 2000 chars) + top 5 comments via `reddit_enrich.fetch_thread_data()`
- Non-Reddit URLs are untouched; enrichment failures are logged to stderr and don't disrupt the search

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short` (no regressions)

## Related Issues

Fixes #324

### DIFF
--- a/skills/last30days/scripts/lib/grounding.py
+++ b/skills/last30days/scripts/lib/grounding.py
@@ -232,9 +232,20 @@ def web_search(
         raise ValueError(f"Unsupported web backend: {backend!r}")
     else:
         return [], {}
-    if items:
+    if items and not _reddit_excluded(config):
         items = _enrich_reddit_items(items)
     return items, artifact
+
+
+def _reddit_excluded(config: dict) -> bool:
+    """Return True when EXCLUDE_SOURCES contains 'reddit'.
+
+    Respects the same suppression knob the pipeline uses for source gating,
+    so a user who set EXCLUDE_SOURCES=reddit doesn't get Reddit content
+    smuggled back in via web-search URLs.
+    """
+    raw = (config.get("EXCLUDE_SOURCES") or "").split(",")
+    return any(s.strip().lower() == "reddit" for s in raw)
 
 
 def _enrich_reddit_items(items: list[dict]) -> list[dict]:
@@ -243,6 +254,10 @@ def _enrich_reddit_items(items: list[dict]) -> list[dict]:
     Claude Code's WebFetch blocks reddit.com, so the model can't retrieve
     Reddit content from web search results. This fetches it via the public
     JSON API (reddit.com/.../.json) which bypasses that restriction.
+
+    Callers should gate this with EXCLUDE_SOURCES=reddit handling (see
+    `_reddit_excluded`) so a user who explicitly excluded Reddit doesn't
+    get Reddit content via web-search URLs.
     """
     from . import reddit_enrich
 
--- a/skills/last30days/scripts/lib/grounding.py
+++ b/skills/last30days/scripts/lib/grounding.py
@@ -260,6 +260,7 @@ def _enrich_reddit_items(items: list[dict]) -> list[dict]:
     get Reddit content via web-search URLs.
     """
     from . import reddit_enrich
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks Dave — the bypass shape is right and the fixed bug is real. Claude Code's WebFetch restriction on reddit.com is genuinely the wrong abstraction for an engine that's grounding off web-search Reddit URLs, and routing through the public JSON API closes that gap without adding any new credentials.

Pushed a small follow-up commit on top to address one concern. The original PR makes the enrichment always-on for any web-search result whose URL contains a Reddit thread, which leaks past one of the engine's user-facing knobs: a user who sets \`EXCLUDE_SOURCES=reddit\` to suppress Reddit results would still get Reddit body/comments smuggled back in through web URLs. That contradicts \`lib/pipeline.available_sources()\`, which gates top-level Reddit on the same env var.

The fix: \`web_search()\` now consults a \`_reddit_excluded(config)\` helper before calling \`_enrich_reddit_items\`. The helper mirrors the parsing pattern from \`pipeline\` (comma-separated, case-insensitive, whitespace-tolerant). When reddit is excluded, the web results themselves still flow through unchanged - they're just not augmented with Reddit body/comments. Behavior for users who DON'T set \`EXCLUDE_SOURCES\` is unchanged.

Added four tests in \`test_grounding_v3.py\` covering the gated path, case-insensitive parsing, the "other sources in EXCLUDE_SOURCES don't trigger the reddit gate" case, and the default-enrichment-still-runs case. 19/19 grounding tests green.

Rebased onto current \`main\` (was ~49 commits behind, no conflicts).

tmchow:
Greptile caught two real issues after the gate commit landed, both addressed in `e2d9d705f6`. First, `_enrich_reddit_items` was reading `parsed.get("selftext", "")` but `parse_thread_data` returns selftext under `parsed["submission"]["selftext"]` — so the headline body-enrichment was dead code and `item["snippet"]` never picked up the thread body. Fixed the key path to `(parsed.get("submission") or {}).get("selftext", "")`. Second, `RedditRateLimitError` was being swallowed by the broad `except Exception`, which meant a single 429 would let the loop keep slamming Reddit on subsequent items — now it's caught explicitly, logs a one-line stderr notice, and breaks. Added two tests in `tests/test_grounding_v3.py` covering both paths; full suite is at 21/21 green.

PR #373, opened 2026-05-10
feat(store): record per-run finding sightings

hnshah:
## Summary

Adds a per-run `finding_sightings` ledger for watchlist research runs so future watchlist deltas can compare what appeared in each run without relying on aggregate finding state.

This keeps the existing `findings` table behavior intact and records a lightweight sighting row whenever a finding appears in a completed run.

## Changes

- Adds schema migration 2 for `finding_sightings`
- Records per-run sightings from `store_findings(...)`
- Adds `get_sightings_for_run(...)` for deterministic run-level lookup
- Extends store/watchlist tests around ranked candidates and repeated sightings
- Updates the Claude cache sync path to match the current `3.2.1` package version so contract CI stays green after #372

## Testing

- [x] Ran `uv run pytest tests/test_plugin_contract.py tests/test_version_consistency.py tests/test_store.py tests/test_watchlist_commands.py tests/test_cli_v3.py -q`
- [x] Ran `PATH="/opt/homebrew/bin:$PATH" bash skills/last30days/scripts/sync.sh`

## Related Issues

Relates to watchlist delta/history work.

### DIFF
--- a/skills/last30days/scripts/store.py
+++ b/skills/last30days/scripts/store.py
@@ -163,7 +163,7 @@ MIGRATIONS: Dict[int, str] = {
     2: """
 CREATE TABLE IF NOT EXISTS finding_sightings (
     id INTEGER PRIMARY KEY,
-    finding_id INTEGER REFERENCES findings(id) ON DELETE CASCADE,
+    finding_id INTEGER NOT NULL REFERENCES findings(id) ON DELETE CASCADE,
     run_id INTEGER REFERENCES research_runs(id) ON DELETE CASCADE,
     topic_id INTEGER REFERENCES topics(id) ON DELETE CASCADE,
     source TEXT NOT NULL,
--- a/skills/last30days/scripts/store.py
+++ b/skills/last30days/scripts/store.py
@@ -159,7 +159,30 @@ _UPDATABLE_FINDING_COLUMNS = frozenset({
 })
 
 # Future migrations keyed by version number
-MIGRATIONS: Dict[int, str] = {}
+MIGRATIONS: Dict[int, str] = {
+    2: """
+CREATE TABLE IF NOT EXISTS finding_sightings (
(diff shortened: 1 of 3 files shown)


tmchow:
Thanks @hnshah — appreciated. Looked at this end-to-end and the substantive design is good: the per-run sightings ledger is the right shape for watchlist deltas (separate from the aggregate \`findings\` table, separate UNIQUE constraint on \`(run_id, finding_id)\`, indexes for the three obvious access patterns).

Greptile's two P2s read as already-addressed in your follow-up commit \`8b8d567 fix(watchlist): refresh sighting retries\`:

- **Redundant SELECT**: that commit added the \`existing_by_url\` parameter and the \`missing_urls\` shortcut, so \`_record_sightings\` only re-queries for genuinely-new URLs (the ones that just got inserted in the same transaction and won't be in the caller's lookup yet). Looks like you may not have replied to mark the P2 resolved — flagging here in case Greptile's UI was misleading.
- **INSERT OR IGNORE freeze**: same commit switched to \`INSERT … ON CONFLICT(run_id, finding_id) DO UPDATE SET …\` so re-runs with refreshed engagement/relevance scores actually update the sighting row rather than silently dropping the new data.

Rebased onto current \`main\` (was ~49 commits behind). The \`sync.sh\` touch in the original PR is naturally dropped by the rebase since that file was removed in PR #405. 43/43 affected tests (\`test_store.py\` + \`test_watchlist_commands.py\`) pass on the rebased branch.

The only thing I'd flag for the future is that the schema migration is now in place but no caller exercises \`get_sightings_for_run()\` outside the tests yet — the watchlist delta path that would consume this is still TBD. The migration is harmless to have early, but if you have a target consumer in mind (deltas dashboard, dossier mode, etc.) feel free to drop a follow-up issue referencing this ledger so the trail stays visible.

PR #380, opened 2026-05-13
fix(youtube): use ?query= not ?keyword= for ScrapeCreators search endpoint

lustrousgorilla:
## Summary

The ScrapeCreators YouTube `/search` endpoint requires `?query=`, but `_sc_youtube_search()` in `scripts/lib/youtube_yt.py` sends `?keyword=`. Every call returns HTTP 400 deterministically:

```
{"error":"missing_parameter","message":"You must provide a query"}
```

The same typo is present in both the urllib fallback branch (line 889) and the `requests` happy path (line 902).

## Changes

- `skills/last30days/scripts/lib/youtube_yt.py` — flip both `{"keyword": keyword}` dict-key occurrences to `{"query": keyword}`. The internal function variable name `keyword` stays the same; only the outbound API parameter is renamed.
- `tests/test_youtube_yt.py` — add two regression tests (one per branch) asserting the outbound parameter key is `query`, not `keyword`.

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short tests/test_youtube_yt.py tests/test_youtube_relevance.py` — 51/51 pass (49 existing + 2 new).
- [x] Manually verified the SC YouTube endpoint behavior: `?keyword=...` returns HTTP 400; `?query=...` returns the expected video list.

## Related Issues

None filed upstream yet — this PR is the fix. Surfaced during a community-signal pass where `/last30days` returned zero YouTube results deterministically across multiple research queries; a direct probe with `?query=` against the same endpoint returned rich data.

### DIFF
--- a/skills/last30days/scripts/lib/youtube_yt.py
+++ b/skills/last30days/scripts/lib/youtube_yt.py
@@ -886,7 +886,7 @@ def _sc_youtube_search(keyword: str, token: str) -> List[Dict[str, Any]]:
     if not _requests:
         try:
             from urllib.parse import urlencode
-            params = urlencode({"keyword": keyword})
+            params = urlencode({"query": keyword})
             url = f"{SCRAPECREATORS_YT_BASE}/search?{params}"
             headers = http.scrapecreators_headers(token)
             headers["User-Agent"] = http.USER_AGENT
@@ -899,7 +899,7 @@ def _sc_youtube_search(keyword: str, token: str) -> List[Dict[str, Any]]:
     try:
         resp = _requests.get(
             f"{SCRAPECREATORS_YT_BASE}/search",
-            params={"keyword": keyword},
+            params={"query": keyword},
             headers=http.scrapecreators_headers(token),
             timeout=30,
         )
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks David — diagnosis and reproduction were exactly right, and the same fix has since landed on `main` as part of edea402 ("fix(sources): unblock SC YouTube + multi-token HN searches"). The `?keyword=` → `?query=` flip is in place; SC's YouTube search endpoint now works as expected.

A few notes for context, none of them on your work:

- The `requests` fallback branch your patch touched no longer exists on main — PR #393 dropped the `requests` runtime dep and routed every provider through the `lib/http` urllib wrapper. So even if your patch had landed first, it would have needed the same followup cleanup that the consolidated fix already did in one step.
- That's also why this PR went `DIRTY/CONFLICTING` after main moved — both branches were editing the same surface in different shapes.

Closing as superseded. Appreciate the manual-probe evidence in the PR body — that kind of "I hit `?keyword=` and got HTTP 400, then `?query=` returned data" detail is what made the call obvious from the description alone.

PR #381, opened 2026-05-13
fix(hackernews): trust Algolia ranking; reject only prefix/author-only matches

lustrousgorilla:
## Summary

`_title_matches_query()` in `scripts/lib/hackernews.py` AND-requires every space-split word of the query to appear in the post title. This drops real relevant hits for any verbose multi-word query — e.g. a topic like `"agent memory mindshare last 30 days"` post-filtered out genuine Show HN posts whose titles contained "agent memory" but not "mindshare last 30 days".

The docstring's stated intent (established by the existing `test_title_matches_query_prefix_only` case) is to catch false positives where Algolia matched solely an HN prefix (`Show HN:`, `Tell HN:`, etc.) or the author username — not to AND-require every query word. Algolia already ranks hits by relevance; the wrapper's post-filter only needs to reject the prefix/author-only cases.

## Changes

- `skills/last30days/scripts/lib/hackernews.py` — replace the AND-every-word loop with: "accept if any query word appears in the stripped title body; reject only when the stripped title is empty (prefix-only) or contains none of the query words (author-only match)." Docstring rewritten to reflect the corrected intent.
- `tests/test_hackernews.py`:
  - `test_title_matches_query_partial_match` — this test asserted the buggy behavior (`"all query words must match"`). Updated to assert the corrected behavior with a docstring explaining why the previous expectation was the bug.
  - Added `test_title_matches_query_verbose_query_keeps_real_show_hn` — regression test using the canonical reproduction case (verbose query + real Show HN with a Karpathy-style LLM-wiki title).
  - Added `test_title_matches_query_no_words_in_title_rejects` — load-bearing case where no query word appears in the stripped title body, so the match must have been prefix-or-author → reject.

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short tests/test_hackernews.py` — 31/31 pass (28 existing-or-updated + 3 new). The previously-passing `test_title_matches_query_partial_match` is the test whose assertion encoded the bug — it now asserts the corrected behavior.

## Related Issues

None filed upstream. Surfaced during a community-signal pass where `/last30days` returned 0 HN results for verbose research queries; a direct Algolia probe with `tags=story&query=karpathy llm wiki` returned 5+ relevant Show HNs in the same 30-day window.

### DIFF
--- a/skills/last30days/scripts/lib/hackernews.py
+++ b/skills/last30days/scripts/lib/hackernews.py
@@ -118,27 +118,35 @@ def search_hackernews(
 
 
 def _title_matches_query(title: str, query: str, author: str = "") -> bool:
-    """Check if the query term appears in the title content, not just an HN prefix or author.
+    """Reject hits where the Algolia match was solely an HN prefix or the author username.
 
-    Returns True if the query (or any multi-word token) appears in the title
-    after stripping "Tell HN:", "Show HN:", "Ask HN:", "Launch HN:" prefixes
-    and ignoring the author name.  Returns True when query is empty (no filter).
+    Trust Algolia's relevance ranking; this filter exists only to catch false
+    positives where the query word matched solely an HN prefix ("Show HN:",
+    "Tell HN:", "Ask HN:", "Launch HN:") or the author's username field, not
+    the actual title content.
+
+    Returns True when query is empty (no filter), or when at least one query
+    word appears in the prefix-stripped title body. Returns False only when
+    the stripped title is empty or contains none of the query words — which
+    means Algolia matched on prefix or author, not on real title content.
+
+    Does NOT AND-require every query word: Algolia already ranked the hit by
+    relevance, and verbose multi-word queries (e.g. "agent memory mindshare
+    last 30 days") rarely have every word in real titles even when the hit
+    is highly relevant.
     """
     if not query:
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks David — your diagnosis was right (the AND-every-word post-filter was dropping real Algolia hits on verbose multi-word queries), and the reproduction case with the Karpathy-LLM-wiki Show HN is exactly the failure mode. Closing this as superseded rather than merging, though, because PR #388 already shipped a fix for the same root cause that's a little stricter on a few edges:

- It normalizes the query the same way `search_hackernews` does (flattening commas/hyphens via `_flatten_query_for_algolia`) so the post-filter tokens line up with what Algolia actually matched on.
- It uses word-boundary regex (`\b<word>\b`) rather than naive substring containment — that matters for short tokens, where this PR's `word in stripped` would falsely match "ai" inside "email" or "ts" inside "artists".
- It pairs the loosened gate with token-overlap relevance scoring at parse time so weak single-token matches get demoted rather than surfacing at the top.

Net effect is the same recovered-results behavior you were targeting, with the short-token false-positive risk handled. Appreciate the test case — the verbose-query Show HN regression is exactly the shape we needed to protect against, and the existing tests on main cover it. Please keep the bug reports coming.

PR #383, opened 2026-05-13
fix(bird_x): retry subprocess on non-JSON stdout (HTML anti-bot interstitial)

lustrousgorilla:
## Summary

Twitter's edge intermittently serves an HTML anti-bot interstitial in place of JSON when the bird-search subprocess hits a per-query rate limit. Before this fix, that response made `json.loads()` in `_run_bird_search()` raise `JSONDecodeError` and the function returned `{"error": ..., "items": []}` with no retry and no diagnostic — silent-empty against an orchestrator that has no way to distinguish "Twitter served HTML; retry likely succeeds" from "no tweets matched the query."

Reported during a `/last30days` community-signal pass where a Karpathy-LLM-wiki subquery returned zero X items, while a second identical run a few seconds later returned full results.

## Changes

`skills/last30days/scripts/lib/bird_x.py`:

- Extract the subprocess invocation into `_invoke_bird_subprocess()` so the retry loop can call it multiple times. Returns `(result, terminal_error)` — `terminal_error` is non-None for unrecoverable cases (`SubprocTimeout`, spawn failure) that should NOT be retried.
- In `_run_bird_search()`, wrap the `json.loads()` parse in a retry loop bounded by `MAX_JSON_DECODE_RETRIES` (=2) with `JSON_DECODE_RETRY_DELAY` (=5s) between attempts.
- On non-JSON stdout, log a diagnostic that names the shape (`looks_html`, first-80-chars stdout preview, attempt counter) so this failure mode becomes legible in logs instead of silently emptying.
- On retry exhaustion, return an error dict whose message explicitly mentions `"anti-bot interstitial"` as the likely cause, distinguishing this failure from a genuine no-results case.

Subprocess-timeout, spawn-failure, and non-zero-return-code paths are unchanged — those are terminal and don't retry.

`tests/test_bird_x.py` — 3 new regression tests:

- `test_retries_subprocess_on_html_interstitial_then_succeeds` — HTML on attempt 1, JSON on attempt 2 → returns success.
- `test_returns_error_after_all_retries_exhausted` — all attempts return HTML → diagnostic error dict mentions the anti-bot interstitial cause; items=[].
- `test_terminal_subprocess_error_is_not_retried` — subprocess timeout → terminal error, no sleep, no retry.

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short tests/test_bird_x.py` — 12/12 pass (9 baseline + 3 new).

## Notes for reviewer

The fix preserves the existing return-shape contract: on permanent failure `_run_bird_search()` still returns `{"error": ..., "items": []}` rather than raising. The diagnostic message in the error string is the only behavior change for an exhausted-retry case. If you'd prefer the function to raise on exhaustion so callers can't silently swallow it, that's a slightly larger contract change (caller-side `try/except` would also need updating) — happy to follow up.

## Related Issues

None filed upstream. Surfaced during a `/last30days` community-signal pass where the X subquery silent-emptied on the first run and returned full results on the second run, with no diagnostic explaining the divergence.

### DIFF
--- a/skills/last30days/scripts/lib/bird_x.py
+++ b/skills/last30days/scripts/lib/bird_x.py
@@ -9,6 +9,7 @@ import json
 import os
 import shutil
 import sys
+import time
 from pathlib import Path
 
 from . import http, log, subproc
@@ -16,6 +17,11 @@ from datetime import datetime
 from typing import Any, Dict, List, Optional, Tuple
 
 from .relevance import token_overlap_relevance as _compute_relevance
+
+# How many times to retry the bird-search subprocess when stdout is non-JSON
+# (typically an HTML anti-bot interstitial from Twitter's edge).
+MAX_JSON_DECODE_RETRIES = 2
+JSON_DECODE_RETRY_DELAY = 5.0  # seconds between retry attempts
 
 
 def _first_of(*values):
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks David — solid bug fix with the right test shape (HTML-then-JSON success, exhausted retries with diagnostic, terminal errors don't retry). Verified the 12 tests pass and the retry-loop refactor preserves the existing return-shape contract.

Rebased onto current `main` (the branch was ~46 commits behind, no code conflicts — just stale base). No other changes.

Minor consistency nit (non-blocking): the diagnostic logs use `source_log("X/bird", ...)` while the existing `source_log` call at line 77 uses `"Bird"`. Easy to align in a follow-up if we ever care; not gating this.

PR #379, opened 2026-05-12
chore: bump sync.sh cache path to 3.2.1

j-sperling:
## Summary

- `pyproject.toml` is at `3.2.1` but `scripts/sync.sh` still references `last30days-3/3.2.0` in the plugin cache target.
- `test_sync_cache_path_uses_skill_version` derives the expected path from the pyproject version and asserts it appears in `sync.sh`, so it currently fails on `main` and on every open PR.

## Changes

- `skills/last30days/scripts/sync.sh`: bump the cache-path version literal `3.2.0` → `3.2.1` so it matches `pyproject.toml`.

## Testing

- [x] `uv run python -m pytest tests/test_plugin_contract.py tests/test_version_consistency.py -q --tb=short` — 8/8 pass on this branch (the same slice fails on `main` without this patch).
- [ ] `bash scripts/sync.sh` — not run; the script only writes to user-home directories on a developer machine.

## Notes

A more durable fix would derive the version dynamically from `pyproject.toml` inside `sync.sh` (e.g. `version=$(grep '^version' pyproject.toml | cut -d'\"' -f2)`) so this drift can't recur, but that's a bigger change. Happy to follow up if you'd prefer.

## Related Issues

Unblocks the `plugin-contract` CI check on PR #378 (and any other open PRs).

PR #386, opened 2026-05-13
security(deps): bump urllib3 to 2.7.0 to fix CVE-2026-44431/-44432

j-sperling:
## Summary

Bumps `urllib3` from 2.6.3 to 2.7.0 in `uv.lock` to clear two HIGH-severity CVEs that trivy flags on every plugin install.

## Changes

- `uv.lock`: `urllib3 2.6.3 -> 2.7.0` (transitive via `requests`)
- `pyproject.toml`: unchanged -- urllib3 remains unpinned
- Refresh produced by: `uv lock --upgrade-package urllib3`

## CVEs

- CVE-2026-44431 -- urllib3 < 2.7.0, HIGH
- CVE-2026-44432 -- urllib3 < 2.7.0, HIGH

## Testing

- [x] Ran `uv run pytest tests/test_plugin_contract.py -q --tb=short` -- all pass
- [ ] `tests/test_version_consistency.py::test_sync_cache_path_uses_skill_version` fails on `main` as well (sync.sh COMMON_TARGETS still references `last30days-3/3.2.0`); not caused by this change, flagging for a separate fix

## Related Issues

None -- drive-by security bump.

### DIFF
--- a/skills/last30days/scripts/sync.sh
+++ b/skills/last30days/scripts/sync.sh
@@ -11,7 +11,7 @@ COMMON_TARGETS=(
   # but local development needs the cache kept in sync with the repo.
   # Do NOT add ~/.claude/skills/last30days - it creates a duplicate
   # /last30days-3 in the slash command menu alongside the plugin version.
-  "$HOME/.claude/plugins/cache/last30days-skill-private/last30days-3/3.2.0"
+  "$HOME/.claude/plugins/cache/last30days-skill-private/last30days-3/3.2.1"
   "$HOME/.claude/plugins/cache/last30days-skill-private/last30days-3-nogem/3.0.0-nogem"
   "$HOME/.agents/skills/last30days"
   "$HOME/.codex/skills/last30days"
--- a/uv.lock
+++ b/uv.lock
@@ -197,7 +197,7 @@ wheels = [
 
 [[package]]
 name = "last30days-skill"
-version = "3.1.1"
+version = "3.2.1"
 source = { virtual = "." }
 dependencies = [
     { name = "requests" },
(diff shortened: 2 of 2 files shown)


j-sperling:
Heads-up: the only failing test (`test_sync_cache_path_uses_skill_version`) fails on `main` too -- it's fixed by #379. Once that lands, this should go green automatically. No regressions from the urllib3 bump itself.

tmchow:
Thanks for the security-minded contribution and the careful CVE attribution — that kind of drive-by maintenance is genuinely valuable.

Unfortunately this PR is now obsolete because of changes that landed on main after you branched:

1. **No `urllib3` to bump.** PR #404 removed the `requests` dependency entirely (the five provider files were migrated to `urllib.request` from the Python stdlib). After that landed, `pyproject.toml` became `dependencies = []` and `uv.lock` no longer transitively pulls `urllib3`. There's nothing in the lockfile for this bump to target — `urllib3` doesn't appear in the current `uv.lock` at all.

2. **`sync.sh` was deleted.** The version-pin bump (`3.2.0` → `3.2.1`) in your branch is moot — PR #405 deleted `sync.sh` and the `test_sync_cache_path_uses_skill_version` test in the same change, precisely to stop this kind of cross-PR cascade. You correctly diagnosed the cascade in your PR description as something separate from your change; the permanent fix shipped two days later.

3. **CVEs are moot for this codebase.** Even if CVE-2026-44431 / CVE-2026-44432 affect urllib3 < 2.7.0, we don't import urllib3 anywhere — the project uses Python stdlib `urllib.request`. trivy flagging it would have been a false positive on a transitive that no longer exists.

Closing because there's no remaining diff to apply. If you spot a CVE that does affect a dependency we still have, please open a fresh PR — we'd happily review.

PR #388, opened 2026-05-14
fix(sources): unblock SC YouTube + multi-token HN searches

bradferguson:
## Summary

Two related fixes that surfaced when running `last30days` from a headless cron context with multi-keyword themed queries (e.g. `"claude, personal agents, agentic infra"`, `"AI model releases, benchmarks, Opus Sonnet Haiku"`). Both bugs caused entire sources to silently return zero items, leading to recurring "DEGRADED RUN" banners on otherwise healthy automated pipelines.

### YouTube (ScrapeCreators)
SC's `/v1/youtube/search` rejects `?keyword=` with HTTP 400:
\`\`\`
{"error":"missing_parameter","message":"You must provide a query"}
\`\`\`
The canonical SC parameter for that endpoint is `query`. Other SC endpoints we use (Reddit, TikTok, Instagram) all happened to work because they use their own per-endpoint parameter names — YouTube was the lone outlier. Two-line change in `_sc_youtube_search`.

### Hacker News (Algolia)
Multi-keyword theme queries returned zero hits across every theme. Algolia treats the `query=` parameter as a strict AND across tokens, so a 4-5 word query like `claude, personal agents, agentic infra` matches no stories in the last 30 days. Reproduced via curl:

| Query | nbHits |
|---|---|
| `claude, personal agents, agentic infra` | 0 |
| `claude agentic` | 154 |
| `claude agents` | 152 |

Two changes in `hackernews.py`:
1. **Flatten commas and hyphens to spaces** before sending — themed queries frequently contain both (e.g. `ts-bun-node`, `browser-automation`) and they tokenize awkwardly in Algolia.
2. **Add `optionalWords`** for all-but-the-first token, so Algolia ranks by how many tokens match instead of requiring all of them. The existing `token_overlap_relevance` scoring at parse time already demotes weak matches, so this doesn't hurt quality — it just stops killing them outright.

Also relaxed the `_title_matches_query` post-filter from all-words to any-word: the previous all-words rule killed every Algolia hit on multi-keyword themes because real HN titles never contain every word of a long theme query verbatim. Token-overlap relevance still demotes weak matches at the scoring stage.

## Test plan
Curl-verified against the live SC and Algolia endpoints with the themed queries that previously returned zero:

- [x] `claude, personal agents, agentic infra` — Reddit/TikTok/Instagram now return real items (was zero); HN now returns hits with optionalWords (was zero); YouTube now returns 25+ videos (was 400 Bad Request)
- [x] `ts-bun-node`, `browser-automation`, `ai-model-releases` — HN no longer empty after hyphen/comma flattening
- [x] Single-word queries unaffected (`optionalWords` only applied when ≥2 tokens)
- [x] Transcript endpoint untouched (uses `url=`, not affected by the param rename)

## Context
Surfaced while debugging a self-hosted feeds pipeline that runs `last30days` from a systemd timer on a small VPS — the assistant-side `--plan` upstream code path isn't exercised there, so the deterministic fallback became visibly important. Both fixes are local to the fallback paths; they don't change behavior when an LLM-authored `--plan` is supplied upstream.

### DIFF
--- a/skills/last30days/scripts/lib/hackernews.py
+++ b/skills/last30days/scripts/lib/hackernews.py
@@ -88,17 +88,26 @@ def search_hackernews(
 
     # Use extracted core subject instead of raw topic for cleaner Algolia matching
     core = extract_core_subject(topic)
-    _log(f"Searching for '{core}' (raw: '{topic}', since {from_date}, count={count})")
+    # Hyphens and commas tokenize awkwardly in Algolia; flatten them so themed
+    # queries like "ts-bun-node" or "claude, personal agents" become plain words.
+    core_flat = _flatten_query_for_algolia(core)
+    _log(f"Searching for '{core_flat}' (raw: '{topic}', since {from_date}, count={count})")
 
     # Use relevance-sorted search with minimum engagement filter.
     # NOTE: restrictSearchableAttributes=title omitted intentionally — it would
     # miss Ask HN/Show HN threads where the topic appears in the body.
     params = {
-        "query": core,
+        "query": core_flat,
         "tags": "story",
         "numericFilters": f"created_at_i>{from_ts},created_at_i<{to_ts},points>2",
         "hitsPerPage": str(count),
     }
+    # Algolia defaults to AND across query tokens, so a 4-5 word theme query
+    # matches no stories. Mark all-but-the-first token as optional so Algolia
+    # ranks by how many tokens match instead of requiring every one.
+    tokens = core_flat.split()
+    if len(tokens) > 1:
+        params["optionalWords"] = " ".join(tokens[1:])
 
     from urllib.parse import urlencode
     url = f"{ALGOLIA_SEARCH_URL}?{urlencode(params)}"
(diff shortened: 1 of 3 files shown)


tmchow:
Thanks — both bugs are real and still present on current `main`. The branch was 43 commits behind and predated the `scripts/lib/` → `skills/last30days/scripts/lib/` layout migration, so I cherry-picked the changes manually onto a fresh branch and force-pushed with `--force-with-lease` anchored at your original SHA. Authorship preserved.

Two tweaks vs your original diff to address Greptile's P2 comments:

- **Alias drift between callsites** — hoisted the comma/hyphen flattening into a new `_flatten_query_for_algolia()` helper used by both `search_hackernews` and `_title_matches_query`. No more two callsites with identical normalization to keep in sync by hand.
- **Any-word + substring was too loose** — Greptile was right that `word in check_text` with short tokens (`"ai"`, `"ts"`, `"go"`) lets `"ai"` falsely match `"email"` or `"artists"`. Kept the any-word relaxation (it's the right call — the all-words gate killed every multi-token theme query), but switched the match to word-boundary `\b<word>\b` so short tokens only fire on actual word matches. Token-overlap relevance still demotes weak hits at the parse stage.

Updated the existing `test_title_matches_query_partial_match` — it encoded the old all-words contract and would have asserted `False` for "AI blockchain" vs "New AI framework", which is exactly the false-negative this PR fixes. Added three new cases: no-token-in-title rejection, word-boundary vs substring (proves `"ai"` doesn't match `"email"`), and hyphen/comma flattening alignment between search param and post-filter.

32/32 hackernews tests pass; `py_compile` clean on both touched files.

PR #389, opened 2026-05-14
fix: save comparison HTML artifacts

kuyua9:
## Summary
- persist the already-rendered comparison HTML when `--emit=html` is used together with `--save-dir`
- keep the existing per-entity markdown save behavior unchanged for non-HTML output
- add regression coverage that the saved comparison HTML contains the comparison title/metadata instead of the first entity report

## Tests
- `python3 -m pytest -q --tb=short tests/test_html_render.py tests/test_save_raw_per_entity.py`
- `python3 -m py_compile skills/last30days/scripts/last30days.py`
- `git diff --check`

### DIFF
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -97,11 +97,13 @@ def save_output(
     save_dir: str,
     suffix: str = "",
     synthesis_md: str | None = None,
+    topic_override: str | None = None,
+    rendered_content: str | None = None,
 ) -> Path:
     from datetime import datetime
     path = Path(save_dir).expanduser().resolve()
     path.mkdir(parents=True, exist_ok=True)
-    slug = slugify(report.topic)
+    slug = slugify(topic_override or report.topic)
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks — real bug, minimal additive fix, good test coverage. Two small changes on top:

1. **Rebased onto current `main`.** The branch was 20 commits behind, post-#405. Conflict-free since neither `last30days.py` nor `test_html_render.py` were touched on `main` since your merge-base. Force-pushed with `--force-with-lease` anchored at your original SHA.

2. **Tweaked the Greptile DRY suggestion.** The bot was right that `entity_reports and args.emit == "html"` appeared in two places, but collapsing the full expression would have changed behavior — the else-branches actually differ (footer-display path falls back to `report.topic`; the `save_output` call falls back to `None` so `save_output` substitutes the report's own topic internally). Hoisted just the condition into a single `is_comparison_html` bool above both branches so the drift risk is gone without touching either callsite's fallback.

23/23 HTML render + save-raw-per-entity tests pass; `py_compile` clean.

PR #390, opened 2026-05-14
fix(openclaw): move SCRAPECREATORS_API_KEY to optionalEnv so the skill loads without it

### LINKED ISSUE #385: SCRAPECREATORS_API_KEY should be optionalEnv, not requires.env (OpenClaw)
## Problem

On OpenClaw, the skill is filtered from the menu because `SCRAPECREATORS_API_KEY` is declared under `requires.env`:

```yaml
metadata:
  openclaw:
    requires:
      env:
        - SCRAPECREATORS_API_KEY
```

The key is not (and should not need to be) in the OpenClaw gateway process environment. The skill's own `scripts/lib/env.py` already sources it from `~/.config/last30days/.env` at runtime. The scripts self-resolve the key — the process env gate is redundant and blocks the skill from loading.

## Impact

- Skill does not appear in OpenClaw's skill menu despite being installed and enabled
- Reddit, HN, Polymarket, GitHub, and web all work with zero API keys
- Users who have the key configured in `~/.config/last30days/.env` (the standard location) are blocked anyway

## Suggested fix

Move `SCRAPECREATORS_API_KEY` from `requires.env` to `optionalEnv`:

```yaml
metadata:
  openclaw:
    requires:
      env: []
      optionalEnv:
        - SCRAPECREATORS_API_KEY
```

This matches the README's claim that "Reddit, HN, Polymarket, and GitHub work immediately" with zero config, and lets the skill load so the scripts can source the key from their own config at runtime.

## Environment

- OpenClaw with clawhub-installed `last30days-official` v3.2.1
- macOS, `SCRAPECREATORS_API_KEY` in `~/.config/last30days/.env`

mjmirza:
## What this fixes

Closes #385.

On OpenClaw the skill is filtered out of the menu entirely because `metadata.openclaw.requires.env` hard-requires `SCRAPECREATORS_API_KEY`. That gate is wrong for two concrete reasons.

1. The skill is documented as zero-config for five sources. The README and the skill description both state Reddit, Hacker News, Polymarket, GitHub, and web search work with no API keys. A hard `requires.env` gate contradicts that and hides the skill from users who could run it today with zero setup.
2. The key is never read from the gateway process environment. `scripts/lib/env.py` sources `SCRAPECREATORS_API_KEY` from the standard config path at load time (`env.py` line 216) and treats it as optional (`env.py` line 252, default `None`). The process-env gate is redundant on every host and actively blocking on OpenClaw.

Net effect today. A user installs the skill, places the key correctly in the documented config location, and the skill still never appears in the OpenClaw menu.

## The change

`SCRAPECREATORS_API_KEY` moves from `requires.env` to `optionalEnv` in `skills/last30days/SKILL.md`. One frontmatter block, two lines. `env` becomes an empty list because `SCRAPECREATORS_API_KEY` was its only entry.

```yaml
    requires:
-     env:
-       - SCRAPECREATORS_API_KEY
+     env: []
      optionalEnv:
+       - SCRAPECREATORS_API_KEY
        - OPENAI_API_KEY
```

`primaryEnv: SCRAPECREATORS_API_KEY` is intentionally left as is. It designates the key for setup and display, not menu gating, so it is out of scope for this fix.

## Why this is safe to merge

This changes a menu-visibility gate only. No runtime code path is touched. Behavior with the key set is identical. Behavior without it goes from "skill hidden" to "skill visible and runs its five zero-config sources", which is exactly what the skill description already promises. There is no host where requiring the key in process env did useful work, because the runtime resolves it from the config file regardless.

Suggested labels: `bug`

### DIFF
--- a/skills/last30days/SKILL.md
+++ b/skills/last30days/SKILL.md
@@ -13,9 +13,9 @@ metadata:
   openclaw:
     emoji: "📰"
     requires:
-      env:
-        - SCRAPECREATORS_API_KEY
+      env: []
       optionalEnv:
+        - SCRAPECREATORS_API_KEY
         - OPENAI_API_KEY
         - XAI_API_KEY
         - OPENROUTER_API_KEY


tmchow:
Thanks for the fix and the thorough writeup, @mjmirza. Closing in favor of #392 which has the same one-line frontmatter change plus a regression test (`test_openclaw_metadata_keeps_scrapecreators_optional`) that locks the fix in so this exact bug can't silently regress on a future SKILL.md edit. The detailed reasoning in your PR body informed the merge — appreciated.

Both PRs were failing CI on the same unrelated issue (the now-removed `test_sync_cache_path_uses_skill_version` from when sync.sh had a stale version pin against SKILL.md). PR #405 deleted sync.sh entirely and that test with it, so #392 now passes after rebase.

PR #399, opened 2026-05-15
feat: honor EXCLUDE_SOURCES env var in source count + pipeline filter

spiky02plateau:
## Summary

Adds a per-run denylist via the existing-but-unused `EXCLUDE_SOURCES` config key. Two coupled changes:

1. **`pipeline.available_sources()`** filters out any source listed in `config["EXCLUDE_SOURCES"]` (comma-separated, case-insensitive, whitespace-tolerant) before returning.
2. **`hooks/scripts/check-config.sh`** "Ready — N sources active" banner subtracts excluded sources from the ScrapeCreators +3 (Reddit comments + TikTok + Instagram), so the count matches what the pipeline actually runs.

## Motivation

Use case: skip TikTok/Instagram on runs where you only want text-substantive sources, without unsetting `SCRAPECREATORS_API_KEY` (which would also kill Reddit comments).

The existing `INCLUDE_SOURCES` allowlist (used by Perplexity opt-in) doesn't cover this denylist case — `tiktok` and `instagram` get appended unconditionally in `available_sources()` whenever the SC key is set, with no opt-out short of removing the key.

This has been running as a local patch on my checkout for ~3 weeks; the pattern is small, additive, and feels like a clean upstream feature gap rather than personal taste.

## Test plan

- [x] New `TestExcludeSources` class in `tests/test_pipeline_v3.py` covering:
  - `EXCLUDE_SOURCES=tiktok,instagram` removes both, keeps reddit/hackernews
  - Empty / unset `EXCLUDE_SOURCES` is a no-op
  - Case-insensitive + whitespace-tolerant parsing
  - Works for non-SC sources too (e.g. `EXCLUDE_SOURCES=hackernews`)
- [x] All 5 new tests pass locally
- [x] Bash hook tested manually: `SCRAPECREATORS_API_KEY=x EXCLUDE_SOURCES=tiktok,instagram bash hooks/scripts/check-config.sh` returns a count 2 lower than without the env var.

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -97,7 +97,16 @@ if [[ -n "$HAS_BSKY" ]]; then
   SOURCE_COUNT=$((SOURCE_COUNT + 1))
 fi
 if [[ -n "$HAS_SCRAPECREATORS" ]]; then
-  SOURCE_COUNT=$((SOURCE_COUNT + 3))  # Reddit comments + TikTok + Instagram
+  # Start with Reddit comments + TikTok + Instagram, subtract any in EXCLUDE_SOURCES
+  SC_ADD=3
+  EXCLUDED="${ENV_EXCLUDE_SOURCES:-${EXCLUDE_SOURCES:-}}"
+  if [[ ",$EXCLUDED," == *",tiktok,"* ]]; then
+    SC_ADD=$((SC_ADD - 1))
+  fi
+  if [[ ",$EXCLUDED," == *",instagram,"* ]]; then
+    SC_ADD=$((SC_ADD - 1))
+  fi
+  SOURCE_COUNT=$((SOURCE_COUNT + SC_ADD))
 fi
 
 if [[ -n "$HAS_SCRAPECREATORS" ]]; then
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -97,13 +97,17 @@ if [[ -n "$HAS_BSKY" ]]; then
   SOURCE_COUNT=$((SOURCE_COUNT + 1))
 fi
 if [[ -n "$HAS_SCRAPECREATORS" ]]; then
-  # Start with Reddit comments + TikTok + Instagram, subtract any in EXCLUDE_SOURCES
+  # Start with Reddit comments + TikTok + Instagram, subtract any in EXCLUDE_SOURCES.
+  # Normalise EXCLUDED (lowercase + collapse whitespace around commas + strip outer
+  # whitespace) so the matching mirrors pipeline.py's .strip().lower() parsing.
   SC_ADD=3
   EXCLUDED="${ENV_EXCLUDE_SOURCES:-${EXCLUDE_SOURCES:-}}"
-  if [[ ",$EXCLUDED," == *",tiktok,"* ]]; then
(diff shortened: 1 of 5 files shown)


spiky02plateau:
Thanks @greptile-apps. Pushed [095bcae](https://github.com/mvanhorn/last30days-skill/pull/399/commits/095bcae) addressing P1, plus a reply on P2 below.

**P1 (case/whitespace divergence)** — fixed. The bash side now normalises `EXCLUDE_SOURCES` (lowercase + collapse whitespace around commas + strip outer whitespace) before substring matching, mirroring `pipeline.py`'s `.strip().lower()` parsing. Reproducer in a clean `$HOME` with config `EXCLUDE_SOURCES=TikTok,Instagram`:

  - before: `/last30days: Ready — 7 sources active.`
  - after:  `/last30days: Ready — 5 sources active.`

**P2 (non-SC excluded sources not deducted)** — intentional approximation; would prefer to leave as-is. Rationale: the bash banner is a startup hook that runs on every Claude Code session, so re-implementing `pipeline.available_sources()`'s full source-discovery logic in bash (yt-dlp probe, env detection, INCLUDE_SOURCES handling, etc.) would duplicate a lot of Python in shell. The current bash logic captures the most common practical case — excluding the ScrapeCreators-backed `tiktok`/`instagram` to save SC API quota — which is the use case that motivated the PR. For arbitrary `EXCLUDE_SOURCES=hackernews`-style excludes the banner will overcount by 1, but `pipeline.available_sources()` is authoritative for what actually runs, and the brief itself shows the true active source list at the end.

Happy to add a one-liner comment in `check-config.sh` noting that the banner is an approximation if you'd like, or to revisit if you think the duplication is worth it.

tmchow:
Thanks — fills a real gap. Two follow-ups on top of your diff:

**Wiring fix.** The filter in `pipeline.available_sources()` was correct, but `env.py::get_config()` builds its config dict from a hardcoded `keys` list that didn't include `EXCLUDE_SOURCES`. Net effect: setting it in the environment silently no-op'd through the Python pipeline. The bash hook worked (reads shell env directly), and the new unit tests passed because they construct config dicts directly and pass them to `available_sources()`, bypassing `get_config()`. Verified empirically before and after the fix. Added `('EXCLUDE_SOURCES', '')` to the keys list plus a new `TestExcludeSourcesEndToEnd` regression test that goes through `get_config()`.

**SKILL.md prose.** Per [AGENTS.md](../blob/main/AGENTS.md): *"A new engine flag with no SKILL.md integration is incomplete — the model invoking the skill won't know the flag exists."* Added a one-liner to the "Build ACTIVE_SOURCES_LIST" checklist so the model subtracts excluded sources from the active-sources display.

Small note: the PR body describes `EXCLUDE_SOURCES` as the "existing-but-unused config key" — but `git grep` shows no references in the codebase outside this PR. Doesn't affect the fix; flagging in case the framing came from a different branch or a stale local grep.

One pre-existing drift this PR surfaced, out of scope here: `SKILL.md:324-325` says tiktok/instagram require `INCLUDE_SOURCES`, but `pipeline.available_sources()` adds them unconditionally whenever `SCRAPECREATORS_API_KEY` is set. The PR's motivation rests on that real (and undocumented) auto-on behavior. Worth a separate cleanup PR to either gate them on `INCLUDE_SOURCES` to match the docs, or update the docs to match the code.

All 7 EXCLUDE_SOURCES tests pass.

tmchow:
@greptileai review

PR #407, opened 2026-05-16
feat(env): macOS Keychain credential source

DamienStevens:
## Summary

Adds the macOS Keychain as the **lowest-priority** credential source on Darwin. Items stored as generic passwords with service name `last30days-<KEY>` for the current user are picked up automatically by `get_config()`. File-based env and process env still win on collision.

The change is strictly additive: no new config knob, no opt-in flag. On non-Darwin (or when the `security` binary is missing) the loader is a no-op, so Linux and Windows behavior is unchanged.

Motivation: Keychain is the standard credential store on macOS — using it avoids leaving plaintext API keys in `~/.config/last30days/.env`. Keys round-trip through Keychain Access.app, sync via iCloud if the user has that on, and don't need to be scrubbed from dotfile backups.

## Changes

- **`skills/last30days/scripts/lib/env.py`**
  - `KEYCHAIN_SERVICE_PREFIX = "last30days-"` module constant
  - `_load_keychain(keys)` helper: platform-gated (Darwin only), `shutil.which`-gated (silent if `security` not on PATH), tolerates subprocess `TimeoutExpired` / `OSError`, strips trailing newline from stdout, returns `{}` for missing items
  - Wired into `get_config()` **before** `get_openai_auth(merged_env)` so `OPENAI_API_KEY` can come from Keychain too
  - `_CONFIG_SOURCE` reports `"keychain"` when no file source is present
  - `get_config()` docstring's priority list updated to include the new source
- **`skills/last30days/scripts/setup-keychain.sh`** (new, executable, shellcheck-clean)
  - Interactive `setup-keychain.sh` prompts for every known key (skip with empty input)
  - `setup-keychain.sh KEY [KEY..]` prompts only for the listed keys
  - `--list` shows which `last30days-*` items currently exist
  - `--delete KEY [KEY..]` removes stored items
  - `--replace` overwrites existing values (default is to skip them)
  - Uses `security add-generic-password -U` so re-runs are idempotent
- **`README.md`**
  - New `### macOS Keychain (optional)` subsection under `## Bring your own keys` with both the `setup-keychain.sh` workflow and the manual `security add-generic-password` invocation
- **`tests/test_env_keychain.py`** (new, 12 tests)
  - Non-Darwin returns `{}`
  - Missing `security` binary returns `{}`
  - Successful lookups, whitespace stripping
  - `TimeoutExpired` / `OSError` are swallowed
  - Empty stdout treated as missing
  - `get_config` integration: `_CONFIG_SOURCE == "keychain"` when keychain-only; env var overrides keychain; global `.env` outranks keychain; falls back to `env_only` when keychain empty
  - Regression guard: `OPENAI_API_KEY` from Keychain reaches `get_openai_auth`
- **`tests/test_env_cookies.py`**
  - `TestGetConfigCookieIntegration::test_get_config_injects_cookies` now also patches `lib.env._load_keychain` so the test stays hermetic on developer machines that have real Keychain entries (otherwise the loader would return real `AUTH_TOKEN` and shadow the mocked browser cookie value).

## Testing

- [x] Ran `uv run pytest -q --tb=short tests/test_plugin_contract.py tests/test_version_consistency.py tests/test_env_keychain.py tests/test_env_v3.py tests/test_env_include_sources_default.py tests/test_env_cookies.py` — **32 passed**
- [x] Full `uv run pytest -q --tb=short` — only pre-existing failures in `test_store.py`, `test_watchlist_commands.py`, `test_setup_openclaw.py`, `test_footer_nudge_suppression.py` remain (also fail on `origin/main` without this patch; unrelated to credential loading)
- [x] `shellcheck scripts/setup-keychain.sh` — clean
- [x] End-to-end on macOS with a populated keychain: `get_config()` returns `_CONFIG_SOURCE='keychain'` and the right values; precedence verified by setting `XAI_API_KEY` in env to confirm it overrides the Keychain value

### Reproducing locally (macOS)

```bash
# Store one key
security add-generic-password -a "$USER" -s last30days-XAI_API_KEY -w "xai-..."

# Or use the helper
skills/last30days/scripts/setup-keychain.sh --list
skills/last30days/scripts/setup-keychain.sh XAI_API_KEY

# Verify
python3 -c "
import sys; sys.path.insert(0, 'skills/last30days/scripts')
from lib import env
cfg = env.get_config()
print('source:', cfg['_CONFIG_SOURCE'])
print('XAI_API_KEY:', 'set' if cfg.get('XAI_API_KEY') else 'unset')
"
```

## Related Issues

None — opportunistic improvement.

## Notes for the maintainer

- I deliberately kept this **always-on for Darwin / no env-var knob**. The fallback order makes it impossible to silently shadow a file-configured key, and the loader is a no-op on non-macOS. Happy to add `LAST30DAYS_KEYCHAIN=0` opt-out if you'd prefer an explicit escape hatch.
- I picked `KEYCHAIN_SERVICE_PREFIX = "last30days-"` as the namespace, mirroring the env-var names directly. Open to renaming if you'd prefer something like `Last30Days API Key` (single shared service with `-l` label per key).

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,24 @@ These platforms don't have relationships with each other. X doesn't know what Re
 | Perplexity Sonar | OpenRouter key | Pay as you go |
 | Web search | Brave Search key | 2,000 free queries/month |
 
+### macOS Keychain (optional)
+
+On macOS you can store keys in the system Keychain instead of a `.env` file. The skill picks them up automatically as the lowest-priority source — `.env` files and process environment still win on collision.
+
+```bash
+# Interactive setup — prompts for each known key, skip with empty input
+skills/last30days/scripts/setup-keychain.sh
+
+# Or store a single key by hand
+security add-generic-password -a "$USER" -s last30days-XAI_API_KEY -w "xai-..."
+
+# Inspect / clean up
+skills/last30days/scripts/setup-keychain.sh --list
+skills/last30days/scripts/setup-keychain.sh --delete XAI_API_KEY
+```
+
+Items are stored under service name `last30days-<KEY>` for the current user. On non-Darwin platforms the loader is a no-op, so there is no behaviour change for Linux/Windows users.
+
 ## How it works
 
 1. **You type a topic.** Person, company, product, technology, "X vs Y." Anything.
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -29,6 +29,11 @@ else:
 
 CODEX_AUTH_FILE = Path(os.environ.get("CODEX_AUTH_FILE", str(Path.home() / ".codex" / "auth.json")))
 
+# macOS Keychain integration: items stored with this service prefix are picked
+# up automatically on Darwin as the lowest-priority credential source.
+# Example: `security add-generic-password -a "$USER" -s last30days-XAI_API_KEY -w "xai-..."`.
+KEYCHAIN_SERVICE_PREFIX = "last30days-"
+
 AuthSource = Literal["api_key", "codex", "none"]
 AuthStatus = Literal["ok", "missing", "expired", "missing_account_id"]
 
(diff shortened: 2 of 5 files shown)


tmchow:
Thanks for this — clean implementation and the test coverage is solid. Pushed a few fixes on top:

- Hoisted the canonical key list into `lib/env.py::KEYCHAIN_KEYS` and added `tests/test_env_keychain.py::test_keychain_keys_match_setup_script`. Greptile spotted that `setup-keychain.sh::ALL_KEYS` was missing `GOOGLE_GENAI_API_KEY` and `XIAOHONGSHU_API_BASE`; with the shared source of truth + parity test, that class of drift fails at CI time now instead of after a user reports a missing key.
- Robust `USER` fallback to `pwd.getpwuid(os.getuid()).pw_name` so the lookup still matches under sudo, in Docker without `--env USER`, or on CI runners that strip the var.

Greptile's third comment (secret briefly visible via `ps` while `security -w` runs) has no clean in-script fix — the README already shows the manual `security add-generic-password` invocation for users with strict secret hygiene, so that note stays in docs.

One open design question before merging: you flagged in the PR body that this is intentionally always-on with no env-var knob, and offered `LAST30DAYS_KEYCHAIN=0` as a possible escape hatch. Was always-on the deliberate call, or would you want to wire that opt-out in as a follow-up commit? Asking because it silently changes behavior for every macOS user on the next install — happy either way, just want it to land as a deliberate choice.

13/13 keychain tests pass locally; shellcheck clean; the four unrelated pre-existing failures noted in the PR description still fail on `main` independently of this branch.

PR #416, opened 2026-05-17
fix(ci): run full pytest suite, repair 13 rotted tests

tmchow:
## Summary

CI was running only `test_plugin_contract.py` and `test_version_consistency.py` (2 of 84 test files), masking 13 rotted tests across 4 clusters. The suite is fully offline-safe — 1402 tests complete in ~7s without network access — so the narrow scope wasn't gating integration flakiness, just stale. `validate.yml` now runs `uv run pytest` against the full suite, and the surfaced failures are fixed.

## Engine fix

`store.findings_from_report` is rerank-first by design: `ranked_candidates` is the primary persistence path, with `hackernews`/`polymarket` unconditionally supplemented from `items_by_source` (they rank poorly but matter for watchlists). When `ranked_candidates` was empty (rerank failed or was skipped), reddit, x, and every other source were silently dropped. The supplement loop now falls back to all sources from `items_by_source` only when `ranked_candidates` is empty; the normal path is unchanged.

## Test repairs

| Tests | Root cause | Fix |
|---|---|---|
| `test_store.py` (6), `test_watchlist_commands.py` (2) | Cascaded from the `findings_from_report` regression above | Resolved by the engine fix |
| `test_get_new_findings_filters_by_date` (1, latent) | Compared local-time dates against SQLite's UTC `datetime('now')` — flakes near midnight UTC. Hidden behind a count assertion that failed first | Use `datetime.now(timezone.utc)` |
| `TestPollDeviceAuth` (3) | `mock_time.time` `side_effect` lists too short — impl added a `last_reminder = time.time()` call after the tests were written | Padded the timeout test; pinned the others to `return_value=0` (their loops terminate via `urlopen`, not the clock) |
| `test_bare_run_emits_web_promo` (1) | Engine reads `~/.config/last30days/.env`, so a contributor's saved `EXA_API_KEY` made `grounding` "available" and suppressed the web promo. Also: missing X made the "x" promo preempt "web" | Set `LAST30DAYS_CONFIG_DIR=""`, run subprocess from a tmpdir (neutralizes `_find_project_env`), stub `XAI_API_KEY` |

## Verification

```
$ uv run pytest
1398 passed, 4 skipped, 2 subtests passed in 7.21s
```

The 4 skips are the opt-in `TestLiveDigg` class (`tests/test_digg.py:420`), gated on `LAST30DAYS_DIGG_LIVE=1` plus the `digg-pp-cli` binary — correct to skip in CI.

---

Claude Code

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -10,7 +10,7 @@ permissions:
   contents: read
 
 jobs:
-  plugin-contract:
+  tests:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -22,5 +22,5 @@ jobs:
       - name: Set up Python
         run: uv python install 3.12
 
-      - name: Run plugin contract tests
-        run: uv run pytest tests/test_plugin_contract.py tests/test_version_consistency.py
+      - name: Run test suite
+        run: uv run pytest
--- a/skills/last30days/scripts/store.py
+++ b/skills/last30days/scripts/store.py
@@ -14,7 +14,7 @@ import argparse
 import json
 import sqlite3
 import sys
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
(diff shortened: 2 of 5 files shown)


tmchow:
Addressed in 5d4f9ef.

Greptile's flag on `_cli_query` was right, and the same bug applies to three sibling call sites in `store.py` — they all compare local-time dates against columns written via SQLite's UTC `datetime('now')`:

- `get_daily_cost` (line 522): "today" defaults to local → wrong day's cost near UTC midnight
- `get_stats` (line 578): "7 days ago" cutoff for runs_7d / successful_7d
- `get_trending` (line 624): "N days ago" cutoff for finding activity
- `_cli_query` (line 731): the `--since` flag (Greptile's flag)

All four now use `datetime.now(timezone.utc)`. Same root cause and same fix as the test repair in the first commit.

PR #425, opened 2026-05-17
fix(hooks): SessionStart check-config.sh exits 1 when no last-run.json

### LINKED ISSUE #424: SessionStart hook (check-config.sh) exits 1 when no last-run.json → Claude Code flags hook as failed
## Summary

The `SessionStart` hook `hooks/scripts/check-config.sh` (v3.3.0) exits with status **1** on every session start when there is no recent run, causing Claude Code to display:

> `SessionStart:startup hook` — `Failed with non-blocking status code: No stderr output`

stdout is correct (the "Ready — N sources active" message prints fine); only the exit code is wrong, so it's a cosmetic-but-persistent red error on every new session.

## Root cause

The script runs under `set -euo pipefail`. In the configured branches, the **last command executed** is:

```bash
[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
```

When `~/.config/last30days/last-run.json` does not exist (fresh install, or user hasn't run `/last30days` yet), `$LAST_RUN_LINE` is empty. `[[ -n "" ]]` evaluates false, the `&&` list returns exit status 1, and because it's the final command in the script, the **whole script exits 1**.

The new-user branch is unaffected because it has an explicit `exit 0` after the same line. The two configured branches (lines ~164 and ~169, `HAS_SCRAPECREATORS` set/unset) do not.

## Reproduce

```bash
# With a configured env and no ~/.config/last30days/last-run.json
echo '{"hook_event_name":"SessionStart","source":"startup"}' \
  | bash hooks/scripts/check-config.sh; echo "exit=$?"
# -> prints the Ready message, exit=1
```

## Suggested fix

Add an explicit `exit 0` at the end of the script (the hook is informational — stdout carries the payload, exit status should always be 0). Alternatively, replace the trailing `[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"` lines with an `if`:

```bash
if [[ -n "$LAST_RUN_LINE" ]]; then echo "$LAST_RUN_LINE"; fi
```

Either makes all branches exit 0.

## Environment

- last30days-skill 3.3.0
- Claude Code CLI v2.1.143, macOS (bash 3.2 `/bin/bash`)
- ScrapeCreators configured (9 sources), no `last-run.json` present

jonasspezia:
Fixes #424.

## Problem

`hooks/scripts/check-config.sh` exits **1** on every session start when `~/.config/last30days/last-run.json` is absent. Claude Code then shows `SessionStart:startup hook — Failed with non-blocking status code`. stdout is correct; only the exit code is wrong.

Root cause: under `set -euo pipefail`, the **last command** in the configured branches is

```bash
[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
```

With no last-run file, `$LAST_RUN_LINE` is empty → `[[ -n "" ]]` is false → the `&&` list returns 1 → the script exits 1.

## Fix

Convert the three `[[ … ]] && echo` lines to explicit `if` blocks so the script always exits 0 (the hook is informational; the payload is on stdout).

## Verification

```
# configured, no last-run.json
echo '{...}' | SCRAPECREATORS_API_KEY=x bash hooks/scripts/check-config.sh; echo $?   # -> 0
# new user (no config)
bash hooks/scripts/check-config.sh; echo $?                                            # -> 0
```

`bash -n` clean. No behaviour change to printed output.

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -109,7 +109,7 @@ if [[ -z "$SETUP_COMPLETE" && -z "$CONFIG_FILE" && -z "${OPENAI_API_KEY:-}" && -
 Reddit, Hacker News, and Polymarket work out of the box.
 The setup wizard can unlock X/Twitter, YouTube, and more.
 EOF
-  [[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
+  if [[ -n "$LAST_RUN_LINE" ]]; then echo "$LAST_RUN_LINE"; fi
   exit 0
 fi
 
@@ -161,12 +161,12 @@ if [[ -n "$HAS_SCRAPECREATORS" ]]; then
   # Fully configured — compact ready message
   echo "/last30days: Ready — ${SOURCE_COUNT} sources active."
   echo "  Research any topic across social + market + web sources (last 30 days)."
-  [[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
+  if [[ -n "$LAST_RUN_LINE" ]]; then echo "$LAST_RUN_LINE"; fi
 else
   # Setup done but missing ScrapeCreators — recommend it
   echo "/last30days: Ready — ${SOURCE_COUNT} sources active."
   echo "  Research any topic across social + market + web sources (last 30 days)."
-  [[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
+  if [[ -n "$LAST_RUN_LINE" ]]; then echo "$LAST_RUN_LINE"; fi
   echo "  Tip: Add ScrapeCreators for Reddit comments + TikTok + Instagram."
   echo "  100 free credits, no credit card — scrapecreators.com"
   echo "  last30days has no affiliation with any API provider."


tmchow:
Right diagnosis. The `&&` form under `set -euo pipefail` propagates exit 1 to the script; the explicit `if` form sidesteps it. CI red is pre-existing main breakage from v3.3.0 manifest mismatch — clears once #421 lands.

tmchow:
@jonasspezia fix ci

ahhhdum:
+1, can repro on v3.3.0. Same root cause — `set -e` doesn't trip on the left of `&&`, so when `LAST_RUN_LINE` is empty the trailing `[[ ... ]] && echo` is the script's last command and its exit 1 becomes the script's exit.

Worth adding a regression test? The existing tests cover the present-last-run case but not the absent one, which is how this slipped into v3.3.0. Suggested:

```python
def test_hook_exits_zero_when_configured_without_last_run(self):
    with tempfile.TemporaryDirectory() as tmp:
        env = os.environ.copy()
        env["HOME"] = str(Path(tmp) / "home")
        env["LAST30DAYS_CONFIG_DIR"] = str(Path(tmp) / "empty-config")
        env["SCRAPECREATORS_API_KEY"] = "dummy"

        result = subprocess.run(
            ["bash", "hooks/scripts/check-config.sh"],
            cwd=REPO_ROOT,
            env=env,
            capture_output=True,
            text=True,
            check=False,
        )

        self.assertEqual(result.returncode, 0, result.stderr)
        self.assertIn("Ready", result.stdout)
        self.assertNotIn("Last run", result.stdout)
```

Happy to send as a follow-up PR if useful.

mattsilv:
Cross-referencing — same underlying bug as #430 (which applies a one-line `exit 0` at end-of-script). Either merge resolves the symptom.

Reproduced on macOS today (last30days v3.3.0, Claude Code 2.1.150). `bash -x` confirms the script ends on the trailing `[[ -n "$LAST_RUN_LINE" ]] && echo ...` short-circuit — when `LAST_RUN_LINE` is empty the test evaluates false and, under `set -e`, becomes the script's exit status. Claude Code surfaces it as:

```
SessionStart:startup hook error
Failed with non-blocking status code: No stderr output
```

every session start.

Related issues this would resolve:

- #449 — SessionStart hook check-config.sh exits 1 when configured
- #440 — SessionStart hook exits 1 when last-run.json is absent (set -e + trailing && short-circuit)
- #424 — SessionStart hook (check-config.sh) exits 1 when no last-run.json → Claude Code flags hook as failed

Sibling PR: #430.

DynamycSound:
Independently verified during triage: the SessionStart hook exits 1 when `last-run.json` is missing because the trailing `[[ -n "$LAST_RUN_LINE" ]] && echo ...` is the script's final command — it evaluates false under `set -e`, so the script's exit status is 1 and Claude Code flags the hook as failed. Confirmed present on `main`; reported in #463, #449, and #440. Heads-up that PRs #430 and #471 target this exact root cause as well — three open PRs for one bug, worth consolidating.

tmchow:
Superseded by #463 (merged) — check-config.sh no longer exits 1 when last-run.json is missing.

PR #430, opened 2026-05-19
fix(hooks/check-config): exit 0 explicitly to stop SessionStart non-blocking error

j-sperling:
## Summary

- `hooks/scripts/check-config.sh` exits 1 on a normal "configured user" code path because its last command is `[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"`. When `LAST_RUN_LINE` is empty (first session before any `/last30days` run has written `last-run.json`), the `[[ ]]` test returns 1, `&&` short-circuits, and bash uses that as the script's exit status.
- Output is correct ("Ready — N sources active"), so no user-visible problem — but SessionStart hook drivers log a non-blocking error every session start. Add a trailing `exit 0` to match the welcome branch's semantics.

## Reproduction

```
$ SCRAPECREATORS_API_KEY=fake AUTH_TOKEN=fake bash hooks/scripts/check-config.sh
/last30days: Ready — 6 sources active.
  Research any topic across social + market + web sources (last 30 days).
$ echo $?
1
```

With the patch: same stdout, `echo $?` → `0`.

## Test plan

- [x] Configured-user branch (`SCRAPECREATORS_API_KEY` set, no `last-run.json`): exits 0
- [x] Welcome branch (no env vars set): unchanged, exits 0
- [x] No stdout/stderr change in either branch
- [x] `shellcheck hooks/scripts/check-config.sh` (no new warnings introduced)

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -171,3 +171,8 @@ else
   echo "  100 free credits, no credit card — scrapecreators.com"
   echo "  last30days has no affiliation with any API provider."
 fi
+
+# The branches above end with `[[ -n "$LAST_RUN_LINE" ]] && echo ...`. When
+# LAST_RUN_LINE is empty, that test returns 1 and is the script's last command,
+# leaking exit=1 to callers (e.g. SessionStart hook drivers) despite no error.
+exit 0


mattsilv:
Reproduced on macOS today (last30days v3.3.0, Claude Code 2.1.150). `bash -x` confirms the script ends on the trailing `[[ -n "$LAST_RUN_LINE" ]] && echo ...` short-circuit — when `LAST_RUN_LINE` is empty the test evaluates false and, under `set -e`, becomes the script's exit status. The hook surfaces in Claude Code as:

```
SessionStart:startup hook error
Failed with non-blocking status code: No stderr output
```

every session start, even though the script ran fine. The explicit `exit 0` in this PR is the correct minimal fix.

Cross-referencing related reports for visibility / so a merge here can close them out:

- #449 — SessionStart hook check-config.sh exits 1 when configured
- #440 — SessionStart hook exits 1 when last-run.json is absent (set -e + trailing && short-circuit)
- #424 — SessionStart hook (check-config.sh) exits 1 when no last-run.json → Claude Code flags hook as failed
- #425 — sibling PR targeting the same symptom (different scope)

DynamycSound:
Independently verified during triage: the SessionStart hook exits 1 when `last-run.json` is missing because the script's final command is `[[ -n "$LAST_RUN_LINE" ]] && echo ...`, which evaluates false under `set -e` — so check-config.sh returns exit 1 and Claude Code flags the hook as failed. Confirmed present on `main`; reported in #463, #449, and #440. Heads-up that PRs #425 and #471 fix this same root cause — three open PRs for one bug, worth consolidating.

PR #433, opened 2026-05-19
fix(subproc): escalate to SIGKILL when child ignores SIGTERM

iliaal:
## Summary

`run_with_timeout` raises `subprocess.TimeoutExpired` past its own `except` block whenever the child ignores SIGTERM, and leaves the child running. Callers in `bird_x.py` and `youtube_yt.py` only catch `SubprocTimeout`, so they crash and leak the process.

## The bug

On timeout, `subproc.run_with_timeout` does:

1. `proc.communicate(timeout=...)` raises `subprocess.TimeoutExpired`.
2. The code sends `os.killpg(SIGTERM)`.
3. `proc.wait(timeout=5)` runs. If the child catches SIGTERM (yt-dlp installs handlers; Node has its own), the wait re-raises `subprocess.TimeoutExpired`, which escapes the `except` block.

The fallback `proc.kill()` on line 86 only runs if `killpg` itself raised, not as escalation. The documented `SubprocTimeout` contract is broken, and the child is still alive after we "cleaned up".

Validated locally with a `trap '' TERM; sleep 30` reproducer that fails on main and passes on this branch.

## The fix

Wrap `proc.wait(timeout=5)` in its own `try/except subprocess.TimeoutExpired`. On expiry, send `os.killpg(SIGKILL)` (falling back to `proc.kill()` if killpg fails), then `proc.wait()` with no timeout. SIGKILL is uncatchable, so the wait returns without blocking. Raise `SubprocTimeout` as documented.

The PR also fixes a stale `sys.path` in `tests/test_subproc.py` that pointed at `<root>/scripts/` (the pre-migration layout) instead of `<root>/skills/last30days/scripts/`. The file was passing in CI by accident, finding leftover pyc cache in an untracked local directory. The new SIGTERM-trap test exposed the gap.

## Test plan

- [x] New `test_sigterm_ignoring_child_is_sigkill_escalated` reproducer fails on main, passes here
- [x] `pytest tests/test_subproc.py`: 10/10 pass (was 9, +1)
- [x] Full suite: `pytest tests/ --ignore=tests/test_exa_search.py`: 1553 passed, 4 skipped
- [x] Mock smoke: `last30days.py "claude code" --emit=compact --mock --quick` runs end-to-end

### DIFF
--- a/skills/last30days/scripts/lib/subproc.py
+++ b/skills/last30days/scripts/lib/subproc.py
@@ -84,7 +84,15 @@ def run_with_timeout(
             os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
         except (ProcessLookupError, PermissionError, OSError):
             proc.kill()
-        proc.wait(timeout=5)
+        try:
+            proc.wait(timeout=5)
+        except subprocess.TimeoutExpired:
+            # Child ignored SIGTERM (or our killpg lost the race); escalate.
+            try:
+                os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+            except (ProcessLookupError, PermissionError, OSError):
+                proc.kill()
+            proc.wait()
         raise SubprocTimeout(f"Command {cmd[0]} timed out after {timeout}s")
 
     return SubprocResult(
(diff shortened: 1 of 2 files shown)


iliaal:

PR #434, opened 2026-05-19
fix(digg): reject non-http xUrl to close stored-XSS in HTML report

iliaal:
## Summary

A malicious Digg API response can set `xUrl` to a `javascript:` or `data:text/html;...` URL. `_format_digg_quote` renders that into a markdown link, which the HTML report's `_inline_markdown` regex copies verbatim into `<a href="...">`. Clicking the Digg attribution in the saved `.html` then executes script in the `file://` origin.

## Reachability

1. `digg.py:_parse_post` accepts any non-empty `xUrl` string with no scheme validation.
2. `digg.enrich_source_items` is wired into `pipeline.py:550`, so any query that activates Digg can pull attacker-controlled URLs.
3. `render.py:_format_digg_quote` emits `[@handle]({xurl}) via Digg: ...`.
4. `html_render.py:_inline_markdown` runs `html.escape(text, quote=True)` first, but `html.escape` does NOT touch `:` or `;`, so `data:text/html;base64,<payload>` survives. The regex then substitutes the URL into `<a href="...">` raw.

I ran a `javascript:alert(1)` reproducer against `_parse_post` on main; the malicious URL flows through unimpeded. After the fix, all five dangerous schemes I tested (`javascript:`, `data:`, `vbscript:`, `file:`, `about:`) are rejected and the post is dropped.

## The fix

In `digg.py`, validate the scheme before accepting `xUrl`. If the scheme is not `http` or `https`, treat the post as if `xUrl` were missing and drop it. One private helper, one extra check.

Reject-at-source matches the existing "post requires a usable X URL" contract (the function already returns None for empty URLs). The HTML renderer's `_inline_markdown` regex accepts any scheme as a defense-in-depth concern, and I'm tracking that as a separate follow-up PR so this diff stays scoped.

## Test plan

- [x] New `test_parse_post_drops_non_http_scheme_xurl` fails on main, passes here. Covers `javascript:`, `data:`, `vbscript:`, `file:`, `about:`.
- [x] Existing happy-path test (`https://x.com/...`) still passes; bare strings without a scheme remain rejected.
- [x] `pytest tests/test_digg.py`: 29 passed, 4 skipped (no new skips).
- [x] Full suite: `pytest tests/ --ignore=tests/test_exa_search.py`: 1553 passed, 4 skipped.
- [x] No live smoke against `digg-pp-cli`: the binary is not on this box. Mock end-to-end run via `last30days.py ... --mock --quick` is clean.

### DIFF
--- a/skills/last30days/scripts/lib/digg.py
+++ b/skills/last30days/scripts/lib/digg.py
@@ -22,6 +22,7 @@ import json
 import shutil
 from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, List, Optional
+from urllib.parse import urlparse
 
 from . import log, subproc
 from .relevance import token_overlap_relevance
@@ -296,6 +297,20 @@ def parse_digg_response(
     return items
 
 
+def _is_safe_http_url(url: str) -> bool:
+    """True iff ``url`` parses with an http or https scheme.
+
+    Used to reject upstream-supplied post URLs whose scheme would be
+    dangerous in a rendered ``<a href>`` (``javascript:``, ``data:``,
+    ``file:``, ``vbscript:``, ``about:``).
+    """
+    try:
+        scheme = urlparse(url).scheme.lower()
+    except ValueError:
+        return False
+    return scheme in ("http", "https")
+
+
 def _parse_post(raw_post: Dict[str, Any]) -> Optional[Dict[str, Any]]:
     """Reduce a digg post payload into the small dict render uses.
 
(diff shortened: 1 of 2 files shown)


iliaal:
Good catch. Accepted with a small adjustment: the suggested `_log` call would have been invisible in the actual attack scenario. `digg._log` uses `log.source_log("Digg", msg)` with default `tty_only=True`, so any rejection log is suppressed in non-interactive runs (Claude Code), which is the attack surface for a compromised Digg API. The fix uses `log.source_log(..., tty_only=False)` directly so security-class drops stay visible regardless of TTY. Added a regression test (`test_parse_post_logs_unsafe_xurl_rejection_even_in_non_tty`) that captures stderr to lock in the visibility contract. Pushed as 051a8b3.

tmchow:
@iliaal fix conflicts.

iliaal:

PR #435, opened 2026-05-19
perf(pipeline): cap inner ThreadPool to 4 under competitor fanout

iliaal:
## Summary

`--competitors N` runs main + N peer subruns in parallel under `fanout.run_competitor_fanout` (cap of 6). Each subrun opened its own inner `ThreadPoolExecutor(max_workers=16)`, peaking around 6 × 16 = 96 worker threads all hammering the same upstream APIs (X, Reddit, YouTube, etc.). 429 storms compounded across subruns, and the nested concurrency was invisible to the existing rate-limit signaling.

## What changed

1. `pipeline.py`: extracted `_inner_max_workers(stream_count, internal_subrun)`. Subrun mode caps the inner pool at 4 (down from 16). Top-level runs keep the 16 ceiling.
2. `pipeline.run`: replaced the inline `max_workers = max(4, min(16, stream_count or 1))` with the helper call. The `internal_subrun: bool` parameter already flowed through to `planner.plan_query`; this PR reuses it to bound the inner executor.
3. `last30days.py:_main_runner`: now sets `internal_subrun=comp_enabled` so the main slot in a competitor fanout gets the cap too. The competitor closure (`_competitor_runner`) already set the flag; only the main slot was missing.

## Bounds

For 6 competitors (max), the upper bound drops from 6 × 16 = 96 to 6 × 4 = 24 inner threads, plus 6 fanout outer threads. The unit test `test_subrun_total_bound_for_six_competitors` asserts this stays under 30.

For 0 competitors (solo runs), the inner pool sizing matches main.

## Test plan

- [x] New `TestInnerMaxWorkers` class. Three tests: normal ceiling at 16, subrun cap at 4, total bound under 6×subrun composition.
- [x] All three fail on main (`_inner_max_workers` is the new helper) and pass on this branch.
- [x] Full suite: `pytest tests/ --ignore=tests/test_exa_search.py`: 1555 passed, 4 skipped.
- [x] Live mock smoke with 2 competitors: `last30days.py "claude code" --emit=compact --mock --quick --competitors-list "openai,anthropic"` runs end-to-end; fanout dispatches main + 2 subruns as expected.


iliaal:
Good catch. Accepted with two extra changes beyond the literal swap: the test name `test_subrun_total_bound_for_six_competitors` had the same brittleness as the `6` literal, and the `30` bound would fail spuriously if `MAX_PARALLEL_SUBRUNS` ever went to 8 (8×4=32 trips `<=30`). Renamed to `test_subrun_caps_total_concurrency_below_uncapped` and replaced the literal bound with a derived assertion: capped concurrency must be at most half of uncapped. That way the test stays meaningful regardless of the cap value. Pushed as 33563cf.

tmchow:
@iliaal fix conflicts here too

iliaal:

PR #438, opened 2026-05-19
refactor(github): split search_github / parse_github_response / enrich_with_comments

iliaal:
## Summary

`search_github` returned a normalized `List[dict]` directly while every other adapter follows `search_X → dict envelope`, `parse_X_response → list[dict]`. The github branch in `pipeline._retrieve_stream` was the only one that called `search_*` and returned `(result, {})` without a parse step. This blocked fixture-driven testing: there was no parse function to feed a synthetic envelope to.

Split into three:

- `search_github(...) -> Dict[str, Any]`: HTTP fetch only. Returns `{"items": [raw items], "context": {core, from_date, to_date, count}}`.
- `parse_github_response(response) -> List[Dict[str, Any]]`: pure function. Normalizes, date-filters, sorts by relevance. No I/O.
- `enrich_with_comments(items, depth, token) -> List[Dict[str, Any]]`: public extraction of the old private `_enrich_top_items`. Resolves the token via env / gh CLI fallback so callers don't have to.

Pipeline now does the standard 3-call dance:

```python
response = github.search_github(...)
items = github.parse_github_response(response)
items = github.enrich_with_comments(items, depth=depth, token=token)
```

## Why the 3-call shape (and not 2)

`enrich_with_comments` makes HTTP calls and needs a token. Keeping it in `parse_github_response` would make parse impure and force every fixture-driven test to either mock HTTP or skip enrichment. Splitting it out matches the YouTube adapter's pattern (`search_youtube → parse_youtube_response → enrich_with_comments`).

## Test plan

- [x] New `TestParseGithubResponse` feeds a synthetic envelope to `parse_github_response`; covers normalization, PR detection, date filter, relevance sort, empty envelope. All five fail on main (`parse_github_response` is the new helper) and pass here.
- [x] Existing `TestSearchGithub` tests updated to chain `search → parse`: search returns envelope, parse turns it into items.
- [x] `pytest tests/test_github.py`: 25 passed (was 12). The 5 new parse tests plus envelope-shape checks on search account for the delta.
- [x] Full suite: `pytest tests/ --ignore=tests/test_exa_search.py`: 1557 passed, 4 skipped.
- [x] Mock smoke: `last30days.py "claude code" --emit=compact --mock --quick --search github` dispatches to the github branch cleanly.

### DIFF
--- a/skills/last30days/scripts/lib/github.py
+++ b/skills/last30days/scripts/lib/github.py
@@ -62,6 +62,17 @@ def _resolve_token(token: Optional[str] = None) -> Optional[str]:
     return None
 
 
+def resolve_token(token: Optional[str] = None) -> Optional[str]:
+    """Public alias for ``_resolve_token``.
+
+    The pipeline calls this once before ``search_github`` and
+    ``enrich_with_comments`` so the ``gh auth token`` subprocess fallback
+    only fires once per query when ``GITHUB_TOKEN`` is unset, instead of
+    twice (once per call site).
+    """
+    return _resolve_token(token)
+
+
 def _fetch_json(
     url: str,
     token: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


iliaal:
Both accepted, with one note on the issue 1 patch. The suggested code change kept the same `_resolve_token(token)` call inside `enrich_with_comments`, so it doesn't avoid the double resolution. The prose was the meaningful part. Implemented the underlying intent: added a public `github.resolve_token` (thin wrapper around `_resolve_token`), and pipeline now resolves once at the boundary and passes the result to both `search_github` and `enrich_with_comments`. When `GITHUB_TOKEN` is unset, the gh-CLI subprocess fallback fires once per query instead of twice. Worst-case timeout drops from 10s to 5s. For issue 2: padded the no-token envelope with the same `context` key the fetch-failure path uses, so callers reading `response["context"]["from_date"]` etc. get a consistent shape across both error paths. Added test coverage for both. Pushed as c5c0239.

PR #454, opened 2026-05-27
fix(observability): require tty_only=False in source modules' log helpers

dzivkovi:
## Problem

`log.source_log` defaults `tty_only=True` to keep interactive output uncluttered. Source modules that want their logs visible must opt out via `tty_only=False`. Three modules (Reddit, GitHub, YouTube) plus xAI / Xquik / bird_x already opted out correctly — but **ten** other source modules quietly shipped without it:

`Perplexity`, `Bluesky`, `HN`, `Instagram`, `Pinterest`, `Polymarket`, `Threads`, `TikTok`, `TruthSocial`, `Digg`.

Under any non-TTY context — Claude Code, Codex, Gemini CLI, CI, captured output, pipe-to-file — every `[Source] ...` line from these modules is silently dropped. Errors, query heartbeats, and success signals all disappear. Users see the source as either "0 items returned" or absent from the footer with no signal that anything was tried or why it failed.

For **Perplexity** specifically, this was compounded by [`render.py:_FOOTER_SOURCES`](skills/last30days/scripts/lib/render.py) omitting `perplexity` entirely — so even successful Perplexity runs never appeared in the "All agents reported back!" emoji-tree footer, making the source look dead. The synthesis LLM consuming the pass-through footer block had no Perplexity signal either.

## How users see the fix

Engine stderr under Claude Code now carries the source heartbeat lines that were previously swallowed:

```
[Perplexity] Querying perplexity/sonar-pro for 'test query' (2026-04-27 to 2026-05-27)
[Perplexity] Got synthesis (1843 chars) with 7 citations
```

The emoji-tree footer now includes Perplexity when it returns data:

```
├─ 🧠 Perplexity: 12 syntheses │ 18 citations
```

A live verification run on an aviation/travel query surfaced an OpenRouter API issue (per-key spend cap) that was previously silently failing — exactly the kind of upstream diagnostic the fix is designed to expose.

## What's in the diff

- 10 source-module `_log` helpers: add `tty_only=False`
- 2 inline `log.source_log(...)` calls in `bird_x.py` patched
- `render.py:_FOOTER_SOURCES`: add Perplexity tuple `("perplexity", "🧠", "Perplexity", "synthesis", [("citations", "citations")])`
- `log.py` docstring: contributor convention block (CONVENTION: source modules under `lib/` must call this with `tty_only=False`)
- `AGENTS.md` Rules: one-line convention bullet so the rule is discoverable
- **`tests/test_source_log_visibility.py` (new)**: auto-discovery enforcement test. Globs `lib/*.py`, AST-parses each module, finds every `log.source_log(...)` call (handles multi-line, nested parens, comments, whitespace variation), asserts `tty_only=False` appears in the kwarg list. Fails CI if any future source module forgets the opt-out. Plus 2 targeted regression tests for Perplexity/Bluesky under mocked non-TTY stderr.
- `tests/test_render_v3.py`: `test_emoji_footer_includes_perplexity_when_present` — constructs a synthetic report with a Perplexity item and asserts the rendered footer contains `🧠 Perplexity` and the citation count.

## Considered and deferred

Flipping `source_log`'s `tty_only` default to `False` would invert the footgun (forgetting the kwarg makes logs *visible* instead of *silent*) but:

1. Re-introduces the "Claude Code log clutter" problem the `tty_only=True` default was added to solve (see the docstring rationale)
2. Has real blast radius: every `_log` helper plus every test that mocks `source_log` would need updating to opt INTO suppression

The auto-discovery test sidesteps the dilemma — keep the safer-for-users default, make the safer-for-contributors enforcement automatic. Cheap and conclusive.

## Test plan

- [x] `pytest tests/test_source_log_visibility.py` — 3 passed (convention + 2 regressions)
- [x] `pytest tests/test_render_v3.py` — 34 passed (including new footer test)
- [x] Targeted suite — 83 passed
- [x] Verified end-to-end via live `/last30days` invocation: Perplexity log lines visible, emoji footer renders, citations integrate through normalize/score/dedupe/RRF/cluster/render

## Related context

The bug class is the leftover wiring from v3.0.0 (`0a9ff16`) — that commit introduced both the `tty_only=True` default and the partial migration. Three call sites were updated correctly (`github.py`, `youtube_yt.py`, `reddit.py`); these ten were missed. Subsequent fixes have already had to patch other v3.0.0 Perplexity-related gaps (`5b29b8f`, `6a5a122`, `7bda021`). This PR closes out the observability side of that pattern.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -38,6 +38,7 @@ Python 3.12+ required. Use `uv` for the env; the venv lives at `.venv/`.
 - `lib/__init__.py` must be bare package marker (comment only, NO eager imports)
 - One-time setup: `npx skills add . -g -y` copies the skill into `~/.agents/skills/<name>/` (real directory) and, for harnesses that support symlinked skill dirs, drops a per-host symlink pointing at that copy. **Working-tree edits do NOT propagate automatically** — the `~/.agents/skills/<name>/` copy is frozen at install time. To sync after edits, re-run `npx skills add . -g -y`. For live-edit on a dev machine, replace the install copy with a symlink to the working tree: `ln -sfn "$PWD/skills/last30days" ~/.agents/skills/last30days` (run from the repo root).
 - Git remote: origin = public (`mvanhorn/last30days-skill`)
+- Every `lib/*.py` call to `log.source_log(...)` must pass `tty_only=False`. The default is `True`, which silently drops every line when stderr isn't a TTY (Claude Code, Codex, CI, captured output) — turning source observability into invisible failure. Enforced by `tests/test_source_log_visibility.py`.
 
 ## Security hygiene
 - Never commit real API keys, browser cookies, auth tokens, app passwords, access tokens, or `.env` contents.
--- a/skills/last30days/scripts/lib/bird_x.py
+++ b/skills/last30days/scripts/lib/bird_x.py
@@ -252,10 +252,11 @@ def _run_bird_search(query: str, count: int, timeout: int) -> Dict[str, Any]:
                 log.source_log(
                     "X/bird",
                     f"{log_msg}; retrying in {JSON_DECODE_RETRY_DELAY:.0f}s",
+                    tty_only=False,
                 )
                 time.sleep(JSON_DECODE_RETRY_DELAY)
                 continue
-            log.source_log("X/bird", log_msg)
+            log.source_log("X/bird", log_msg, tty_only=False)
             return {
                 "error": (
                     f"Invalid JSON response after {MAX_JSON_DECODE_RETRIES} attempts "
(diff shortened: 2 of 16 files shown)


dzivkovi:
Thanks @greptile-apps[bot] — applied your suggestion in 1651312 (force-pushed to keep the PR as one clean commit). Changed `item_word` from `"synthesis"` (which would have rendered as `synthesiss` for multi-item runs through the unconditional `f"{item_word}s"` pluralizer in `_footer_line_for_source`) to `"result"`.

Also extended the test to lock in the multi-item plural behavior — `test_emoji_footer_perplexity_pluralizes_correctly` now exercises a 3-item Perplexity report and asserts:
- The footer says `3 results` (not `3 synthesiss` or `3 syntheses`)
- Citation engagement still aggregates correctly across multiple items (4+3+2 = 9 citations)

Good catch. The single-item test path I had originally only validated the `"1 synthesis"` (grammatically correct by accident) case.

Issue #463, opened 2026-05-31
SessionStart hook (check-config.sh) exits 1 when no prior run, triggers Claude Code hook-failure warning


anthonyjackman:
## Summary

The `SessionStart` hook `hooks/scripts/check-config.sh` can exit with status **1** on a normal, fully-configured session. Claude Code reports this on every session start as:

> `SessionStart:startup hook ... failed with non-blocking status code: no stderr output`

It's non-blocking (the session is unaffected) but it surfaces a confusing error to users even though setup is correct.

## Root cause

The script's final statement is:

```bash
[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
```

When `LAST_RUN_LINE` is empty — i.e. there is no `~/.config/last30days/last-run.json` yet (no prior run, or it isn't parseable) — the `[[ -n "" ]]` test evaluates **false**, the `&&` short-circuits, and the compound command returns exit status **1**. Because this is the last statement in the script, the script itself exits 1.

This affects every code path that ends on one of these `[[ -n "$LAST_RUN_LINE" ]] && echo ...` lines (the "fully configured" and the missing-ScrapeCreators branches), so a fully set-up user with no prior run sees the warning.

## Reproduce

```bash
# fresh config, no last-run.json
printf 'set -e\n[[ -n "" ]] && echo x\n' | bash; echo "exit=$?"   # -> exit=1
```

Or just start a Claude Code session with last30days configured but no prior `/last30days` run.

## Fix

Add a trailing `exit 0` to the script (the hook output is purely informational, so a clean exit is always correct). Alternatively guard the final line:

```bash
{ [[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"; } || true
```

Present in v3.3.0 and v3.3.1 (and v3.2.1 on the equivalent final line).

## Environment

- macOS (Darwin), /bin/bash 3.2
- last30days v3.3.1 via Claude Code plugin marketplace

spiderman2382:
> ## Summary
> The `SessionStart` hook `hooks/scripts/check-config.sh` can exit with status **1** on a normal, fully-configured session. Claude Code reports this on every session start as:
> 
> > `SessionStart:startup hook ... failed with non-blocking status code: no stderr output`
> 
> It's non-blocking (the session is unaffected) but it surfaces a confusing error to users even though setup is correct.
> 
> ## Root cause
> The script's final statement is:
> 
[...]

Issue #449, opened 2026-05-22
SessionStart hook check-config.sh exits 1 when configured


edimuj:
The `check-config.sh` SessionStart hook exits with code 1 on every session start when the plugin is configured, so Claude Code shows "SessionStart:startup hook error" in the header.

## What's happening

The script has `set -euo pipefail` on line 2. The unconfigured code path (line 104-113) ends with `exit 0`, but the configured paths (lines 160-173) just fall through to end-of-script. The last thing that runs is:

```bash
[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
```

When `last-run.json` doesn't exist (pretty common -- it's only created after the first `/last30days` run), `$LAST_RUN_LINE` is empty, the `[[ -n ... ]]` test returns false (exit code 1), and `set -e` makes that the script's exit code.

## Fix

Add `exit 0` at the end of the script (after line 173). Or swap the `[[ -n ... ]] && echo` pattern for an if-statement.

## Environment

- Version: 3.3.0
- Platform: Ubuntu Linux
- Node: v24.15.0

## Repro

```bash
export CLAUDE_PLUGIN_ROOT="/path/to/last30days/3.3.0"
bash "$CLAUDE_PLUGIN_ROOT/hooks/scripts/check-config.sh"
echo $?  # prints 1
```

Issue #440, opened 2026-05-19
SessionStart hook exits 1 when last-run.json is absent (set -e + trailing && short-circuit)


nipulchokshi-debug:
## Summary
`hooks/scripts/check-config.sh` exits with status 1 on every Claude Code session start when `~/.config/last30days/last-run.json` does not yet exist (i.e. the user has installed the plugin and completed setup, but hasn't run `/last30days` yet, or `LAST30DAYS_CONFIG_DIR` points somewhere without a `last-run.json`).

Claude Code surfaces this as:

```
SessionStart:startup hook error
Failed with non-blocking status code: No stderr output
```

…on every startup, even though the hook is functionally succeeding.

## Root cause
The script has `set -euo pipefail` and ends with:

```bash
if [[ -n "$HAS_SCRAPECREATORS" ]]; then
  echo "/last30days: Ready — ${SOURCE_COUNT} sources active."
  echo "  Research any topic across social + market + web sources (last 30 days)."
  [[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
else
  ...
  [[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
  echo "  Tip: ..."
  echo "  ..."
  echo "  last30days has no affiliation with any API provider."
fi
```

In the `if` branch (ScrapeCreators configured), the **last** command executed is `[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"`. When `LAST_RUN_LINE` is empty, the `[[ ... ]]` test fails, the short-circuit `&&` skips the echo, and the script's final exit status becomes 1. `set -e` doesn't kill the script (an `&&` chain in a compound statement is "tested"), but the script *returns* with `$? = 1`.

The `else` branch is fine because subsequent `echo` lines mask it.

## Repro
1. Install plugin (v3.2.4), complete setup so `SCRAPECREATORS_API_KEY` is set.
2. Don't run `/last30days` (so no `last-run.json`).
3. Start a new Claude Code session → "SessionStart:startup hook error".

```
$ bash hooks/scripts/check-config.sh; echo "exit=$?"
/last30days: Ready — 6 sources active.
  Research any topic across social + market + web sources (last 30 days).
exit=1
```

## Fix
Append `exit 0` at the end of the script, or replace the trailing test with one that doesn't poison `$?`:

```bash
[...]

PR #471, opened 2026-06-02
fix(hooks): check-config.sh exits 0 when last-run.json is missing

ahhhdum:
## What
`hooks/scripts/check-config.sh` can exit **non-zero on a successful SessionStart**. The
configured-user branch ends with:

```bash
[[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
```

When there is no last-run line (first run, `python3` unavailable, or a fresh config dir),
`[[ -n "" ]]` returns 1 and the `&& echo` short-circuits. As the script's **last command**,
that exit status (1) becomes the script's exit status. `set -e` does not trip on the left of
`&&`, so it is silent. Claude Code surfaces it as:

```
SessionStart:startup hook error / No stderr output
```

(the script writes only to stdout, so the diagnostic is empty and confusing).

## Fix
Replace the three trailing `[[ -n "$LAST_RUN_LINE" ]] && echo …` patterns with explicit
`if/then/fi` blocks — functionally equivalent on the reachable branches, and it prevents
reintroduction if the script grows new tails. Adds a regression test asserting the hook
exits 0 when SCRAPECREATORS is configured but no `last-run.json` exists.

## Repro (before fix)
```bash
SCRAPECREATORS_API_KEY=dummy HOME=/tmp/empty bash hooks/scripts/check-config.sh
# exit 1, stderr empty, stdout has the ready/welcome message
```

Surfaced on Windows (Git Bash) as a startup-hook error every session, but the bug is
platform-independent.

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -109,7 +109,9 @@ if [[ -z "$SETUP_COMPLETE" && -z "$CONFIG_FILE" && -z "${OPENAI_API_KEY:-}" && -
 Reddit, Hacker News, and Polymarket work out of the box.
 The setup wizard can unlock X/Twitter, YouTube, and more.
 EOF
-  [[ -n "$LAST_RUN_LINE" ]] && echo "$LAST_RUN_LINE"
+  if [[ -n "$LAST_RUN_LINE" ]]; then
+    echo "$LAST_RUN_LINE"
+  fi
   exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


DynamycSound:
Independently verified during triage: confirms the SessionStart hook exits 1 when `last-run.json` is absent — the trailing `[[ -n "$LAST_RUN_LINE" ]] && echo ...` is the script's last command and evaluates false under `set -e`, so Claude Code flags the hook as failed. Confirmed present on `main`; reported in #463, #449, and #440. Heads-up that PRs #425 and #430 target this same root cause — three open PRs for one bug, worth consolidating.

tmchow:
Superseded by #463 (merged) — check-config.sh no longer exits 1 when last-run.json is missing.

PR #480, opened 2026-06-05
fix(hooks): mention yt-dlp in new-user welcome when detected (fixes #394)

### LINKED ISSUE #394: check-config.sh: yt-dlp detection skipped when no config file exists (new-user early exit)
## Bug

In `hooks/scripts/check-config.sh`, the new-user detection path (lines 62–70) exits early with `exit 0` before reaching the `yt-dlp` detection block (line 77). This means users who have `yt-dlp` installed but haven't created a config file yet are never told YouTube is available.

## Repro

1. Install the skill fresh — no `~/.config/last30days/.env`, no API keys set
2. Have `yt-dlp` installed on PATH
3. Start a Claude Code session

**Expected:** Hook reports something like "Ready — 4 sources active (YouTube via yt-dlp detected)"
**Actual:** Hook prints the new-user message and exits before checking yt-dlp at all

```bash
# The early exit at line 70 — before yt-dlp check at line 77
if [[ -z "$SETUP_COMPLETE" && -z "$CONFIG_FILE" && -z "${OPENAI_API_KEY:-}" && ... ]]; then
  cat <<'WELCOME'
/last30days: Ready to use. Run /last30days to get started — setup takes 30 seconds.
...
WELCOME
  exit 0   # <-- yt-dlp never checked
fi

# Line 77 — never reached for new users with yt-dlp
if command -v yt-dlp &>/dev/null; then
  HAS_YTDLP="yes"
fi
```

## Fix

Move the `yt-dlp` check (and ideally all capability detection) before the early-exit branch, or include it in the new-user message output so users know what's already available.

## Impact

New users with yt-dlp installed run the full pipeline without YouTube transcript searches, not knowing they have the capability. Discovered when running three research sessions before realizing YouTube data was never included.

23241a6749:
## Summary

The `SessionStart` hook's new-user early-exit runs **before** the capability-detection block. Result: a user who installs the skill fresh and already has `yt-dlp` on `PATH` is never told YouTube is available. They run `/last30days` with YouTube silently dark, and only discover the gap when their briefs come back without video transcripts.

## What

| File | Change |
|---|---|
| `hooks/scripts/check-config.sh` | Hoist `HAS_YTDLP` detection above the new-user `if` block; emit a "Detected: yt-dlp is installed (YouTube transcripts ready, no setup needed)." line in the welcome message when present; drop the now-redundant duplicate detection in the setup-done branch |
| `tests/test_check_config_ytdlp_detection.py` | New pytest module, 3 cases (new user + yt-dlp / new user + no yt-dlp / existing user + yt-dlp regression) |

## Why

Closes #394 — first-run users with `yt-dlp` installed aren't told YouTube is available because the detection ran after the new-user early-exit. `yt-dlp` is the only capability check that doesn't need a config file or env var, so it's the only one that makes sense in the new-user path.

## Test

```bash
uv run pytest tests/  # 1611 passed, 4 skipped
```

Manual smoke:

```bash
# New user + yt-dlp -> welcome includes detection line
PATH="$(mktemp -d):$PATH" bash hooks/scripts/check-config.sh  # touch yt-dlp in $PATH first

# New user + no yt-dlp -> welcome unchanged
bash hooks/scripts/check-config.sh
```

## Risk

**Low.** Capability detection is read-only and only runs on `command -v` lookup. The new welcome line is purely additive output. The setup-done branch's source-counting reuses the hoisted `HAS_YTDLP` variable, so behavior for configured users is unchanged.

## Notes for reviewer

- I considered also detecting `gh` (GitHub CLI) in the new-user path, but `gh` is only useful in conjunction with the ScrapeCreators auto-auth flow (which is itself behind a setup step), so it doesn't belong in a "no setup done" path.
- I deliberately did NOT move the env-var-based detection (`SCRAPECREATORS_API_KEY`, `AUTH_TOKEN`, etc.) above the early-exit — those are only meaningful when the user has actually completed setup, and showing them in the new-user path would be confusing.


23241a6749:
Pushed a follow-up commit (19bcd1a) addressing both Greptile findings:

**Issue 1 (test):** The setup-done regression test now runs the hook twice — once with yt-dlp on PATH, once without — parses the numeric source count from `Ready — N sources active.`, and asserts the count is exactly 1 higher with yt-dlp. It also writes a minimal `last-run.json` under `tmp_path` so it isn't coupled to the pre-existing #440 exit-1 bug (which is being fixed separately in #430/#471).

**Issue 2 (shell):** The new-user welcome is now split into two variants — the yt-dlp-detected case says `Reddit, Hacker News, Polymarket, and YouTube (yt-dlp detected) work out of the box` and the wizard line drops `YouTube`; the no-yt-dlp case is unchanged. The two contradictory lines are no longer printed back-to-back.

Full suite: `1611 passed, 4 skipped` locally. CI should re-trigger on the new commit.

PR #488, opened 2026-06-06
fix(reddit): relevance floor + relevance-first ranking so viral off-topic posts don't dominate

rzachsmith:
## Summary

Reddit results were dominated by high-upvote, off-topic content (relationship drama, AITA, viral news) on most queries. Both the keyed (ScrapeCreators) and keyless (RSS) paths ranked the final list engagement-first with no relevance floor, so a 10k-upvote post with zero topic overlap outranked on-topic posts.

This adds a relevance floor and relevance-first ranking to both paths.

## Changes

- **Relevance floor** (`reddit.py` Phase 6, `reddit_keyless.py`): drop zero-overlap posts (relevance == 0) when anything relevant remains, so viral junk can't fill the section. Guard mirrors the existing date filter - keep all only when nothing scored above zero.
- **Relevance-first ranking**: rank by relevance with a bounded (<= 0.25) log-scaled engagement tiebreaker, so an off-topic viral post can never outrank an on-topic one, but discussion volume still orders similarly-relevant posts.
- **Composes with #484**: that change made comment-*enrichment slot* selection relevance-aware, but the final user-visible ordering was still engagement-first. Here, `_slot_priority` slot selection is left unchanged; a final relevance-first sort now governs display order.
- Adds `tests/test_reddit_relevance_ranking.py`.

## Testing

- [x] Ran `python -m pytest -q` - full suite green (0 failed, 4 pre-existing skips).
- Manual before/after on a real query ("best web platform for browsing new and used cars for sale"): before, the top Reddit results were AITA / BestofRedditorUpdates / viral news at relevance 0.0; after, off-topic zero-overlap posts are dropped and on-topic car-shopping threads lead.

## Related Issues

Follow-up to #484 (relevance-aware enrichment-slot selection) - this fixes the final ranking the user sees, which #484 left engagement-first.

### DIFF
--- a/skills/last30days/scripts/lib/reddit.py
+++ b/skills/last30days/scripts/lib/reddit.py
@@ -29,22 +29,9 @@ SCRAPECREATORS_BASE = "https://api.scrapecreators.com/v1/reddit"
 # Reddit's highest-upvote content (relationship drama, AITA, viral news) often
 # has near-zero topic overlap. Engagement-only ranking floats it above on-topic
 # posts, especially on bare global searches where no subreddits were resolved
-# upstream. A relevance floor + relevance-first ranking keeps an off-topic viral
-# post from ever outranking an on-topic one.
-RELEVANCE_FLOOR = 0.1
-MIN_ON_TOPIC = 5
-
-
-def _relevance_rank_key(item: Dict[str, Any]) -> float:
-    """Rank by relevance first, with a bounded engagement bonus as tiebreaker.
-
-    The log-scaled bonus is capped at 0.25 so it orders similarly-relevant posts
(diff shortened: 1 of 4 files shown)


rzachsmith:
Thanks for the review. Addressed all three in c20b1cc (no behavior change):

1. Moved `RELEVANCE_FLOOR` / `MIN_ON_TOPIC` into `relevance.py` as a single source of truth; both paths now import them so they can't diverge.
2. Moved `_relevance_rank_key` below `_total_engagement` so it no longer forward-references it.
3. Added the dropped-post log to the keyless floor, matching the keyed path's observability.

PR #495, opened 2026-06-07
fix(cli): preserve github setup flag

### LINKED ISSUE #456: argparse: `--github` flag fails with 'ambiguous option' due to prefix conflict with `--github-user`/`--github-repo`

sdhilip200:
## Summary

Fixes argparse handling for the setup GitHub auth sub-flag so setup --github is preserved for setup handling instead of being treated as an ambiguous abbreviation of --github-user or --github-repo.

## Changes

- Disable argparse long-option abbreviation in the main CLI parser.
- Add regression coverage for setup --github parsing.

## Testing

- [x] Ran python3 -m pytest -q --tb=short

## Related Issues

Fixes #456

### DIFF
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -236,7 +236,10 @@ def persist_report(report: schema.Report) -> dict[str, int]:
 
 
 def build_parser() -> argparse.ArgumentParser:
-    parser = argparse.ArgumentParser(description="Research a topic across live social, market, and grounded web sources.")
+    parser = argparse.ArgumentParser(
+        description="Research a topic across live social, market, and grounded web sources.",
+        allow_abbrev=False,
+    )
     parser.add_argument("topic", nargs="*", help="Research topic")
     parser.add_argument("--emit", default="compact", choices=["compact", "json", "context", "md", "html"])
     parser.add_argument("--search", help="Comma-separated source list")
--- a/tests/test_cli_v3.py
+++ b/tests/test_cli_v3.py
@@ -102,6 +102,12 @@ class CliV3Tests(unittest.TestCase):
         self.assertEqual(["biosecurity", "ai", "agents"], args.topic)
         self.assertEqual([], extra)
 
+    def test_build_parser_preserves_github_setup_subflag(self):
+        parser = cli.build_parser()
+        args, extra = parser.parse_known_args(["setup", "--github"])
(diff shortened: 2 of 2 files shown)


tmchow:
Superseded by #477 (merged) — allow_abbrev=False is on main. The extra regression test from this PR is welcome as a follow-up if still missing.

PR #499, opened 2026-06-07
fix(cookies): scan non-default Firefox profiles when default lacks X auth (#498)

### LINKED ISSUE #498: Firefox cookie extraction misses X login on multi-profile setups (only reads profiles.ini default)
## Summary
On macOS with multiple Firefox profiles, `scripts/lib/cookie_extract.py` reads cookies only from the single profile that `profiles.ini` marks as default. If the user is logged into x.com on a *different* profile, `auth_token`/`ct0` are never found, so X search returns 0 results even with `FROM_BROWSER=auto` (or `=firefox`) and a valid X session.

## Environment
- macOS, Firefox with multiple profiles
- last30days v3.3.2 (installed via `npx skills add`)

## Repro
1. Have 2+ Firefox profiles.
2. Be logged into x.com on a profile that is NOT the `profiles.ini` default (e.g. a `*.default-release-1` profile created after a migration).
3. Set `FROM_BROWSER=auto` + `BROWSER_CONSENT=true`, run any named-entity query.
4. Engine reports `X: 0 posts` / "Unlock X: log into x.com…" despite a valid login.

## Root cause
`_try_firefox_dir()` calls `_find_default_profile()` and queries only that one profile's `cookies.sqlite`. There's no fallback to other profiles and no profile-override setting, so a logged-in-but-non-default profile is never checked. (Verified by querying each profile's `cookies.sqlite` directly: only the non-default profile held `.x.com` `auth_token`/`ct0`.)

## Suggested fix
Try the default profile first, then fall back to scanning every profile's `cookies.sqlite` and return the first that actually contains the requested cookie names:

\`\`\`python
def _try_firefox_dir(profiles_dir, domain, cookie_names):
    import glob
    default_profile = _find_default_profile(profiles_dir)
    if default_profile is not None:
        result = _query_cookies_db(default_profile / "cookies.sqlite", domain, cookie_names)
        if result is not None:
            return result
    seen = set()
    for pattern in ("Profiles/*/cookies.sqlite", "*/cookies.sqlite"):
        for db_str in sorted(glob.glob(str(profiles_dir / pattern))):
            db = Path(db_str)
            if db in seen:
                continue
            seen.add(db)
            if default_profile is not None and db.parent == default_profile:
                continue
            result = _query_cookies_db(db, domain, cookie_names)
            if result is not None:
                return result
    return None
\`\`\`

This keeps the existing default-profile behavior and only widens the search when the default yields nothing. An optional `FIREFOX_PROFILE` env override would also help. Happy to send a PR if useful.

evanclan:
## Summary
Fixes #498. When a user is logged into x.com on a Firefox profile that is not the `profiles.ini` default, `FROM_BROWSER=auto`/`firefox` now finds `auth_token`/`ct0` by scanning the other profile databases after the default comes up empty.

## Problem
`_try_firefox_dir()` only queried the single default profile's `cookies.sqlite`. On multi-profile macOS setups where X login lives on a secondary profile (e.g. `*.default-release-1` after migration), the engine reported `X: 0 posts` despite a valid browser session.

```python
# skills/last30days/scripts/lib/cookie_extract.py (before)
profile_path = _find_default_profile(profiles_dir)
return _query_cookies_db(profile_path / "cookies.sqlite", domain, cookie_names)
```

## Solution
Keep default-profile-first behavior, then fall back to every other `cookies.sqlite` under the Firefox directory (`Profiles/*/cookies.sqlite` and `*/cookies.sqlite`). Added a regression test where the default profile holds only `.example.com` cookies but a non-default profile holds the X auth pair.

## Out of scope
- `FIREFOX_PROFILE` env override (mentioned in #498 as a nice follow-up).
- Chrome/Brave multi-profile parity — Firefox-only path per the report.

## Test plan
- [x] `python3 -m pytest tests/test_cookie_extract.py -v` (13 passed, including new `test_non_default_profile_fallback_when_default_has_no_domain_cookies`)
- [ ] CI pytest suite

## Related
- #498 — Firefox cookie extraction misses X login on multi-profile setups

Made with [Cursor](https://cursor.com)

### DIFF
--- a/skills/last30days/scripts/lib/cookie_extract.py
+++ b/skills/last30days/scripts/lib/cookie_extract.py
@@ -8,6 +8,7 @@ Only uses Python stdlib — no external dependencies.
 
 import configparser
 import functools
+import glob
 import logging
 import platform
 import shutil
@@ -190,11 +191,29 @@ def _query_cookies_db(
 
 def _try_firefox_dir(profiles_dir: Path, domain: str, cookie_names: List[str]) -> Optional[Dict[str, str]]:
     """Try to extract cookies from a Firefox profiles directory."""
-    profile_path = _find_default_profile(profiles_dir)
-    if profile_path is None:
+    default_profile = _find_default_profile(profiles_dir)
+    if default_profile is not None:
+        result = _query_cookies_db(default_profile / "cookies.sqlite", domain, cookie_names)
+        if result is not None:
+            return result
+
+    seen: set[Path] = set()
+    if default_profile is not None:
+        seen.add(default_profile / "cookies.sqlite")
+
+    for pattern in ("Profiles/*/cookies.sqlite", "*/cookies.sqlite"):
+        for db_str in sorted(glob.glob(str(profiles_dir / pattern))):
+            db = Path(db_str)
+            if db in seen:
+                continue
+            seen.add(db)
+            result = _query_cookies_db(db, domain, cookie_names)
+            if result is not None:
+                return result
+
+    if default_profile is None:
         logger.debug("No Firefox profile found in %s", profiles_dir)
-        return None
-    return _query_cookies_db(profile_path / "cookies.sqlite", domain, cookie_names)
+    return None
 
 
 def extract_firefox_cookies(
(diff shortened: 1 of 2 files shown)


tmchow:
Thanks! Firefox non-default profile fallback is landing in #521. Closing as duplicate.

PR #503, opened 2026-06-08
fix(evaluator): scrub ambient credentials in tests

abnormal749:
The evaluator env test patches a dummy config value, but developer shells can still export real API keys. Since create_eval_env intentionally prefers os.environ over config, those ambient credentials can make the assertion fail and leak the key in pytest's diff output.

Patch the test's environment to blank unrelated credential variables while keeping the explicit dummy API-key coverage. Runtime configuration precedence is unchanged.

Testing: uv run python -m pytest -q --tb=short

## Summary

Prevent the evaluator env test from reading real API keys exported in a developer shell.

## Changes

  - Blank unrelated credential environment variables inside `test_create_eval_env_and_run_last30days`
  - Keep the explicit dummy `GOOGLE_API_KEY` coverage intact
  - Leave runtime configuration precedence unchanged

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short`

## Related Issues

None


abnormal749:
Thanks for the catch. I updated the PR to share the evaluator credential key list between `create_eval_env()` and the test, so the  scrubbed env stays in sync with the production loop.

PR #504, opened 2026-06-08
Preserve YouTube items with transcripts during relevance pruninFix YouTube transcript-bearing items being pruned by relevance filteringg

### LINKED ISSUE #468: YouTube videos pruned by relevance scoring despite successful transcript extraction
### Problem

YouTube videos are dropped by `prune_low_relevance()` in `signals.py` (line 226) even when transcripts are successfully extracted. Relevance is computed from `title` and `description` only — transcript content is not considered.

For broad or long topics, YouTube video titles have low lexical overlap with the topic, resulting in `local_relevance < 0.15` and the video being pruned.

**Observed behavior:** Engine logs show `Got transcripts for 6/6 videos (0 failed)` but final output reports `YouTube: 0 videos`. Transcripts are fetched but videos never appear in results.

**Root cause chain:**
1. YouTube search finds 8 videos ✅
2. Transcripts extracted for 6/6 ✅
3. `normalize_score_dedupe()` → `signals.prune_low_relevance()` → drops items with `local_relevance < 0.15` ❌
4. Short YouTube titles (5-10 words) vs long research topics (10+ words) = low lexical overlap = low relevance

### Proposed Solution

Two approaches (in order of preference):

**A. Exempt items with transcripts from relevance pruning:**
In `signals.py:prune_low_relevance()`, skip the `minimum` threshold check when `item.source == "youtube"` and the item has a non-empty `transcript_snippet`. Rationale: if we successfully extracted a transcript, the video has relevant content regardless of title overlap.

```python
# In signals.py, passes() function (line 240):
def passes(item: schema.SourceItem) -> bool:
    rel = item.local_relevance if item.local_relevance is not None else 0.0
    # YouTube items with transcripts should not be pruned by title relevance
    if item.source == "youtube" and getattr(item, "transcript_snippet", None):
        return True
    if rel < minimum:
        return False
    # ... rest unchanged
```

**B. Include transcript in relevance computation**: 
In `normalize.py`, include `transcript_snippet` (or `transcript_highlights`) in the text scored against the topic. More correct but more invasive — requires changes to how `local_relevance` is computed.

### Alternatives Considered

1. **Lower the threshold for YouTube** — doesn't solve the root cause, just makes it less aggressive
2. **Score by transcript only, ignore title** — could work but transcript is expensive to score for every item
3. **Post-filter: keep YouTube items if they have transcripts regardless of score** — essentially approach A, different location

saurabhsingh72487-hub:
## Summary

Fixes #468.

YouTube videos with successfully extracted transcripts could still be removed by `prune_low_relevance()` because relevance filtering only considered lexical overlap and not the presence of transcript content.

This change preserves YouTube items when a non-empty `transcript_snippet` is available, preventing transcript-bearing videos from being dropped due to weak title/description matches.

## Changes

* Preserve YouTube items that contain a `transcript_snippet` during relevance pruning.
* Added regression test coverage to ensure transcript-bearing YouTube items are retained.

## Testing

```bash
python -m pytest tests/test_signals_v3.py -q
python -m pytest tests/test_signals_v3.py tests/test_youtube_yt.py -q
```

### DIFF
--- a/skills/last30days/scripts/lib/signals.py
+++ b/skills/last30days/scripts/lib/signals.py
@@ -239,14 +239,27 @@ def prune_low_relevance(
 
     def passes(item: schema.SourceItem) -> bool:
         rel = item.local_relevance if item.local_relevance is not None else 0.0
+
+        transcript = ""
+        if isinstance(item.metadata, dict):
+            transcript = item.metadata.get("transcript_snippet", "")
+
+        if item.source == "youtube" and transcript:
+            return True
+
         if rel < minimum:
             return False
-        if item.source in _SOCIAL_SOURCES and (item.engagement_score is None or item.engagement_score == 0):
+
+        if item.source in _SOCIAL_SOURCES and (
+            item.engagement_score is None or item.engagement_score == 0
+        ):
             if rel < minimum * 1.5:
                 return False
+
         sole_source = sources_present == {item.source}
         if not _passes_engagement_floor(item, sole_source):
             return False
+
         return True
 
     filtered = [item for item in items if passes(item)]
--- a/tests/test_signals_v3.py
+++ b/tests/test_signals_v3.py
@@ -195,28 +195,28 @@ class SignalsV3Tests(unittest.TestCase):
             freshness_mode="evergreen_ok",
         )
         self.assertEqual("relevant", ranked[0].item_id)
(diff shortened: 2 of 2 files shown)


tmchow:
Thanks! The YouTube transcript pruning exemption merged in #577. Closing as duplicate.

PR #514, opened 2026-06-09
Feat/bilibili adapter pr

gilgamesh1124:
## Summary

Adds Bilibili (B站) as a first-class web source, modeled after the existing `xiaohongshu_api` adapter.

- Data source: Bilibili public web search API (`/x/web-interface/wbi/search/type`), no login or API key required
- WBI signing implemented locally from SocialSisterYi/bilibili-API-collect's public spec
- `buvid3` cookie auto-bootstrapped on first use; cached at module level for the process lifetime
- The `/nav` endpoint returns `code=-101` ("not logged in") for anonymous callers but still includes the `wbi_img` keys — the adapter validates the data shape it needs rather than gating on the status code, so anonymous bootstrap works
- Engagement formula: `play*1 + like*2 + danmaku*1.5 + favorite*2.5 + coin*3`, log10-compressed and clamped to `[0.05, 1.0]`
- Kill switch via `LAST30DAYS_DISABLE_BILIBILI=1`

## Out of scope (planned follow-ups)

- Hot-comment enrichment for the Best Takes section (separate PR)
- Extracting the shared `_to_int` Chinese-numeral parser into `normalize.py` (once Weibo/Zhihu adapters land and there are 3 copies)
- SKILL.md user-facing source advertisement (deferred to maintainer's discretion)

## Test plan

- [x] `pytest tests/test_bilibili.py tests/test_bilibili_wbi.py` — 46 unit tests, fixture-based, no live network
- [x] Manual live smoke test: `search_bilibili('游戏', ...)` returned 9 real videos ranked by engagement
- [x] Existing test suite unaffected

## Notes for reviewers

- UI color uses `Colors.PURPLE` as a placeholder — `Colors` has no PINK/MAGENTA value. Happy to adjust if you have a preferred treatment for Bilibili pink (`#FB7299`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- **Bilibili (B站) as a default web source.** Searches the public WBI-signed `/x/web-interface/wbi/search/type` endpoint with an auto-bootstrapped `buvid3` cookie — no user configuration or API key required. Items are ranked by engagement (coin > favorite > like > danmaku > play), surfacing the videos B站 users actually invested in. Disable via `LAST30DAYS_DISABLE_BILIBILI=1`.
+- **Bilibili (B站) as a default web source.** Searches the public WBI-signed `/x/web-interface/wbi/search/type` endpoint with an auto-bootstrapped `buvid3` cookie — no user configuration or API key required. Items are ranked by the engagement the search API returns (favorite > like > danmaku > play). Disable via `LAST30DAYS_DISABLE_BILIBILI=1`.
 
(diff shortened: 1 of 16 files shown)


gilgamesh1124:
Re: server-side date filtering — /x/web-interface/wbi/search/type doesn't expose a reliable from/to date-range parameter, only order=pubdate sorting. Client-side filtering on the authoritative pubdate is correct; the only cost is fetching a slightly larger page. Adding an unverified server-side param risks breaking the WBI signature. Leaving as-is; happy to revisit in a follow-up if the API surface supports it.

tmchow:
Thanks! A full Bilibili source adapter is a significant new surface area (new API credentials, pipeline wiring, tests, SKILL.md). Closing so we can discuss scope in an issue before a PR this size.

PR #517, opened 2026-06-09
fix: avoid xargs quote-parsing crash in check-config.sh env trim

### LINKED ISSUE #506: `check-config.sh` SessionStart hook exits 1 when a `.env` value contains an unbalanced quote
Really solid project — the stdlib-only runtime (`pyproject.toml` `dependencies = []`) and the breadth of the test suite make this pleasant to read through. While going over the config-loading path I hit one reproducible hook failure and one date-parsing edge worth a look.

**1. `xargs` aborts the whole SessionStart hook on an unbalanced quote in `.env`**

`hooks/scripts/check-config.sh:35` trims values with `xargs`:

```bash
value=$(echo "$value" | xargs | sed 's/^["'\''"]//;s/["'\''"]$//')
```

`xargs` interprets quotes in its input, so any value containing an unbalanced `'` or `"` fails with `xargs: unterminated quote`. Combined with `set -euo pipefail` (line 2), `pipefail` propagates that non-zero status and the entire hook exits 1 — on every session start for that user.

Reproduced against the current script:

```
$ printf 'SETUP_COMPLETE=1\nXAI_API_KEY=xai-Toms'"'"'key123\n' > ~/.config/last30days/.env
$ bash hooks/scripts/check-config.sh ; echo $?
xargs: unterminated quote
1
```

Balanced quotes, spaces, and plain values all exit 0; only an odd quote breaks it. An apostrophe is realistic in pasted credentials, and the `sed` on the same line already strips surrounding quotes, so quoted values seem to be expected input.

This looks distinct from the earlier SessionStart exit-1 reports (#463/#440/#424), which concern the no-config / missing `last-run.json` path — this one is a parse-time crash on the *content* of a valid `.env`. A quote-safe trim via parameter expansion (instead of `xargs`) avoids it.

**2. `parse_date` treats any all-digit string as a Unix epoch before trying ISO**

`lib/dates.py:26-31` calls `float(date_str)` first, so `'2026'` → `1970-01-01` and `'20260115'` → `1970-08-23`. A 1970 date then falls outside the window in `normalize.filter_by_date_range` (`normalize.py:24`) and the item is silently dropped.

In practice the numeric-first order is correct for Reddit `created_utc`, and YouTube pre-converts `YYYYMMDD` (`youtube_yt.py:328`), so I couldn't trigger it from current real payloads — but `lib/grounding.py:305` and `lib/threads.py:54` route free-form date fields straight through it.

Happy to dig deeper or open a PR for the hook fix if useful — was the `parse_date` numeric-first ordering intentional for the `created_utc` case, or worth gating on an epoch magnitude check?

— Mathieu Bellot

eldar702:
## Summary
Closes #506. `hooks/scripts/check-config.sh` trimmed `.env` keys/values by piping them through `xargs`. `xargs` parses shell quoting, so a value containing an unbalanced quote (e.g. `XAI_API_KEY=xai-Toms'key123`) makes `xargs` exit 1 — and under `set -euo pipefail` that crashes the SessionStart hook on **every** session for that user.

## Root cause
`hooks/scripts/check-config.sh` (~L33-35): `key=$(echo "$key" | xargs)` / `value=$(echo "$value" | xargs | sed ...)`.

## Fix
Replace the `xargs`-based whitespace trim with bash parameter-expansion trimming (which does not parse shell quotes). The existing `sed` quote-stripping, inline-comment stripping, comment-line skipping, and empty-`.env` handling are all preserved. `+10/-2` in one file, no new deps. (Same file + bug-class as the previously-merged PR #337.)

## Testing
Added `tests/test_check_config_xargs_crash.py`: runs the hook with a `.env` value containing an unbalanced quote and asserts exit 0; also checks balanced quotes are still stripped and empty `.env` exits 0. RED before the fix (exit 1 from xargs), GREEN after.

---
*AI-assisted contribution — drafted with Claude and verified against current `main`.*

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -31,8 +31,16 @@ load_env_vars() {
       # Skip comments, empty lines
       [[ "$key" =~ ^[[:space:]]*# ]] && continue
       [[ -z "$key" ]] && continue
-      key=$(echo "$key" | xargs)
-      value=$(echo "$value" | xargs | sed 's/^["'\''"]//;s/["'\''"]$//')
+      # Trim leading/trailing whitespace using bash parameter expansion so
+      # that values containing unbalanced shell-quote characters (e.g. an
+      # apostrophe mid-token like "xai-Toms'key123") do not cause xargs to
+      # exit 1 under set -euo pipefail and crash the SessionStart hook.
+      # (xargs parses shell quoting; parameter expansion does not.)
+      key="${key#"${key%%[! ]*}"}"   # strip leading spaces
+      key="${key%"${key##*[! ]}"}"   # strip trailing spaces
+      value="${value#"${value%%[! ]*}"}"  # strip leading spaces
+      value="${value%"${value##*[! ]}"}"  # strip trailing spaces
+      value=$(echo "$value" | sed 's/^["'\''"]//;s/["'\''"]$//')
       # Strip inline comments (# preceded by whitespace) to prevent
       # command substitution in backtick-containing comments
       value="${value%%[[:space:]]#*}"
--- a/tests/test_check_config_xargs_crash.py
+++ b/tests/test_check_config_xargs_crash.py
@@ -1,0 +1,156 @@
+"""Tests for issue #506: check-config.sh crashes on .env values with unbalanced quotes.
+
+Under `set -euo pipefail`, the `xargs`-based whitespace trimming in
+`load_env_vars()` parses shell quoting. An unbalanced single quote in a
+value (e.g. ``XAI_API_KEY=xai-Toms'key123``) causes `xargs` to exit 1,
+which propagates to the whole hook and kills every Claude Code session
+start for that user.
+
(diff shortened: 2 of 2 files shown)


DynamycSound:
Independently verified during triage: confirms the `check-config.sh` crash on `.env` values containing an unbalanced quote — the `echo "$value" | xargs` trim in `load_env_vars()` aborts with `unmatched quote` and takes the SessionStart hook down with it. Reported in #506. Confirmed present on `main`; this patch resolves the root cause.

eldar702:
Thanks for the independent triage confirmation @DynamycSound — appreciated. The fix targets exactly that path (quoting/escaping the `.env` value before the `xargs` trim so an unbalanced quote can't abort `load_env_vars()` and take the SessionStart hook down). Happy to adjust if a maintainer prefers a different approach.

eldar702:
Closing — upstream already fixed the same `xargs`/unterminated-quote crash via #545 (`fix: replace xargs with sed in check-config.sh to avoid unterminated quote crash`), so this PR is superseded. Thanks!

PR #518, opened 2026-06-09
fix(security): allowlist link URL schemes in HTML report renderer

aaronjmars:
## Summary

The shareable HTML emit path (`--emit=html`) renders synthesized markdown
into an `<a href="...">` without filtering the link URL scheme
(`skills/last30days/scripts/lib/html_render.py:_inline_markdown`). The
synthesized markdown is the LLM's brief, which carries text and links
through from external sources (Reddit, X, HN, TikTok, Polymarket, etc.) —
none of which are trusted to omit `javascript:` or `data:text/html` URIs
in their post bodies.

Saved HTML reports open in the user's browser, typically under `file://`,
where a `javascript:` link executes JavaScript in the file's origin on
click, and a `data:text/html,...` URI renders attacker-controlled HTML
in the same context. The pre-link `html.escape()` pass doesn't help: the
URL portion of `[label](javascript:...)` contains no characters that
`html.escape` rewrites, so the malicious scheme survives intact into the
`href` attribute.

## Impact

A click on a malicious link in a generated report runs JavaScript in the
user's local-file origin. From there, the standard `file://` exploit
chain applies (read other files in the same directory tree on
Chromium-based browsers, exfiltrate via `fetch` / form POST, plant a
persistent payload via `localStorage` against any same-origin page the
user opens later, etc.). The route to the renderer is the regular
`/last30days <topic> --emit=html` path — no special engine flag, no
authentication boundary in front of it.

## Location

`skills/last30days/scripts/lib/html_render.py:599` (`_inline_markdown`).
The vulnerable substitution is the markdown-link rewrite at the end of
that function:

```python
escaped = re.sub(
    r"\[([^\]]+)\]\(([^)\s]+)\)",
    r'<a href="\2">\1</a>',
    escaped,
)
```

## Fix

Add a small `_is_safe_link_url` helper that accepts:

- relative URLs (no scheme, fragments, or path-only forms), and
- schemes `http`, `https`, `mailto`.

Anything else — `javascript:`, `data:`, `vbscript:`, `file:`, `ftp:`,
plus any URL containing a control character that would be stripped by the
browser's pre-scheme normalization — is rendered as plain bracketed text
(`[label](url)`). The label still surfaces, but the `<a href>` is
suppressed. Allowed links also gain `rel="noopener noreferrer"` as
defense-in-depth.

## Detected by

Aeon + semgrep (`p/security-audit`, `p/owasp-top-ten`, `p/python`,
`p/command-injection`).

- Severity: high (stored XSS in shareable artifact, exploit is a single
  click)
- CWE-79 (Improper Neutralization of Input During Web Page Generation),
  CWE-80 (Improper Neutralization of Script-Related HTML Tags in a Web
  Page).

## Verification

`python3 -m unittest -v tests.test_html_render` → **24/24 passing**
(`Ran 24 tests in 0.013s OK`). The full repo's unittest discovery runs
1200 tests with 0 failures (the only 16 errors are pre-existing pytest
import failures in test modules unrelated to this change).

New tests:

- `test_markdown_links_reject_javascript_scheme` — sweeps
  `javascript:`, `JAVASCRIPT:`, `vbscript:`, `file:///etc/passwd`.
- `test_markdown_links_reject_data_uri` — covers
  `data:text/html,<svg/onload=...>`.
- `test_markdown_links_strip_leading_whitespace_javascript` — pins the
  no-clickable-href guarantee for whitespace-prefixed URLs.
- `test_markdown_links_allow_relative_url` — guards against false
  positives on relative URLs.
- `test_markdown_links_allow_mailto` — guards against false positives
  on `mailto:`.

Updated tests:

- `test_markdown_links_convert` and `test_synthesis_md_embedded` now
  assert the new `rel="noopener noreferrer"` attribute.

---
Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron).

### DIFF
--- a/skills/last30days/scripts/lib/html_render.py
+++ b/skills/last30days/scripts/lib/html_render.py
@@ -595,6 +595,48 @@ def _promote_meta_marker(body: str) -> str:
 
 _ENGINE_FOOTER_STORE: dict[str, str] = {}
 
+# Schemes allowed in synthesized markdown links. The HTML artifact is opened in
+# a browser, so a permissive link parser exposes a stored-XSS vector: a
+# `[label](javascript:...)` or `[label](data:text/html,...)` link surviving the
+# LLM synthesis would render as a clickable script payload in the saved file.
+# Restrict link URLs to a small allowlist of schemes and accept relative URLs
+# (no scheme, no `:` before the first `/` or `?`). Anything else is rendered as
+# plain bracketed text so the label remains readable.
+_SAFE_LINK_SCHEMES = frozenset({"http", "https", "mailto"})
+
+
+def _is_safe_link_url(url: str) -> bool:
+    """Return True if a markdown link URL is safe to render as an `<a href>`.
+
+    A URL is safe if it either:
+    - has no scheme (relative URL, fragment, or path-only), or
+    - uses a scheme in the allowlist (http, https, mailto).
+
+    The scheme check is case-insensitive and tolerant of surrounding
+    whitespace per RFC 3986.
+    """
+    stripped = url.strip()
+    if not stripped:
+        return False
+    # Reject any control characters (e.g. `javascript&#x3a;alert(1)` decoded
+    # to a literal CR/LF that the browser would treat as a scheme separator).
+    if any(ord(ch) < 0x20 for ch in stripped):
+        return False
+    # No scheme — relative URL or fragment. Allow.
+    colon = stripped.find(":")
+    if colon == -1:
+        return True
+    slash = stripped.find("/")
+    question = stripped.find("?")
+    hash_ = stripped.find("#")
+    # The first `:` that comes after a `/`, `?`, or `#` is path/query/fragment,
+    # not a scheme separator. e.g. `/path:with:colons` is a relative URL.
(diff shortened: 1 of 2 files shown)


giuseppebisemi:
The `rel="noopener noreferrer"` you're adding is also what makes opening these links in a new tab safe, so the two really go together.

The HTML brief gets shared around (Slack, Notion, email), and someone clicking a source link usually doesn't want to lose the brief itself. That makes `target="_blank"` the behavior you'd want here. The catch: `target="_blank"` without `rel="noopener"` is a reverse-tabnabbing hole, and these links carry untrusted external content. So it belongs on top of your allowlist, not in a separate change fighting over the same line.

I'd rather not open a competing PR while this is in review. Once it merges I can send a one-line follow-up adding `target="_blank"` to the allowed-link branch, so your security fix stays focused. Flagging it here so we don't step on each other.

aaronjmars:
Makes sense — agreed that `target="_blank"` and `rel="noopener noreferrer"` belong together on the allowlist branch. Happy for you to send that one-line follow-up once this merges; I'll keep this PR scoped to the scheme allowlist so we don't collide. Thanks for flagging!

jenniferscott11718-byte (approved):

PR #522, opened 2026-06-09
fix: resolve Windows compatibility issues in tests and subprocess execution

dataprox:
This PR resolves multiple cross-platform compatibility issues that cause test failures when running the test suite on Windows.

### Fixes included:
1. **Keychain Import Guard**: Safely wrap UNIX-specific `import pwd` and `os.getuid` calls in a `try-except` block to prevent crashes when Darwin generic password retrieval is simulated in tests on Windows.
2. **Explicit UTF-8 Encoding**: Explicitly pass `encoding="utf-8"` to file reads and subprocess runners (using `text=True`) to avoid `UnicodeDecodeError` when Python defaults to the local ANSI code page (e.g. `gbk` on Windows).
3. **Subprocess/Command Translation**:
   - Mapped shell commands in tests (`sh`, `bash`) to Windows equivalents (`cmd` or `powershell`) on Windows.
   - Merged system environment variables (`SystemRoot`, etc.) on Windows so child processes can start correctly.
4. **Safe Timeout & Conditional Skips**:
   - Checked for availability of `os.killpg`/`os.getpgid` before calling them in `subproc.py` on timeout.
   - Automatically skip `openssl` or `bash` dependent tests if those tools are not installed.

With these changes, the entire test suite passes on Windows.

### DIFF
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -130,11 +130,16 @@ def _load_keychain(keys: list[str]) -> dict[str, str]:
         return {}
 
     import subprocess
-    import pwd
     # USER can be unset under sudo, in Docker without --env USER, or in some CI
     # runners; fall back to the OS user record so lookups still match items
     # stored by setup-keychain.sh (which uses $USER).
-    user = os.environ.get("USER") or pwd.getpwuid(os.getuid()).pw_name
+    user = os.environ.get("USER")
+    if not user:
+        try:
+            import pwd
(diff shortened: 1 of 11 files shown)


DynamycSound:
Independently verified during triage: confirms the Windows subprocess-timeout crash (`os.killpg`/`os.getpgid` raise `AttributeError`, uncaught by the `ProcessLookupError`/`PermissionError`/`OSError` handler), reported in #527, #110, #91, and #156. This overlaps with PR #520, which targets the same `subproc.py` root cause — worth consolidating to avoid a double-fix.

PR #533, opened 2026-06-10
fix(youtube): prioritize in-window videos for transcript fetching

### LINKED ISSUE #531: YouTube transcript budget is spent on out-of-window videos, leaving the final report with 0 transcripts (and a false "stale yt-dlp" nudge)
## Summary

In `search_and_transcribe()` (`scripts/lib/youtube_yt.py`), transcript candidates are selected from the merged search results sorted **by views only**. For topics where an evergreen back-catalog dominates view counts (famous lecturers, educators, musicians), every transcript slot is consumed by old videos that the `strict_recent` freshness scorer later discards. The recent videos that actually survive into the final report never get transcript fetches, so the footer reports `0/N with transcripts` even though yt-dlp is installed and every fetch succeeded.

The quality nudge then misdiagnoses this as a stale binary ("The most common remaining cause is a stale yt-dlp binary"), sending users to debug an install that's working fine.

## Environment

- last30days v3.3.2 (Claude Code plugin install)
- macOS (Darwin 25.5.0), Python 3.14, yt-dlp 2026.3.17_2 (Homebrew, current)

## Reproduction

`/last30days andrej karpathy` (intent `breaking_news`, `freshness_mode=strict_recent`, default depth). Any subject with multi-million-view old videos reproduces it.

## Observed log sequence

```
[YouTube] Found 24 videos (1 within date range, keeping all)
[YouTube] Fetching transcripts for up to 6 videos (target: 2): ['kCc8FmEb1nY', '7xTGNNLPyMI', 'zjkBMFhNj_g', 'VMj-3S1tku0', 'EWvNQjAaOHw', 'LCEmiRjPEtQ']
[YouTube] Got transcripts for 6/6 videos (0 failed)
...
✓ Research complete - ... YouTube: 2 videos ...
├─ 🔴 YouTube: 2 videos │ 161,765 views │ 0/2 with transcripts
```

All 6 candidates are Karpathy's classic lectures from 2023-2024 ("Let's build GPT from scratch", makemore series, "Intro to LLMs") — outside the 30-day window. All 6 transcripts fetched successfully and were then thrown away with their videos by freshness scoring. The two June 2026 videos that made the final report were never candidates. The nudge then reported: *"YouTube returned 2 videos but only 0 transcripts captured. The most common remaining cause is a stale yt-dlp binary"* — a false positive.

## Root cause

```python
# search_and_transcribe(), youtube_yt.py
items.sort(key=lambda x: x.get("engagement", {}).get("views", 0), reverse=True)
...
attempt_count = min(len(items), transcript_limit * 3)
candidate_ids = [item["video_id"] for item in items[:attempt_count]]
```

Candidate selection ignores the date window that the rest of the pipeline enforces, while `search_youtube()` deliberately keeps out-of-window results when in-window yield is low (`keeping all`). The two policies combine pathologically: the keep-all fallback floods the top of the view-sorted list with exactly the videos the scorer will reject.

## Suggested fix

Prefer in-window videos for the transcript budget; fall back to out-of-window ones only after. Both halves keep view ordering, so behavior is unchanged for topics where recent videos already rank on top:

```python
        attempt_count = min(len(items), transcript_limit * 3)
        # Prefer in-window videos for the transcript budget. Pure view-sorted
        # order lets an evergreen back-catalog (famous lecturers, musicians)
        # consume every slot with videos the freshness scorer later discards,
        # leaving the surviving recent videos with no transcripts.
        in_window = [i for i in items if i.get("date") and i["date"] >= from_date]
        out_of_window = [i for i in items if not (i.get("date") and i["date"] >= from_date)]
        candidate_ids = [item["video_id"] for item in (in_window + out_of_window)[:attempt_count]]
```

Tested locally against v3.3.2: `tests/test_youtube_yt.py` passes 44/44 with this change, and the failing scenario above now fetches transcripts for the in-window videos first.

A secondary improvement: the stale-binary quality nudge could check whether transcript *fetches* actually failed (it has `captions_disabled` and per-fetch results available) before blaming the binary — in this case 6/6 succeeded, which rules out a stale yt-dlp.

Happy to send this as a PR if useful.

Manvikamboz:
#531 Issue

Fixes an issue where YouTube transcript candidates were selected purely by view count, allowing high-view out-of-window videos to consume the transcript budget before recent videos were considered.

As a result, videos that survived freshness filtering often had no transcripts fetched, causing final reports to show `0/N with transcripts` even when transcript extraction had succeeded.

Fixes #531

## Changes

* Added `_prioritize_recent_for_transcripts(items, from_date)` helper.
* Prioritize in-window videos (`date >= from_date`) before out-of-window videos when selecting transcript candidates.
* Preserve view-based ordering within each group.
* Applied the prioritization logic to both:

  * `search_and_transcribe()` (yt-dlp path)
  * `search_youtube_sc()` (ScrapeCreators path)
* Left the final report display ordering unchanged (global view-descending sort remains intact).
* Clarified the stale yt-dlp quality nudge to indicate that it applies when transcript extraction was attempted but failed, rather than when transcripts were never fetched for surviving report items.

## Testing

* [x] Ran `uv run python -m pytest -q --tb=short`
* Added 6 unit tests for `_prioritize_recent_for_transcripts()`
* Added regression test covering the reported evergreen-video scenario
* Added combined ordering test verifying:

  * In-window videos are always prioritized
  * View ordering is preserved within each bucket
  * Out-of-window videos are used only after in-window candidates are exhausted

## Related Issues

Fixes #531

### DIFF
--- a/skills/last30days/scripts/lib/quality_nudge.py
+++ b/skills/last30days/scripts/lib/quality_nudge.py
@@ -285,7 +285,8 @@ def _build_nudge_text(
             )
         free_suggestions.append(
             f"YouTube returned {videos} videos but only {transcripts} transcripts "
-            f"captured{captions_note}. The most common remaining cause is a stale "
+            f"captured{captions_note}. This means transcript extraction was attempted "
+            "but failed for the candidates. The most common cause is a stale "
             "yt-dlp binary - YouTube's caption format changes frequently and old "
             "binaries silently fail every transcript. Update via your package "
             "manager: scoop update yt-dlp (Windows), brew upgrade yt-dlp (macOS), "
--- a/skills/last30days/scripts/lib/youtube_yt.py
+++ b/skills/last30days/scripts/lib/youtube_yt.py
@@ -642,6 +642,40 @@ def fetch_transcripts_parallel(
     return results
 
 
+def _prioritize_recent_for_transcripts(
+    items: List[Dict[str, Any]],
+    from_date: str,
+) -> List[Dict[str, Any]]:
+    """Return items prioritizing those within the date window for transcript fetching.
+
+    In-window videos (date >= from_date) are placed first, sorted by views
+    descending. Out-of-window videos follow, also sorted by views descending.
+    This ensures the transcript budget is spent on videos most likely to
+    survive the freshness filtering stage and appear in the final report.
+
+    Preserves existing view-based ordering for topics where all videos are
+    recent (no behavioral change when in-window == all items).
+
+    Args:
+        items: Video item dicts (each has 'date' and 'engagement.views' keys)
+        from_date: Start of the requested date window (YYYY-MM-DD)
+
+    Returns:
+        Re-ordered list: in-window items first (by views), then out-of-window (by views).
+    """
+    in_window = [
+        i for i in items
+        if i.get("date") and i["date"] >= from_date
+    ]
+    out_of_window = [
+        i for i in items
+        if not (i.get("date") and i["date"] >= from_date)
+    ]
+    in_window.sort(key=lambda x: x.get("engagement", {}).get("views", 0), reverse=True)
+    out_of_window.sort(key=lambda x: x.get("engagement", {}).get("views", 0), reverse=True)
+    return in_window + out_of_window
+
+
 def search_and_transcribe(
     topic: str,
     from_date: str,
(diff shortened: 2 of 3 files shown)


Manvikamboz:
Implemented the proposed transcript-prioritization fix and added regression coverage for the reported scenario. All checks are passing.

tmchow:
Thanks! The transcript-budget recency fix for #531 merged in #567. Closing as duplicate — the in-window-first ordering approach is superseded.

PR #543, opened 2026-06-11
fix(youtube): backfill transcripts for survivors after relevance selection (#542)

### LINKED ISSUE #542: YouTube: 0/N transcripts shipped despite 6/6 successful fetches — candidates picked by views pre-ranking, finals picked by relevance post-ranking
## Summary

The footer can report `0/N with transcripts` even when every transcript fetch succeeded. Transcript candidates are chosen **by view count before ranking**, but the final videos are chosen **by relevance after ranking** — when the two sets don't overlap, all fetched transcripts are discarded and the quality nudge then misdiagnoses it as a stale `yt-dlp` binary.

## Environment

- last30days v3.3.2 (Claude Code plugin install, macOS)
- yt-dlp 2026.6.9 (installed via brew the same day — current)
- Topic: `fastest growing claude code skills on github` (default depth, 3-subquery `--plan`)

## What happened

Engine logs from the run — every fetch succeeded:

```
[YouTube] Fetching transcripts for up to 6 videos (target: 2): ['jeA-KBv0b68', 'Ffh9OeJ7yxw', 'kZ-zzHVUrO4', 'zKBPwDpBfhs', 'uogzSxOw4LU', 'MzhIr7BfpI0']
[YouTube] Got transcripts for 6/6 videos (0 failed)
```

Final output — four *different* video IDs, zero overlap with the fetched set:

```
🔴 YouTube: 4 videos │ 156,056 views │ 0/4 with transcripts
```

Final videos: `dmrMH3FaQDo`, `w-ZESK23pRw`, `L2JKgj7WzU4`, `sBF3UumkL4Y`. None were transcript candidates.

The quality nudge then told me:

> YouTube returned 4 videos but only 0 transcripts captured. The most common remaining cause is a stale yt-dlp binary...

The binary was fine. Manual check on one of the final four pulled a 206KB caption file instantly:

```
$ yt-dlp --skip-download --write-auto-subs --sub-langs en --sub-format json3 "https://www.youtube.com/watch?v=dmrMH3FaQDo"
[info] Writing video subtitles to: test_dmrMH3FaQDo.en.json3   # 206KB, <1s
```

## Root cause

`scripts/lib/youtube_yt.py`, `search_and_transcribe()`:

```python
# Sort merged results by views descending
items.sort(key=lambda x: x.get("engagement", {}).get("views", 0), reverse=True)
...
# Step 2: Fetch transcripts for top videos by views.
transcript_limit = TRANSCRIPT_LIMITS.get(depth, TRANSCRIPT_LIMITS["default"])  # default: 2
attempt_count = min(len(items), transcript_limit * 3)
candidate_ids = [item["video_id"] for item in items[:attempt_count]]
```

Transcripts go to the top-6-by-views. Downstream, relevance reranking selects the final videos — a different ordering. On broad topics the view-count winners (high-engagement but tangential videos) and the relevance winners diverge completely, so 100% of the transcript budget is spent on videos that never ship. With `TRANSCRIPT_LIMITS["default"] = 2`, even partial overlap is unlikely.

## Suggested fix

Ideas, in order of preference:

1. **Fetch transcripts after final selection** (or add a second fetch pass for any *selected* video that's missing one). Guarantees the budget is spent on videos that actually appear in the brief, at the cost of moving the fetch later in the pipeline.
2. If pre-fetch must stay (latency/parallelism), **blend relevance into candidate ordering** instead of raw views — e.g. sort by `relevance × log(views)` so candidates approximate the final selection.
3. Either way, **teach the quality nudge to distinguish the two failure modes**: `fetched == 0` → binary/caption advice is right; `fetched > 0 but selected ∩ fetched == ∅` → it's a selection mismatch, and telling users to upgrade a healthy binary sends them down the wrong path.

Happy to provide the full run log or raw output file if useful. Love the skill — the 6/6-fetched-yet-0/4-shipped shape was only findable because the logs are so good.

rolandal:
## Summary

Fixes the `0/N with transcripts` footer on runs where every transcript fetch actually succeeded. `search_and_transcribe()` spends its transcript budget on each search's top-by-views candidates, but final selection ranks by relevance — when the two sets are disjoint, every surviving video ships transcript-less and the quality nudge misdiagnoses a healthy `yt-dlp` as stale. This adds a second-pass backfill in `_finalize_items_by_source()` that spends the same per-depth budget on the survivors, mirroring the existing digg post-selection enrichment pattern.

## Changes

- `skills/last30days/scripts/lib/youtube_yt.py`: new `backfill_transcripts(items, topic, depth)` — no-op when the per-depth budget is already met, yt-dlp is missing, or depth is `quick` (budget 0). Reuses `fetch_transcripts_parallel`, `extract_transcript_highlights`, `TRANSCRIPT_LIMITS`, and the retrieval pass's `need × 3` attempt heuristic. Marks newly discovered `captions_disabled` so quality_nudge's denominator stays correct. Never overwrites an existing snippet.
- `skills/last30days/scripts/lib/pipeline.py`: youtube branch in `_finalize_items_by_source()` (sibling of the digg branch, same budget-at-the-survivors principle); `depth`/`mock` threaded through the single call site, `mock=True` skips the network pass.
- `tests/test_youtube_backfill.py`: 10 tests — the #542 disjoint-sets shape, budget-met no-op, partial-budget gap fetch, quick-depth no-op, missing-binary no-op, captions-disabled skip and discovery, snippet non-clobbering, and finalize wiring incl. mock-mode skip.

Because `RetrievalBundle.add_items()` appends the same `SourceItem` objects to both `items_by_source` and `items_by_source_and_query`, and finalize runs before `weighted_rrf`, the backfilled metadata reaches the footer, the raw dump, and the ranked evidence — not just the counter.

## Testing

- [x] Ran `uv run python -m pytest -q --tb=short` — 1627 passed, 4 skipped (live-gated, same as main)
- [x] Live before/after on the original failing topic (`fastest growing claude code skills on github`, default depth, same machine):

Before (v3.3.2):

```
[YouTube] Got transcripts for 6/6 videos (0 failed)
├─ 🔴 YouTube: 4 videos │ 156,056 views │ 0/4 with transcripts
```

After (this branch) — the mismatch reproduced organically and the backfill caught it:

```
[YouTube] Got transcripts for 6/6 videos (0 failed)
[YouTube] Backfilling transcripts for 4 finalized videos (target: 2)
[YouTube] Got transcripts for 4/4 videos (0 failed)
├─ 🔴 YouTube: 4 videos │ 124,582 views │ 4/4 with transcripts
```

Disclosed tradeoff: when the backfill fires it adds one parallel fetch round between retrieval and rerank (~20s on the run above: 185s vs 162s). It only fires when survivors lack transcripts, capped at `need × 3` attempts; runs whose retrieval candidates survive ranking are untouched.

## Related Issues

Fixes #542

### DIFF
--- a/skills/last30days/scripts/lib/pipeline.py
+++ b/skills/last30days/scripts/lib/pipeline.py
@@ -455,7 +455,9 @@ def run(
         if bundle.items_by_source.get(source):
             del bundle.errors_by_source[source]
 
-    items_by_source = _finalize_items_by_source(bundle.items_by_source, topic=topic, config=config)
+    items_by_source = _finalize_items_by_source(
+        bundle.items_by_source, topic=topic, config=config, depth=depth, mock=mock,
+    )
     candidates = weighted_rrf(bundle.items_by_source_and_query, plan, pool_limit=settings["pool_limit"])
     ranked_candidates = rerank.rerank_candidates(
         topic=topic,
@@ -525,11 +527,21 @@ def _finalize_items_by_source(
     items_by_source_raw: dict[str, list[schema.SourceItem]],
     topic: str = "",
     config: dict | None = None,
+    depth: str = "default",
+    mock: bool = False,
 ) -> dict[str, list[schema.SourceItem]]:
     finalized = {}
     for source, items in items_by_source_raw.items():
         items = sorted(items, key=lambda item: item.local_rank_score or 0.0, reverse=True)
         items = dedupe.dedupe_items(items)
+        if source == "youtube" and items and not mock:
+            # Same budget-at-the-survivors principle as the digg branch
+            # below: retrieval-time transcripts go to each search's
+            # top-by-views candidates, while final selection ranks by
+            # relevance. Backfill survivors that arrived without one so the
+            # transcript budget lands on videos the brief actually shows
+            # (#542).
+            youtube_yt.backfill_transcripts(items, topic=topic, depth=depth)
         # Post-merge topic-relevance filter for Polymarket: comparison queries
         # fan out into per-entity subqueries ("Hermes", "OpenClaw") whose topic
         # is too narrow for Gamma API to filter meaningfully. Re-validating the
--- a/skills/last30days/scripts/lib/youtube_yt.py
+++ b/skills/last30days/scripts/lib/youtube_yt.py
@@ -642,6 +642,65 @@ def fetch_transcripts_parallel(
     return results
 
 
+def backfill_transcripts(items: List[Any], topic: str = "", depth: str = "default") -> None:
+    """Second-pass transcript fetch for finalized items that lack one.
+
+    search_and_transcribe() spends its transcript budget on each search's
+    top-by-views candidates, but the pipeline's final selection ranks by
+    relevance - the two orderings can be fully disjoint, shipping a brief
+    where every surviving video is transcript-less while the fetched
+    transcripts ride videos that never appear (#542). Designed to run from
+    _finalize_items_by_source() so the per-depth budget is spent on the
+    survivors, mirroring the digg post-selection enrichment.
+
+    Mutates SourceItem.metadata in place; safe no-op when the budget is
+    already met, yt-dlp is missing, or every survivor has a transcript.
+    """
+    limit = TRANSCRIPT_LIMITS.get(depth, TRANSCRIPT_LIMITS["default"])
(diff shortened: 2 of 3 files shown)


rolandal:
Addressed the Greptile note in ea52c2d — the finalize wiring test now asserts `topic` forwarding alongside `depth`.

PR #545, opened 2026-06-11
fix: replace xargs with sed in check-config.sh to avoid unterminated quote crash

### LINKED ISSUE #506: `check-config.sh` SessionStart hook exits 1 when a `.env` value contains an unbalanced quote

23241a6749:
xargs interprets quotes in its input, so a `.env` value containing an unbalanced quote (e.g. `XAI_API_KEY=xai-Tom's-key`) causes `xargs: unterminated quote`. With `set -euo pipefail`, `pipefail` propagates that non-zero status and the entire SessionStart hook exits 1.

Replaced `| xargs |` with `sed 's/^[[:space:]]\*//;s/[[:space:]]\*$//'` for whitespace trimming, which doesn't interpret quotes. The existing quote-stripping step is unchanged.

- hooks/scripts/check-config.sh: two xargs calls replaced with sed
- tests/test_last_run_state.py: new test for .env with apostrophe in value
- CHANGELOG.md + CONTRIBUTORS.md updated

Fixes #506


tmchow:
@23241a6749 please resolve merge conflicts

PR #546, opened 2026-06-11
fix(setup): stop first-run wizard from triggering Chrome Keychain prompt

An-idd:
## Summary

On macOS, the first-run setup wizard pops a **"Chrome Safe Storage" Keychain password prompt** — the same dialog malware uses to steal browser cookies. The wizard probes browser cookies in `"auto"` mode, which includes Chrome; reading Chrome/Brave cookies on macOS requires the Chrome Safe Storage Keychain key, triggering the prompt.

This directly contradicts the steady-state path (`env.extract_browser_credentials`), which **deliberately** defaults to Firefox/Safari only and documents *why* it skips Chrome:

> Chrome is skipped because `security find-generic-password` triggers a macOS Keychain prompt that cannot be reliably suppressed.

Two code paths, opposite policies on the same sensitive operation. It also compounds on persistence: when no cookies are found (e.g. the user denies the prompt), setup persisted `FROM_BROWSER=auto`, so **every subsequent run re-probed Chrome and re-triggered the prompt**.

## Changes

- Add `env.cookie_extraction_browsers()` as the single source of truth for which browsers to probe, honoring `FROM_BROWSER` (default Firefox/Safari; Chrome opt-in via `chrome`/`auto`; `off` disables). Both the wizard and the steady-state path now use it, so they can't diverge again.
- `run_auto_setup()` honors that policy instead of hardcoding `"auto"`.
- `write_setup_config()` / the `setup` caller no longer persist `FROM_BROWSER=auto` when no browser yielded cookies — it's left unset so the safe default applies and the prompt doesn't recur.
- Behavior is unchanged for users who explicitly opt in with `FROM_BROWSER=auto|chrome`.

## Testing

- [x] `uv run pytest -q` — 1624 passed, 4 skipped
- Added `cookie_extraction_browsers()` unit tests + a regression test that first-run setup never requests Chrome unless `FROM_BROWSER=auto`.
- Updated `write_setup_config` expectations to the new (no-`auto`) default.

**Controlled before/after** (same machine, default config / no `FROM_BROWSER`, Chrome cookie DB present), intercepting at the Keychain-access boundary so no live dialog fires:

| code | browsers the wizard requests | reaches "Chrome Safe Storage" |
|------|------------------------------|-------------------------------|
| before | `['auto', 'auto']` | **yes** → dialog |
| after  | `['firefox', 'safari', ...]` | **no** |

## Related Issues

None filed — found while running the skill on macOS.

---

_Unrelated, noticed in passing (out of scope for this PR):_ `SKILL.md` Step 0 tells the host model to `Read skills/last30days/nux-wizard.md`, but that file isn't in the repo (it was never committed). Worth a look — it means there's no human-readable consent step before the cookie scan.

### DIFF
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -588,13 +588,15 @@ def main() -> int:
             return 0
         sys.stderr.write("Running auto-setup...\n")
         results = setup_wizard.run_auto_setup(config)
-        # Persist only the browser that actually yielded cookies. When none did,
-        # leave FROM_BROWSER unset so the safe default (Firefox/Safari, no
-        # Keychain prompt) applies instead of pinning "auto", which would make
-        # every subsequent run probe Chrome and re-trigger the prompt.
-        from_browser = None
-        if results.get("cookies_found"):
-            from_browser = next(iter(results["cookies_found"].values()))
+        # Persist FROM_BROWSER only when every service's cookies came from the
+        # SAME single browser — then we can fast-path future runs to it. If
+        # different services matched different browsers, or none matched, leave
+        # FROM_BROWSER unset: the safe default (Firefox/Safari) then covers all
+        # of them with no Keychain prompt. We deliberately do NOT pin "auto"
+        # here (it would re-probe Chrome and re-trigger the prompt) nor a single
+        # browser (it would silently skip the service that used the other one).
+        found_browsers = set(results.get("cookies_found", {}).values())
+        from_browser = found_browsers.pop() if len(found_browsers) == 1 else None
         setup_wizard.write_setup_config(env.CONFIG_FILE, from_browser=from_browser)
         results["env_written"] = True
         sys.stderr.write(setup_wizard.get_setup_status_text(results) + "\n")
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -588,10 +588,13 @@ def main() -> int:
             return 0
         sys.stderr.write("Running auto-setup...\n")
         results = setup_wizard.run_auto_setup(config)
-        from_browser = "auto"
+        # Persist only the browser that actually yielded cookies. When none did,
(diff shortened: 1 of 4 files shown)


An-idd:
Note: this overlaps in *symptom* with #541 but not in *cause* — they're complementary, not duplicates.

- **#541** fixes the **test suite** popping the dialog: `tests/test_chrome_cookies.py` mocked a wrapper the production path doesn't call, so `pytest` itself ran `security find-generic-password -s "Chrome Safe Storage"`.
- **This PR (#546)** fixes the **production** path: the first-run setup wizard probes browsers in `"auto"` mode (which includes Chrome), so a real first run pops the dialog regardless of tests.

Zero file overlap — #541 touches only `tests/test_chrome_cookies.py`; this touches `env.py` / `setup_wizard.py` / `last30days.py` / `test_setup_wizard.py`. Both are worth merging; together they close the Chrome Keychain prompt in both `pytest` and real usage.

An-idd:
Thanks @greptile-apps — addressed both notes in bf27664:

**1. Stale docstring (P2)** — fixed. `run_auto_setup`'s docstring no longer says "auto mode"; it now points at `cookie_extraction_browsers()` and the Firefox/Safari default.

**2. Multi-service browser persistence** — good catch on the comment overstating a 1:1 relationship; reworded. I did **not** take the suggested "persist `auto` when >1 browser" route, though: `FROM_BROWSER=auto` re-probes Chrome on every subsequent run, which is exactly the Keychain prompt this PR removes. Pinning a single browser is also wrong (it silently skips the service that matched the other one). Instead:

```python
found_browsers = set(results.get("cookies_found", {}).values())
from_browser = found_browsers.pop() if len(found_browsers) == 1 else None
```

So FROM_BROWSER is persisted only when every service agreed on one browser; otherwise it's left unset and the safe Firefox/Safari default covers all services with no prompt. Net effect: never re-introduces the Chrome dialog, never silently drops a service.

PR #547, opened 2026-06-11
fix(relevance,dedupe): CJK-aware tokenization for Chinese sources

An-idd:
## Summary

Relevance scoring and near-duplicate detection tokenize by splitting on whitespace (`str.split()`). Chinese has no spaces between words, so a whole Chinese sentence collapses into a single token — token-overlap relevance and Jaccard dedup effectively **stop working for Chinese-language content**. This already affects the existing **Xiaohongshu** source (and any future CJK source).

## Changes

- Add `lib/cjk.py` with `segment(text)`, which splits text into CJK and non-CJK runs:
  - Non-CJK (ASCII/Latin) runs keep the original `\w+` word behavior — **English is unchanged**.
  - CJK runs are segmented by **jieba when installed**, falling back to **character bigrams** when it is not. Bigrams are dictionary-free and still give robust overlap signal (e.g. query `大模型` → `{大模, 模型}` overlaps text `国产大模型评测`).
- jieba stays **optional** — present: used; absent: bigram fallback. It is never added to the hard dependency set, so the skill keeps its zero-dependency, install-anywhere property (`pyproject dependencies = []`).
- Wire into `relevance.tokenize()` and `dedupe` (`_tokenize` / `token_jaccard`); both also union a small Chinese stopword set into their existing English stopwords.

## Testing

- [x] `uv run pytest tests/test_cjk.py tests/test_relevance.py tests/test_dedupe_v3.py -q` — all green
- [x] Full suite passes (1627 passed, 4 skipped)
- New `tests/test_cjk.py`: segmentation, Chinese relevance match/no-match, English non-regression, Chinese near-duplicate detection.

Quick before/after (default build, no jieba installed → bigram path):

| query vs text | before | after |
|---|---|---|
| `国产大模型 测评` vs `国产大模型最新测评` | ~0.0 (one giant token) | **0.93** |
| `国产大模型 测评` vs `今天天气很好` | — | 0.0 (correctly rejected) |
| `react hooks` vs `guide to react hooks` (English) | 1.0 | **1.0** (unchanged) |

## Limitations (stopword behavior in the bigram fallback)

The Chinese stopword set is fully effective only on the **jieba** path, which segments real words. On the **zero-dependency bigram fallback** it is partial, by construction:

- Single-character stopwords (的 / 了 / 是 / 在 …) are never emitted as tokens (every CJK token is a 2-char bigram), and bigrams that *contain* a stopword char (e.g. `品的`, `的真`) survive as mild noise tokens.
- Two-character stopwords (这个 / 什么 / 可以 …) are filtered only when they happen to align to a bigram boundary.

This does **not** affect correctness: matching is driven by the query's tokens, so stopword-bearing noise bigrams only appear on the text side and slightly dilute precision — they don't cause false matches. Verified: `产品 测评` vs `这个产品的真实测评` → 0.87 (correct match), unrelated text → 0.0. The core win is that Chinese is segmented at all (whole-sentence-as-one-token was the actual bug); the stopword union is a secondary refinement that mainly pays off when jieba is installed.

## Related

Complements #514 (Bilibili adapter) and the existing Xiaohongshu source — both produce Chinese items that currently can't be deduped or ranked correctly. This is an independent, source-agnostic fix to the tokenization layer.

## Related Issues

None filed.

### DIFF
--- a/skills/last30days/scripts/lib/cjk.py
+++ b/skills/last30days/scripts/lib/cjk.py
@@ -1,0 +1,105 @@
+"""CJK-aware tokenization for relevance scoring and near-duplicate detection.
+
+The skill ships with zero hard dependencies (pyproject ``dependencies = []``)
+so it installs across 50+ Agent Skills hosts as plain Python. Chinese text has
+no whitespace word boundaries, so the original ``str.split()`` tokenizers in
+relevance.py / dedupe.py collapse a whole sentence into a single token and
+break token-overlap scoring and Jaccard de-duplication for Chinese sources
+(Xiaohongshu, Bilibili).
+
+``segment(text)`` fixes that. It splits text into maximal CJK and non-CJK runs:
+
+- Non-CJK (ASCII / Latin) runs keep the original ``\\w+`` word behaviour.
+- CJK runs are routed through jieba when it is installed (best quality), and
+  fall back to character bigrams when jieba is absent. Bigrams are a
+  dictionary-free segmentation that still gives robust overlap signal — e.g.
+  query "大模型" -> {大模, 模型} overlaps text "国产大模型评测" -> {..大模, 模型..}.
+
+jieba stays OPTIONAL: present -> used; absent -> bigram fallback. We never add
+it to the hard dependency set, preserving the install-anywhere property.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import List
+
+# CJK ideographs + Japanese kana + Korean hangul. The Chinese ideograph block
+# (一-鿿) and its extension-A (㐀-䶿) cover the cases we care
+# about; kana/hangul are included so mixed-language text degrades gracefully.
+_CJK_CHARS = r"㐀-䶿一-鿿豈-﫿぀-ヿ가-힯"
+_CJK_RE = re.compile(f"[{_CJK_CHARS}]")
+_CJK_RUN_RE = re.compile(f"[{_CJK_CHARS}]+")
+_LATIN_RE = re.compile(r"\w+")
+
+# High-frequency Chinese function words that dilute overlap signal, mirroring
+# the role of the English STOPWORDS sets in relevance.py / dedupe.py.
+CHINESE_STOPWORDS = frozenset(
+    {
+        "的", "了", "和", "是", "在", "我", "有", "也", "就", "不", "人", "都",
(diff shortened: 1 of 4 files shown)


An-idd:
Thanks @greptile-apps — went through all five. Three fixed in e0a4e90, two I'm intentionally leaving (with reasoning):

**1. jieba lazy-init thread-safety — fixed.** Replaced the lazy `_get_jieba()` + mutable globals with a one-time eager import at module load. Relevance scoring does run inside the pipeline's `ThreadPoolExecutor`, so the race was real; binding `_jieba` at import settles it before any worker thread runs. jieba guards its own dictionary build with an internal lock, so concurrent `cut()` is fine after that.

**2. Phrase bonus dead for Chinese — fixed (with a small correction).** It wasn't dead for *every* Chinese query — a single contiguous query (`国产大模型`) still substring-matches. It was dead specifically for **multi-token** queries (`国产大模型 测评`), where the space never appears in continuous source text. Fixed by retrying the containment with spaces stripped, gated on `has_cjk` so English stays space-sensitive (no spurious `reacthooks` match). Verified: contiguous CJK match now scores above the scattered case.

**3. Tests assumed jieba absent — fixed.** Good catch. The bigram-path tests now force `cjk._jieba = None` via a base class, so they're deterministic whether or not jieba is installed in CI. Added phrase-bonus and import-binding tests.

**4. `except Exception` too broad — keeping it, intentionally.** jieba is an optional enhancement; the contract is that *any* failure to load it (missing package, corrupted install, missing data files, a `setLogLevel` signature change across versions) must degrade to the bigram path rather than crash the skill. Narrowing to `ImportError` would let a corrupted-but-importable jieba throw `AttributeError` and take the whole run down. I documented this reasoning in a comment.

**5. Stopword-only short-string bigram inflation — acknowledged as a known limitation, not fixing here.** Correct that `我们的` → `{我们, 们的}` and single-char stopwords don't filter in the bigram path. It only matters for very short (1–4 char) function-word-only strings, and matching is query-driven so it doesn't cause false positives in relevance. I've written this up in the PR's "Limitations" section. A real fix needs stopword-aware bigram filtering, which is better as a follow-up if Xiaohongshu/Bilibili tags turn out to surface it in practice.

(Also noted #5's sub-point about `normalize_text` splitting CJK at full-width punctuation — agreed that's correct behavior; the double-`lower()` is harmless and I left it for minimal diff.)

PR #550, opened 2026-06-12
Fix Windows cp1252 read crashes and false permission warning

### LINKED ISSUE #549: Windows support: cp1252 `UnicodeDecodeError` on non-ASCII plan files, silent version-detection failure, and a false `should be 600` permission warning every session
### Summary

On Windows, Python's default text encoding is cp1252, not UTF-8. Several
read-side file operations omit encoding="utf-8", so they fail when the file
contains non-ASCII bytes (accented entity names like Nestlé, or emoji). The
skill itself writes UTF-8 (last30days.py uses write_text(..., encoding="utf-8")),
so these reads can choke on the skill's own output.

Context: the write side of this problem was already fixed across #117, #181,
#225 and #17. The skill_meta.py read fix (Bug 2 below) is also included in
the open PRs #426 and #522 — listed here for completeness. Bugs 1 and 3 are
not covered by any open issue or PR I could find.

On Windows, Python's default text encoding is `cp1252`, not UTF-8. Several
read-side file operations omit `encoding="utf-8"`, so they fail when the file
contains non-ASCII bytes (accented entity names like `Nestlé`, or emoji). The
skill itself writes UTF-8 (e.g. `last30days.py` uses `write_text(..., encoding="utf-8")`),
so these reads can choke on the skill's own output.

**Bug 1 — `--plan` file read crashes on non-ASCII (uncaught)**

`skills/last30days/scripts/last30days.py`, two sites:

```python
# ~line 328, parse_competitors_plan()
if os.path.isfile(plan_str):
    try:
        plan_str = open(plan_str).read()        # no encoding
    except OSError as exc:                        # does NOT catch UnicodeDecodeError
        sys.stderr.write(f"[CompetitorsPlan] Cannot read plan file: {exc}\n")
        raise SystemExit(2)

# ~line 641, main()
if os.path.isfile(plan_str):
    plan_str = open(plan_str).read()              # no encoding, no guard
```

`UnicodeDecodeError` subclasses `ValueError`, not `OSError`, so the `except OSError`
guard does not catch it — the process dies with an uncaught traceback when a plan
file contains accented/emoji entity or topic names (a common case for competitor
comparisons).

**Bug 2 — version detection silently fails on Windows**

`skills/last30days/scripts/lib/skill_meta.py`, `read_skill_version()`:

```python
try:
    text = skill_md_path.read_text()              # no encoding
except (OSError, UnicodeDecodeError):
    return None
```

`SKILL.md` contains emoji (`✓`, `⏳`, `⚡`). Under cp1252 this raises
`UnicodeDecodeError`, which is caught and turned into `None` — so version detection
silently breaks on Windows rather than crashing.

**Bug 3 — false `should be 600` permission warning on every Windows session**

`hooks/scripts/check-config.sh`, `check_perms()`:

```bash
perms=$(stat -c '%a' "$file" 2>/dev/null || stat -f '%Lp' "$file" 2>/dev/null || echo "")
if [[ -n "$perms" && "$perms" != "600" && "$perms" != "400" ]]; then
    echo "/last30days: WARNING — $file has permissions $perms (should be 600)."
```

On Git-for-Windows, `bash`/MSYS mounts run in `noacl` mode: `stat -c '%a'` always
returns `644` (and `group=UNKNOWN`) for every regular file, regardless of the real
NTFS ACL. So this check fires a false `WARNING — ... should be 600` on every session
start for Windows users, and `chmod 600` cannot fix it (it's a no-op under noacl).
Windows file permissions are governed by ACLs (`icacls`), not POSIX mode bits.

Fix: skip the POSIX permission check on MSYS/MinGW/Cygwin:

```diff
   if [[ ! -f "$file" ]]; then return; fi
+  # Git-for-Windows / MSYS / Cygwin run stat in noacl mode (always 644),
+  # so this POSIX check is a false positive. Windows perms use ACLs.
+  case "$(uname -s 2>/dev/null)" in
+    MINGW*|MSYS*|CYGWIN*) return ;;
+  esac
   local perms
```

**Fix (Bugs 1 & 2)**

Add `encoding="utf-8"` to all three reads:

```diff
-            plan_str = open(plan_str).read()
+            plan_str = open(plan_str, encoding="utf-8").read()
```
```diff
-                plan_str = open(plan_str).read()
+                plan_str = open(plan_str, encoding="utf-8").read()
```
```diff
-        text = skill_md_path.read_text()
+        text = skill_md_path.read_text(encoding="utf-8")
```

**Optional hardening (not required for the crash fix)**

- In `parse_competitors_plan`, broaden the guard to also catch decode errors,
  e.g. `except (OSError, ValueError) as exc:` ,
  so a malformed/mis-encoded plan reports cleanly instead of crashing.
- A few non-ASCII-safe writes remain (`evaluate_search_quality.py` `summary.md`
  via `write_text(...)` without encoding, `last30days.py:551` `last-run.json`).
  These currently emit ASCII (`json.dumps` default), but adding `encoding="utf-8"`
  there would future-proof them.

**Repro**

```bash
# On Windows (cp1252 default):
printf '{"Nestl\xc3\xa9": {"subreddits": ["food"]}}' > plan.json   # UTF-8 accented key
python skills/last30days/scripts/last30days.py --plan plan.json ...
# -> UnicodeDecodeError traceback before the encoding fix
```

---

## Suggested PR

**Branch:** `fix/windows-support`
**Title:** Fix Windows cp1252 read crashes and false permission warning
**Commit message:**

```
fix(windows): utf-8 reads + skip noacl permission check

- last30days.py: open(plan_str) -> encoding="utf-8" (2 sites). The plan
  file holds entity/topic names; UnicodeDecodeError (a ValueError) was not
  caught by the surrounding except OSError, crashing on non-ASCII input.
- skill_meta.py: read_text() -> encoding="utf-8". SKILL.md has emoji;
  cp1252 made version detection silently return None on Windows.
- check-config.sh: skip the POSIX 600 check on MSYS/MinGW/Cygwin, where
  stat runs in noacl mode and always reports 644 regardless of the NTFS
  ACL, firing a false "should be 600" warning every session.
```

### Steps to Reproduce

Bug 1 (plan-file read crash):
1. Windows 11, default locale codec cp1252 (no PYTHONUTF8=1)
2. Create a UTF-8 plan file with a non-ASCII entity name:
   printf '{"Nestl\xc3\xa9": {"subreddits": ["food"]}}' > plan.json
3. Run: python skills/last30days/scripts/last30days.py --plan plan.json ...
4. The plan file is opened without encoding="utf-8" -> UnicodeDecodeError
   traceback (not caught by the surrounding `except OSError`)

Bug 2 (hook false warning):
1. Git for Windows / MSYS environment, ~/.config/last30days/.env present
2. Start a Claude Code session (session-start hook runs check-config.sh)
3. Warning appears: ".env has permissions 644 (should be 600)"
4. The suggested fix `chmod 600` is a no-op: MSYS mounts NTFS in noacl
   mode, so stat always reports synthetic 644 regardless of the real
   NTFS ACL (verified: icacls-restricted file still reports 644)

### Expected Behavior

Bug 1: The plan file is read as UTF-8 (the skill itself writes UTF-8),
so comparison mode works with accented/emoji entity names. A malformed
plan file should produce the clean "[CompetitorsPlan] Cannot read plan
file" error path, not an uncaught traceback.

Bug 2: No permission warning on Git-for-Windows/MSYS/Cygwin, where
POSIX mode bits are synthetic and the suggested chmod fix cannot work.
File permissions on Windows are managed via NTFS ACLs (icacls).

### Error / Traceback

```text

```

### Install Method

Claude Code plugin

### OS

_No response_

23241a6749:
Fixes #549

**Changes**

- **last30days.py** — both `--plan` / `--competitors-plan` file reads now specify `encoding="utf-8"`. `UnicodeDecodeError` (a `ValueError`) was not caught by `except OSError`, producing an uncaught traceback on non-ASCII plan file content. The exception guard has been widened to `except (OSError, UnicodeDecodeError)`.
- **skill_meta.py** — `read_text()` now passes `encoding="utf-8"`. SKILL.md contains emoji; cp1252 made version detection silently return `None` on Windows.
- **check-config.sh** — `check_perms()` now skips the POSIX `600` permission check on MSYS/MinGW/Cygwin, where `stat` runs in noacl mode and always reports `644` regardless of the NTFS ACL, firing a false `should be 600` warning every session.
- **Tests** — added `test_file_with_non_ascii_content` to verify UTF-8 plan file reading.
- **CHANGELOG + CONTRIBUTORS** — updated.


tmchow:
Thanks! The Windows cp1252 / check-config / UTF-8 plan-file fixes landed in #578. Closing as duplicate so we don't double-merge the same changes.

PR #455, opened 2026-05-28
fix(threads): use 'query' param for ScrapeCreators v1 search (regression since PR #393)

dzivkovi:
## Summary

ScrapeCreators' `/v1/threads/search` endpoint requires `params={"query": ...}`, not `params={"keyword": ...}`. The code in `lib/threads.py` sends `keyword`, which causes the API to return HTTP 400 on every Threads request. **Every install with `SCRAPECREATORS_API_KEY` configured has been silently losing Threads since 2026-05-15** — the regression was introduced by PR #393 ("refactor: drop requests dep, route all providers through lib/http urllib wrapper") at commit 80a1a47e.

All other ScrapeCreators v1 search endpoints (Pinterest, TikTok, Instagram) use `query` consistently. Only Threads regressed.

## Reproduction

```bash
# Confirms the fix:
curl "https://api.scrapecreators.com/v1/threads/search?query=tourism" \
  -H "x-api-key: $SCRAPECREATORS_API_KEY"
# Returns 200 OK with {"success": true, "credits_remaining": N, "posts": [...]}

# Confirms the bug (current main):
curl "https://api.scrapecreators.com/v1/threads/search?keyword=tourism" \
  -H "x-api-key: $SCRAPECREATORS_API_KEY"
# Returns HTTP 400 Bad Request
```

## What changed

| File | Change |
|---|---|
| `skills/last30days/scripts/lib/threads.py` | One-line fix: `params={"keyword"}` → `params={"query"}`. Inline anti-regression comment points at the new `docs/solutions/` entry. |
| `tests/test_threads.py` (new, 17 tests) | Load-bearing: `TestSearchThreadsRequestShape::test_uses_query_param_not_keyword` pins the API contract with `assertIn("query")` + `assertNotIn("keyword")`. `TestSearchThreadsResponseShape` exercises the real `{"success", "credits_remaining", "posts"}` envelope. Convention mirrors `tests/test_bluesky.py`. |
| `docs/solutions/design-patterns/scrapecreators-v1-search-uses-query-param-2026-05-27.md` (new) | Captures the cross-source convention, the failure mode, detection cURL commands, and a 5-step pattern checklist for future ScrapeCreators source integrations. Cross-references the two compounding observability gaps that hid the regression. |

## Verification

**TDD:** The regression test failed against the unfixed code, passes after the one-line fix. All 17 new tests pass.

[...]

PR #558, opened 2026-06-14
fix(threads): use 'query' param and cap to 1-2 word queries

cypres0099:
## Problem

The Threads source has been **dark for all users**. The ScrapeCreators Threads search endpoint changed its contract and now rejects the `keyword` parameter the adapter sends:

```json
{"success":false,"error":"bad_request","message":"You must provide a 'query'"}
```

HTTP 400 on every call. `search_threads` catches it and returns `{"items": [], ...}`, so it fails silently — runs just show `Threads: 0 results`.

## Fix

**1. Correct the param name** — `params={"query": ...}` instead of `{"keyword": ...}`.

**2. Cap query length.** Even with the right param, the endpoint only returns hits for short (1–2 word) queries, and the planner feeds sources verbose boolean strings (e.g. `"skincare routine favorites OR recommendations"`). Empirically against the live API:

| query | posts |
|---|---|
| `skincare routine favorites or` | 0 |
| `skincare routine favorites` (3 words) | 0 |
| `skincare routine` | 19 |
| `skincare` | 20 |
| `AI video generator tutorial` (4 words) | 0 |
| `AI video` | 20 |

So `_extract_core_subject` now strips boolean operators (`or`/`and`) and caps to the two most salient words (`max_words=2`).

## Verification

Tested end-to-end against the live API on v3.3.2. Threads now returns parsed posts with correct dates, engagement, and relevance for both single-word topics and verbose planner queries (e.g. `"AI video tools"` → 4 posts, `"skincare"` → 2 posts).

- `python3 -m py_compile` passes.
- No existing test asserts the `keyword` param (the only `*thread*` test is `test_competitors_plan_threading.py`, unrelated Python threading).

### DIFF
--- a/skills/last30days/scripts/lib/threads.py
+++ b/skills/last30days/scripts/lib/threads.py
@@ -29,15 +29,22 @@ def _log(msg: str):
 
 
 def _extract_core_subject(topic: str) -> str:
-    """Extract core subject from verbose query for Threads search."""
+    """Extract core subject from verbose query for Threads search.
+
+    The ScrapeCreators Threads keyword endpoint only returns hits for short
+    (1-2 word) queries; 3+ words or leaked boolean operators (the planner
+    emits "A OR B") return zero. Strip boolean operators and cap to the two
+    most salient words.
+    """
     from .query import extract_core_subject
     _THREADS_NOISE = frozenset({
         'best', 'top', 'good', 'great', 'awesome',
         'latest', 'new', 'news', 'update', 'updates',
         'trending', 'hottest', 'popular', 'viral',
         'practices', 'features', 'recommendations', 'advice',
+        'or', 'and',
     })
-    return extract_core_subject(topic, noise=_THREADS_NOISE)
+    return extract_core_subject(topic, noise=_THREADS_NOISE, max_words=2)
 
 
 def _parse_date(item: Dict[str, Any]) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


cypres0099:
Updated this branch to address Greptile's all-noise fallback edge case.

What changed:
- Threads query extraction now applies a final 2-word cap even when the shared helper falls back to the original all-noise text.
- Added `tests/test_threads.py` coverage for both a verbose planner query and the all-noise fallback case.

Verification:
- `uv run python -m pytest tests/test_threads.py tests/test_query.py -q --tb=short` -> passed
- `OPENAI_API_KEY= ANTHROPIC_API_KEY= GEMINI_API_KEY= GOOGLE_API_KEY= PERPLEXITY_API_KEY= SCRAPECREATORS_API_KEY= uv run pytest` -> 1619 passed, 4 skipped

dzivkovi:
Nice - this is more complete than my #455 (you also handle the verbose/boolean-query case that returns zero hits). Closing mine in favor of yours. For context, the root cause is that ScrapeCreators' v1 search endpoints take `query`, not the `keyword` their docs sidebar shows - `keyword=` returns HTTP 400. I documented the why [here](https://github.com/dzivkovi/last30days-skill/blob/main/docs/solutions/design-patterns/scrapecreators-v1-search-uses-query-param-2026-05-27.md) if it's useful; happy to fold it in or send a small docs-only follow-up. Thanks for picking this up.

PR #580, opened 2026-06-17
feat: keyless web coverage and honest source degradation

mvanhorn:
## Why

The engine runs fully keyless except for one capability: general web search. With no paid web key set, the `grounding` source silently disappears and the brief is built from social sources only. Several sources also hide failures, returning empty on error so a source dropping from 20 results to 1 looks healthy.

This closes both gaps and a few adjacent keyless holes, without changing behavior for users who already have keys.

## What changed

- Keyless web tier (host-gated): a free URL-to-markdown fetch (Jina Reader) and a keyless web search floor (DuckDuckGo, optional SearXNG fallback). It is the floor of a strict ladder: host-native search > paid backend > keyless. Capability-gated by `LAST30DAYS_NATIVE_SEARCH` (set by the SKILL.md path only when the runtime has a native search tool), so Claude Code and other native-search hosts never fall to the free tier or preempt a paid backend; OpenClaw-style hosts and headless/cron do.
- Honest degradation: a typed source-health probe (missing / broken / timeout / ok) and a fix so partial failures surface as a distinct "degraded" warning instead of being silently dropped.
- Throttled keyless Reddit tiers (token-bucket) so a broad multi-query run cannot stampede the endpoints under the parallel fan-out.
- Unauthenticated GitHub fallback: GitHub search now works without a token via the anon REST tier (capped to the low-rate limit).
- Caption-free transcription module (compress to chunk to provider-fallback Whisper). Groundwork: shipped and tested, not yet auto-invoked by the engine.
- Secret hygiene: the `.env` is now written 0o600 atomically.

## Config added

- `LAST30DAYS_NATIVE_SEARCH` - host-native-search signal (suppresses the keyless floor).
- `LAST30DAYS_SEARXNG_URL` - optional keyless-search fallback instance.
- `GROQ_API_KEY` - Whisper provider for the transcription module.

## Testing

Full suite green. Each unit ships with its own test file (fetch, search, health, reddit throttle, github unauth, transcription, secret hygiene).

## Notes

- Plan: `docs/plans/2026-06-17-001-feat-keyless-coverage-honest-degradation-plan.md`.
- Per the beta-channel convention this would normally land via `/last30days-beta` first; opening here for review.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -158,15 +158,22 @@ When you invoke `/last30days` from Claude Code, Codex, or Gemini, the host model
 
 ## Web search backend priority
 
-Used by `--auto-resolve` (when WebSearch isn't available from the host) and Step 2 supplements. Auto-detect priority (override per-run with `--web-backend=<name>`):
+The search-source preference ladder, strict best-to-floor:
 
-1. **Brave** - `BRAVE_API_KEY`
-2. **Exa** - `EXA_API_KEY`
-3. **Serper** - `SERPER_API_KEY`
-4. **Parallel** - `PARALLEL_API_KEY`
-5. **Host's native WebSearch** - Claude Code, Codex, Gemini all have one built in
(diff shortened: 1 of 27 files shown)


tmchow:
Landed on main.

mvanhorn:
Thanks for reviewing and merging, @tmchow. Keyless web coverage with honest source degradation keeps the engine usable without a paid key.

PR #596, opened 2026-06-18
fix(setup): use $HOME-relative PATH hints for off-path Digg CLI

tmchow:
## Summary
- Follow-up to #590: when Digg CLI is installed but not on PATH, setup status now shows copy-pasteable `$HOME/...` directory hints derived from the probed binary path
- Adds `_digg_bin_dir_hint()` and a generic fallback when `digg_path` is missing
- Updates off-path tests to assert `$HOME/.local/bin` and `$HOME/go/bin` instead of machine-specific absolute paths

## Test plan
- [x] `uv run pytest tests/test_setup_wizard.py -q`


jenniferscott11718-byte:
Thank you

PR #597, opened 2026-06-18
fix(store): serialize concurrent writers and upsert on URL conflict

iliaal:
## Before
store.py documents WAL-mode "safe concurrent access (cron + user)", but two paths broke that promise. `_connect` never set `busy_timeout`, so SQLite's default 0ms made a second writer raise "database is locked" the instant it hit the write lock instead of waiting. And `store_findings` ran a dedup SELECT, then a separate INSERT; `findings.source_url` is UNIQUE and the read is not atomic with the write, so two runs could both see a URL as missing and both insert it. The second commit hit IntegrityError and rolled back the whole batch, dropping every finding in it and skipping the run counters.

## After
`_connect` sets `PRAGMA busy_timeout=5000`, so a contending writer waits for the lock. `store_findings` inserts with `ON CONFLICT(source_url) DO UPDATE`, mirroring the re-sighting path (bump last_seen and sighting_count, keep the higher engagement). A racing duplicate upserts instead of crashing.

## Scope rationale
- Not changing `init_db()`, which re-runs the schema DDL on every add_topic/get_setting/set_setting call. That widens the write-lock window but is a contention concern, not a correctness one, and the 5s busy_timeout absorbs it. Left for a separate follow-up.

## Test plan
- [x] `uv run pytest tests/test_store.py` passes (27 tests)
- [x] New `test_store_findings_upserts_on_concurrent_duplicate_url` forces a stale dedup read. Confirmed it raises `IntegrityError: UNIQUE constraint failed: findings.source_url` without the fix and passes with it.

### DIFF
--- a/skills/last30days/scripts/store.py
+++ b/skills/last30days/scripts/store.py
@@ -193,6 +193,10 @@ def _connect(db_path: Optional[Path] = None) -> sqlite3.Connection:
     conn.execute("PRAGMA journal_mode=WAL")
     conn.execute("PRAGMA synchronous=NORMAL")
     conn.execute("PRAGMA foreign_keys=ON")
+    # WAL lets readers coexist with one writer, but two writers (cron + user)
+    # still contend for the write lock. Default busy_timeout is 0, so the loser
+    # raises "database is locked" instantly; wait instead.
+    conn.execute("PRAGMA busy_timeout=5000")
     return conn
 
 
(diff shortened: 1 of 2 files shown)


iliaal:
Good catch. The upsert closed the data-loss race but left `new_count = len(insert_rows)`, which counted conflict-resolved rows as new and inflated `research_runs.findings_new`. Re-derived the split after the write: an inserted URL whose `sighting_count` is now > 1 was a conflict, so it counts as an update. The regression test now asserts the run counters too. Addressed in 87d2be3.

PR #599, opened 2026-06-18
fix(cookie_extract): close world-readable window on temp cookie copy

iliaal:
## Before
`_query_cookies_db` copies the browser cookie database, which holds live X auth_token and ct0 session secrets, into a system temp file, then chmods it 0600. `shutil.copy2` copies the source file's mode onto the destination. Firefox `cookies.sqlite` is commonly 0644, looser on WSL `/mnt/c` mounts, so between the copy and the chmod the decrypted secrets sat world-readable in shared `/tmp`, readable by any other local user who won the race.

## After
The copy uses `shutil.copyfile`, which writes content only and leaves the 0600 perms that `mkstemp` created. The temp file is never readable by others. The existing chmod stays as defense in depth.

## Test plan
- [x] `uv run pytest tests/test_cookie_extract.py` passes
- [x] New `test_temp_cookie_copy_never_world_readable` asserts the copy is 0600 the instant it exists. Confirmed it reports 0644 without the fix.

### DIFF
--- a/skills/last30days/scripts/lib/cookie_extract.py
+++ b/skills/last30days/scripts/lib/cookie_extract.py
@@ -157,7 +157,12 @@ def _query_cookies_db(
     tmp_path = None
     try:
         tmp_fd, tmp_path = tempfile.mkstemp(suffix=".sqlite")
-        shutil.copy2(str(db_path), tmp_path)
+        # mkstemp creates the file 0600. copy2 would copy the source's mode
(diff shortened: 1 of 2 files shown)


iliaal:
Added the `@pytest.mark.skipif(os.name == "nt", ...)` guard to match the sibling `test_temp_cookie_db_copy_is_owner_only`. `mkstemp` is 0o666 on Windows and the POSIX permission exposure doesn't apply there. Addressed in 94f3ace.

PR #600, opened 2026-06-18
fix(watchlist): validate webhook scheme and match Slack host exactly

iliaal:
## Before
`_deliver_findings` chose the Slack path with `"hooks.slack.com" in channel`, an unanchored substring test that ran before any scheme check, while the generic branch required `https://`. A `delivery_channel` like `http://evil.example/hooks.slack.com` was treated as Slack and POSTed in cleartext to whatever host the URL named, leaking the notification payload while the operator believed Slack was configured.

## After
`_deliver_findings` parses the channel, requires an https scheme, and matches Slack on the exact hostname (`parsed.hostname == "hooks.slack.com"`). A non-https channel is reported on stderr instead of being dropped without a word.

## Test plan
- [x] `uv run pytest tests/test_watchlist_delivery.py` passes (17 tests)
- [x] Two new tests cover the cleartext-bypass URL and a URL carrying the Slack host only in its path. Both confirmed to fail without the fix.

### DIFF
--- a/skills/last30days/scripts/watchlist.py
+++ b/skills/last30days/scripts/watchlist.py
@@ -8,6 +8,7 @@ import json
 import subprocess
 import sys
 import time
+import urllib.parse
 from pathlib import Path
 
 SCRIPT_DIR = Path(__file__).parent.resolve()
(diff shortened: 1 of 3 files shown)


iliaal:
Added the same https scheme check in `cmd_config`, so a non-https `delivery_channel` is rejected at write time with a clear message instead of surfacing only at delivery. Addressed in e603c7b.

PR #601, opened 2026-06-18
fix: fail fast and rank correctly on valid-looking input across CLI paths

iliaal:
## Before
Three spots silently did the wrong thing on input that looked fine. last30days.py only warned to stderr on malformed `--plan` JSON, then ran a full, API-consuming research with the internal planner. evaluate_search_quality.py keyed its judgment cache on the topic slug alone, so a rerun with a different `--judge-model` returned the prior model's grades and skewed precision@k and nDCG. briefing.py sliced `this_week[:5]` with a comment claiming the list was sorted by engagement, but `get_new_findings` returns first_seen DESC, so the weekly digest headlined the most recent items rather than the highest-engagement ones.

## After
Malformed `--plan` JSON raises SystemExit(2), matching the `--plan` file-read branch and parse_competitors_plan. The judgment cache stores its judge model and re-judges on a mismatch. The weekly digest sorts by engagement before slicing, matching the daily path.

## Scope rationale
- Three unrelated one-spot fixes, grouped because they share a failure mode: each accepts plausible input and produces a wrong result with no error. Each has its own regression test and could split into separate PRs if you prefer.

## Test plan
- [x] `uv run pytest tests/test_cli_v3.py tests/test_briefing_v3.py tests/test_evaluator_v3.py` passes
- [x] Each fix has a regression test confirmed to fail on the prior behavior (run silently completing, stale grades returned, recency-ordered top 5).

### DIFF
--- a/skills/last30days/scripts/briefing.py
+++ b/skills/last30days/scripts/briefing.py
@@ -188,7 +188,15 @@ def generate_weekly() -> dict:
             "this_week_engagement": this_engagement,
             "last_week_engagement": last_engagement,
             "engagement_change_pct": round(engagement_change, 1),
-            "top_findings": this_week[:5],  # Top 5 by engagement (already sorted)
+            # get_new_findings returns first_seen DESC, so sort by engagement
+            # before slicing — otherwise the digest headlines the most recent
+            # items, not the highest-engagement ones (the daily path keys on
+            # engagement too).
+            "top_findings": sorted(
+                this_week,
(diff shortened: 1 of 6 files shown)


iliaal:
Added a stderr warning when a different-model cache is discarded and no Gemini API key is set to re-judge, so the zero-metrics case isn't silent. The normal not-configured path (no cache) stays quiet. Addressed in faf8782.

PR #604, opened 2026-06-18
ci: harden release supply chain

shaanmajid:
## Summary

Hardens the release and Scorecard workflows by tightening job permissions, pinning GitHub Actions to SHAs, replacing the old provenance action with `actions/attest`, and adding provenance attestations for the Claude Desktop `.mcpb` release artifacts.

The original CI/CD hardening PR from ~2 weeks ago #472 reported 0 findings from zizmor when it was created, but it went stale as upstream workflows changed afterward; after being merged today it reports several extant zizmor findings. This PR refreshes that work against current `main` and is also clean under online zizmor locally. To avoid the same drift happening again, I recommend merging this relatively soon; if it goes stale or checks start diverging again, feel free to ping me and I can refresh it back to clean!

## Changes

- Set top-level release workflow permissions to `{}` and move required scopes to individual jobs.
- Reduce build jobs from `contents: write` to `contents: read`; keep `contents: write` only on the release publishing job.
- Pin release and Scorecard workflow actions to commit SHAs with version comments.
- Replace `actions/attest-build-provenance` with `actions/attest`.
- Add `.mcpb` artifact attestations for each matrix-built Claude Desktop bundle.
- Move tag interpolation through shell env vars to avoid direct expression injection in shell.
- Disable `setup-go` cache in the release build and read Go version from `mcp/go.mod`.
- Add Dependabot coverage for the `/mcp` Go module with cooldown.

## Testing

- [x] Ran `actionlint .github/workflows/release.yml .github/workflows/scorecard.yml .github/workflows/validate.yml`
- [x] Ran zizmor online checks on repository
- [x] Confirmed token-backed zizmor reported no findings and did not print the GitHub token.
- [x] Verified pinned action SHAs resolve to the intended upstream tags.

## Related Issues

Relates to the earlier supply-chain hardening / SHA pinning work in #472.

Follow-up supply-chain work intentionally left out of this PR:

- Revisit `GOSUMDB=off` in the `go install github.com/mvanhorn/cli-printing-press/...` release step.
- Revisit Scorecard’s existing `permissions: read-all`.
- Add Go vulnerability/audit coverage such as `govulncheck`.
- Consider splitting release into separate build and publish workflows. That would better isolate `id-token` / `attestations` permissions from the final `contents: write` release-publishing step.


tmchow:
Fixed in e3a481df370135257aa6f54c5af5d5cfbb007a78: both release.yml upload-artifact steps now use the same actions/upload-artifact pin as scorecard.yml, `043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1`.

Verified against upstream tags with `git ls-remote https://github.com/actions/upload-artifact.git refs/tags/v7.0.1 refs/tags/v4.6.2`: `ea165f8...` is v4.6.2 and `043fb46...` is v7.0.1. Also ran `git diff --check` and `uv run pytest tests/test_scorecard_workflow.py`.

PR #610, opened 2026-06-18
feat(x): FROM + ABOUT lanes, both weighted (fix from:-AND-bug, add mentions)

mvanhorn:
## Why

PR4 of the X-search plan — the explicit "tweets **FROM** the person with weight + tweets **TO/ABOUT** them with weight" requirement. Two verified failures:

1. **FROM lane returned ~0.** `search_handles` built `from:{handle} {core_topic} since:` — X search is literal AND, so for a person whose topic ≈ their name it only matched tweets where they wrote their own name. The unfiltered branch was unreachable (topic always truthy).
2. **No ABOUT lane.** The pipeline only ran `from:` and keyword searches; tweets *mentioning* the person were never collected.

## What changed

- **FROM lane (U7):** `search_handles` pulls the real timeline via `from:handle since:` and uses the topic for **relevance ranking only**, never AND'd into the query. Engagement-weighted downstream via normalization.
- **ABOUT lane (U8):** new `search_mentions` queries `@handle since:` (tweets mentioning the account), **excludes the handle's own tweets** (those are the FROM lane), engagement-weighted, deduped against FROM by URL at normalize time.
- `pipeline` merges both lanes into the X pool.

## Testing

- `TestXFromAndAboutLanes` (4) — FROM drops the topic-AND; ABOUT queries `@handle`; ABOUT excludes own tweets; empty-safe. Pipeline regression green; full suite green (except 3 pre-existing `sed`-env failures).
- **Live:** `from:mvanhorn` → his real posts; `@mvanhorn` → mentions (Trevin's tweet about him), own tweets excluded.

Independent of PR1 (#607) / PR2 (#608) / PR3 (#609). PR3 and this both touch `search_handles` — expect a trivial merge (PR3 adds a success-log line; this rewrites the query) the second-merged PR resolves. Plan: `docs/plans/2026-06-17-004-...`.


mvanhorn:
Addressed the Greptile findings:
- **Issue 1 (lost FROM results):** split into independent try/except per lane — an ABOUT-lane failure no longer discards the FROM-lane results.
- **Issue 2 (twitter.com):** own-tweet exclusion now matches both `x.com` and `twitter.com` permalinks (added a regression test).
- **Issue 3 (log verb):** keeping `Searching:` deliberately — it's the established field-debug grep token shared with the keyword and FROM/handle lanes (`grep '[Bird] Searching:'` should catch all lanes); the result count after it conveys completion.

PR #613, opened 2026-06-18
feat: X first-party authorship grounding + interaction-signal ranking

mvanhorn:
Fixes the X ranking path systematically burying the highest-signal class of item for a person topic: **the subject's own posts**. The entity-miss demotion keys on whether a post's *text* contains the subject's name, but nobody writes their own name in their own tweets, so every `from:{handle}` result took −25 (rerank) **and** −20 (final) and landed at ~0.

Implements the plan in `docs/plans/2026-06-17-005` (6 units, dependency-ordered):

- **U1+U2 — first-party authorship exemption.** Thread the run's resolved-handle set (`--x-handle`, `--x-related`, github-user) into rerank scoring; a post authored by one of those is exempt from both entity-miss penalties (+ a small bounded authorship credit). Third-party collision-noise suppression is untouched (regression-guarded).
- **U3 — engagement rescue floor.** A high-engagement X post that is first-party or entity-grounded gets a `final_score` floor scaled by its percentile within the run's X pool. Off-topic name-collision posts are explicitly excluded, so viral collisions stay buried.
- **U4 — FROM-lane cap 3 → 8.** The subject's own timeline is the best source for a person topic; named caps (`FROM_LANE_COUNT_PER=8`, mention/related stay modest).
- **U5 — interaction signal.** A first-party post directed at another account (reply / leading @mention, parsed from text at ingest) is floated regardless of like-count and tagged `interaction:→@handle` in the EVIDENCE block. Purely structural; no names.
- **U6 — LAW 10.** Hoists the first-party rule into SKILL.md's guaranteed-loaded contract band.

Structural guarantee (HTD flowchart in the plan): the rescue floor and interaction float are reachable only from the first-party or entity-grounded branches, never from the entity-miss branch — so all the disambiguation work from v3.5.0 stays intact.

Tests: +60 across `test_rerank_v3.py`, `test_bird_x.py`, `test_pipeline_v3.py` (authorship exemption, rescue-floor gating, interaction classification, leading-mention parse, FROM-cap). Full suite green.


mvanhorn:
Addressed both findings in 438ca26:

- **LLM-path gap (the 3/5 reason):** added a deterministic `FIRST_PARTY_FLOOR` post-pass so a first-party post clears the zero band regardless of scoring path — the backstop for the case where a paid rerank provider caps the post via the grounding hint. Also taught the LLM directly: `resolved_handles` now threads into `_build_prompt`, first-party candidates are marked (`author: @handle` + `first_party: true`), and the grounding hint carries an explicit EXCEPTION so the model scores them on merits instead of capping at ≤30.
- **P2 docstring grammar:** applied.

+6 tests (floor lift-only / third-party-untouched / empty-handles noop / prompt markers). Full suite green.

PR #619, opened 2026-06-18
fix: wrap pwd import in try/except for Windows compatibility

365diascollaboration-prog:
## What

`env.py` does a lazy `import pwd` inside `_load_keychain_vars()`. The `pwd` module is Unix-only and raises `ImportError` on Windows. While the function already returns early when no `security` binary is found (the common Windows case), a `security` binary in PATH would trigger the import and crash.

## Fix

Wrap the `import pwd` block in `try/except (ImportError, AttributeError)` with a `getpass.getuser()` fallback — stdlib cross-platform, no new dependencies.

## Scope

Single file, 4-line change. No behavior change on macOS/Linux where `pwd` is always available.

### DIFF
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -130,11 +130,15 @@ def _load_keychain(keys: list[str]) -> dict[str, str]:
         return {}
 
     import subprocess
-    import pwd
-    # USER can be unset under sudo, in Docker without --env USER, or in some CI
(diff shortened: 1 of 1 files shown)


365diascollaboration-prog:
Closing after Greptile's review — the existing `platform.system() != 'Darwin'` guard on line 124 already makes this code path unreachable on Windows, so the try/except adds no real protection. The fix is harmless but unnecessary. Thanks for the sharp review.

PR #624, opened 2026-06-18
fix: guard os.killpg/getpgid with hasattr for Windows compatibility

### LINKED ISSUE #588: YouTube transcripts crash on Windows — os.killpg AttributeError in subproc.py
## Bug

On Windows, **all** YouTube transcript fetches fail with:

```
[YouTube] Unexpected transcript error for <id>: AttributeError: module 'os' has no attribute 'killpg'
```

In `scripts/lib/subproc.py`, `run_with_timeout` calls this in the `subprocess.TimeoutExpired` handler:

```python
os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
```

`os.killpg`/`os.getpgid` are POSIX-only. On Windows they raise `AttributeError`, which is **not** in the surrounding `except (ProcessLookupError, PermissionError, OSError)`, so the error escapes the cleanup path and kills the transcript fetch. Result: YouTube videos return metadata (views/comments) but **zero transcripts** on Windows.

Note: spawn already guards correctly with `preexec = os.setsid if hasattr(os, "setsid") else None` — the kill path just missed the same guard.

## Suggested fix

Guard the group-kill with `hasattr` (mirroring the `os.setsid` guard) and add `AttributeError` to the caught exceptions, falling back to `proc.kill()`:

```python
    except subprocess.TimeoutExpired:
        try:
            if hasattr(os, "killpg") and hasattr(os, "getpgid"):
                os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
            else:
                proc.kill()
        except (ProcessLookupError, PermissionError, OSError, AttributeError):
            proc.kill()
        proc.wait(timeout=5)
        raise SubprocTimeout(f"Command {cmd[0]} timed out after {timeout}s")
```

Verified on Windows (Python 3.12): with this change, `fetch_transcript` returns transcript text where it previously crashed. Happy to PR.

365diascollaboration-prog:
## Problem

Closes #588. Related to #156.

`run_with_timeout` in `subproc.py` already guards the spawn path correctly:

```python
preexec = os.setsid if hasattr(os, "setsid") else None
```

But the timeout cleanup path calls `os.killpg`/`os.getpgid` unconditionally:

```python
except subprocess.TimeoutExpired:
    try:
        os.killpg(os.getpgid(proc.pid), signal.SIGTERM)  # ← AttributeError on Windows
    except (ProcessLookupError, PermissionError, OSError):
        proc.kill()
```

`os.killpg` and `os.getpgid` are POSIX-only. On Windows they don't exist, so any subprocess timeout raises `AttributeError`, which is not caught — it escapes the cleanup path and surfaces to callers as an unexpected exception. Result: all YouTube transcript fetches that hit a timeout fail with `AttributeError` instead of gracefully returning no transcript.

## Fix

Mirror the existing `os.setsid` guard pattern: check `hasattr` before the group-kill and fall back to `proc.kill()`. Also add `AttributeError` to the caught exception set as a backstop.

```python
except subprocess.TimeoutExpired:
    try:
        if hasattr(os, "killpg") and hasattr(os, "getpgid"):
            os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
        else:
            proc.kill()
    except (ProcessLookupError, PermissionError, OSError, AttributeError):
        proc.kill()
    proc.wait(timeout=5)
    raise SubprocTimeout(...)
```

## Scope

Single file (`subproc.py`), 3-line change. No behavior change on macOS/Linux where `os.killpg` is always available.


365diascollaboration-prog:
Good catch — addressed in the latest commit. `proc.wait(timeout=5)` now runs inside a `try/except subprocess.TimeoutExpired` block: on second timeout it force-kills with `proc.kill()` and does a final blocking `proc.wait()` before raising `SubprocTimeout`. Callers will always see `SubprocTimeout`, never a raw `subprocess.TimeoutExpired`.

365diascollaboration-prog:
Fixed in latest commit — added `timeout=5` to the escalation `proc.wait()` and swallow the second `TimeoutExpired`. If the process is in D-state and unkillable, we leave it as a zombie rather than hanging the interpreter indefinitely.

PR #635, opened 2026-06-20
fix: add webapp environment detection to prevent false dependency errors on claude.ai

askusWT:
### Problem

When the skill is installed via claude.ai web (Settings > Customize > Skills), 
the model reads SKILL.md and encounters references to `scripts/last30days.py`, 
ScrapeCreators API keys, and Python 3.12+ runtime requirements. It incorrectly 
concludes the skill cannot run, warns the user about missing dependencies, and 
frames its web_search-based output as a degraded fallback.

The skill actually works correctly in the webapp — `web_search` is the native 
equivalent of the Python engine and requires no configuration. The failure is 
purely in the model's self-narration, not in the output.

### Fix

Adds `STEP 0.1: ENVIRONMENT DETECTION` immediately after the stale-clone check 
(Step 0), before the SKILL CONTRACT. This early placement ensures it is read 
before any dependency checks are triggered.

The block:
- Identifies the claude.ai webapp environment by absence of Bash tool + `/mnt/skills/` path
- Explicitly marks missing Python/API keys as non-errors in that context
- Redirects to `web_search` as the engine equivalent
- Documents the named failure mode in the same format as existing regressions in the file, for future debugging

### Tested
Reproduced the failure on claude.ai webapp (Sonnet 4.6, 2026-06-20). 
Applied fix to local SKILL.md copy and confirmed correct behaviour on re-run.

### DIFF
--- a/skills/last30days/SKILL.md
+++ b/skills/last30days/SKILL.md
@@ -90,6 +90,31 @@ If the SKILL.md path you just Read contains `/.claude/plugins/marketplaces/` AND
 
 ---
 
+# STEP 0.1: ENVIRONMENT DETECTION — claude.ai WEBAPP
+
+**Run this check BEFORE Step 1 and BEFORE flagging any missing dependencies:**
+
+Determine which environment you are running in:
+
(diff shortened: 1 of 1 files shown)


askusWT:
Closing — work incomplete, will reopen when ready.

PR #636, opened 2026-06-20
Salvage partial YouTube transcript on non-zero yt-dlp exit

SWAG489:
## Problem

On any run where the searched YouTube videos **do** have captions, `_fetch_transcript_ytdlp` could still return 0/N transcripts.

Root cause is the default `--sub-lang en,es,pt`. For an English video, yt-dlp writes `en.vtt` successfully, then `es`/`pt` hit `HTTP 429: Too Many Requests` and yt-dlp **exits non-zero** — even though the `en` track is already on disk. The code gated the VTT read on `result.returncode == 0`, so the `en` transcript was discarded. Worse, `429` matches `_TRANSIENT_RE`, so the fetch retried (2×), hammering the same rate limit, then returned `None`.

Reproduced locally with the exact engine command on a single video:

```
[info] Writing video subtitles to: .../Gqh_KdHP1Xk.en.vtt        # en succeeds
ERROR: Unable to download video subtitles for 'es': HTTP Error 429: Too Many Requests   # non-zero exit
```

`fetch_transcript` returned 0 chars despite `en.vtt` on disk. Setting `LAST30DAYS_YT_SUB_LANGS=en` worked around it (0 → 13,864 chars), confirming the diagnosis. With the default langs and parallel fetching, this showed up as 0/N transcripts across whole runs.

## Fix

On a non-zero exit, read any VTT already on disk **before** classifying the failure. A partial success (`en` written, `es`/`pt` 429'd) is a real transcript — return it instead of discarding it and retrying back into the rate limit.

- Parallel fetch verified **0/N → 3/3** locally after the change.
- Behavior unchanged when no VTT was written (still classifies / retries / sets `ytdlp_error`).
- Adds `test_partial_success_returns_vtt_despite_nonzero_exit`; full `test_youtube_yt.py` green (59 tests).

### DIFF
--- a/skills/last30days/scripts/lib/youtube_yt.py
+++ b/skills/last30days/scripts/lib/youtube_yt.py
@@ -668,6 +668,18 @@ def _fetch_transcript_ytdlp(
             # Genuine no-captions: return quietly (caller may still try direct).
             return None
 
+        # Non-zero exit, but yt-dlp may have written a usable VTT before the
+        # failing language errored. With the default `--sub-lang en,es,pt`, an
+        # English video fetches `en` fine, then `es`/`pt` hit a 429 and yt-dlp
+        # exits non-zero — yet the `en` track is already on disk. A partial
+        # success is still a real transcript, so salvage any VTT before
+        # classifying this as an error (and, worse, retrying straight back into
+        # the same rate limit). This is the root cause of the 0/N transcript
+        # runs reported when every video had captions.
+        partial_vtt = _read_vtt(video_id, temp_dir)
+        if partial_vtt is not None:
+            return partial_vtt
+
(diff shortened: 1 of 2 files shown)


SWAG489:
Good catch — addressed in 5ce339a. Added `self.assertEqual(run_mock.call_count, 1)` so the test now locks in the no-retry-before-salvage guarantee. Full `test_youtube_yt.py` stays green (59 tests).

mvanhorn:
Thanks @SWAG489, clean diagnosis. Pinpointing that the default `en,es,pt` langs write `en.vtt` then 429 on `es`/`pt`, and that gating the read on `returncode == 0` discarded the good track, was exactly the root cause. Merged, and it flips a lot of `0/N` runs on its own. Appreciate the regression test too.

PR #639, opened 2026-06-20
fix(hackernews): drop invalid points>2 numericFilter that 400'd every HN query

cescox:
## Problem

Every Hacker News search returns HTTP 400 and zero stories. The Algolia request builds:

```
numericFilters = created_at_i>{from},created_at_i<{to},points>2
```

But `points` is not in the HN index's `numericAttributesForFiltering` (only `created_at_i` is), so Algolia rejects the request:

```
{"code":400,"message":"invalid numeric attribute(points), attribute not specified in numericAttributesForFiltering setting"}
```

Observed in a real run: HN returned 0 across every subquery while all other sources worked. Looks like index bit-rot — `points` was filterable on an older index revision and isn't anymore.

## Fix

Drop the `points>2` clause from `numericFilters`. Low-engagement stories are already demoted by parse-time relevance scoring (`rank + engagement_boost` in `parse_hackernews_response`), so no server-side floor is needed. If a hard floor is still wanted, happy to add a client-side `MIN_POINTS` drop in a follow-up.

## Verification

- Live: `search_hackernews("claude code", <last 30 days>)` now returns **15 stories** (was 0, HTTP 400).
- `uv run pytest tests/test_hackernews.py` → **32 passed**.

## Test change

`test_search_hackernews_engagement_filter` asserted that `points` + `%3E2` **were present** in the URL — i.e. it pinned the exact invalid filter that broke every query (a bug-encoding test). Renamed to `test_search_hackernews_no_points_numericfilter` and inverted to assert the invalid clause is absent while the date filter remains, as a regression guard.

### DIFF
--- a/skills/last30days/scripts/lib/hackernews.py
+++ b/skills/last30days/scripts/lib/hackernews.py
@@ -93,13 +93,18 @@ def search_hackernews(
     core_flat = _flatten_query_for_algolia(core)
     _log(f"Searching for '{core_flat}' (raw: '{topic}', since {from_date}, count={count})")
 
-    # Use relevance-sorted search with minimum engagement filter.
+    # Use relevance-sorted search. The HN Algolia index only allows
+    # `created_at_i` in numericFilters; `points` is NOT in its
+    # `numericAttributesForFiltering`, so a `points>N` clause makes the API
+    # return HTTP 400 ("invalid numeric attribute(points)") and zero stories.
+    # Low-engagement stories are demoted by parse-time relevance scoring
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thanks @cescox, sharp diagnosis. Pinning the universal HN 400 to `points` being dropped from the index's `numericAttributesForFiltering` (with the error swallowed, so HN silently returned 0 on every run) was exactly right, and leaning on parse-time relevance for the low-engagement floor instead of a server-side filter is the correct call. Merged, shipping in 3.7.1.

PR #640, opened 2026-06-20
fix(polymarket): filter on stable topic; clean mangled market labels

cescox:
Two independent Polymarket bugs, both observed on a real run for topic "Claude Code skills" (which surfaced irrelevant markets like Claude-downtime / US-gov-ban and rendered a mangled footer fragment: `an Anthropic Claude model score at: an 19%, ... Claude go down 3-5 times in June: Claude 65%`).

## Fix A — relevance filter was fed the per-subquery string, not the topic

`_retrieve_stream` passed `topic=subquery.search_query` to `parse_polymarket_response`. `search_query` narrows differently on each fanout pass, so `_passes_topic_filter`'s informative-word set changed per subquery — dropping **everything** on narrow subqueries ("claude code subagents workflow") while letting off-topic Claude markets through on broad ones ("claude code"). 

Now keys off the stable original topic via `raw_topic or topic or subquery.search_query` — the same precedence the reddit/youtube/jobs branches in this function already use for `raw_topic`.

## Fix B — mangled labels from article-leading truncation

The fragment `an Anthropic Claude model score at: an 19%` was two independently-truncated strings concatenated, both keeping a leading article:
- `_shorten_polymarket_title` (render) 6-word fallback → `"an Anthropic Claude model score at"`
- `_shorten_question` (polymarket) 40-char fallback → `"an"`
- `_polymarket_top_markets` joined them as `f"{descriptor}: {lead_name} {pct}"`

Fixes:
1. Strip a leading article in both fallbacks (`^(?:a|an|the)\s+`).
2. Append the outcome label only when it adds information — suppress it when empty, a Yes/No proxy, a bare article, or already the descriptor's leading token.

Result: `an Anthropic Claude model score at: an 19%` → `Anthropic Claude model score at 19%`.

## Verification

- `uv run pytest tests/test_polymarket.py tests/test_polymarket_disambiguation.py tests/test_render_v3.py` → **121 passed** (8 new tests: article-stripping in both shorteners, redundant-outcome suppression incl. the exact `"an" + 19%` case, and the threshold-by-word-count contrast the wiring fix eliminates).
- Full suite otherwise green (unrelated pre-existing `test_check_config_ytdlp_detection` env failures only).

## Note (intentionally out of scope)

`_NOISE_WORDS` blocklists `skill`/`skills`, which discards the most discriminating word for skill-related topics. Left unchanged here since it affects filtering behavior more broadly — flagging for a separate decision.

### DIFF
--- a/skills/last30days/scripts/lib/pipeline.py
+++ b/skills/last30days/scripts/lib/pipeline.py
@@ -1238,7 +1238,12 @@ def _retrieve_stream(
         return truthsocial.parse_truthsocial_response(result), {}
     if source == "polymarket":
         result = polymarket.search_polymarket(subquery.search_query, from_date, to_date, depth=depth)
-        return polymarket.parse_polymarket_response(result, topic=subquery.search_query), {}
+        # Relevance filtering keys off the stable original research topic, not the
+        # per-subquery search_query (which narrows differently on each fanout pass
+        # and would let off-topic markets through on broad subqueries while dropping
+        # everything on narrow ones).
+        relevance_topic = raw_topic or topic or subquery.search_query
+        return polymarket.parse_polymarket_response(result, topic=relevance_topic), {}
     if source == "github":
         # Resolve once at the pipeline boundary so search and enrich
         # share the result; otherwise each call would re-run the env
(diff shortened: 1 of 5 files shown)


mvanhorn:
Thanks @cescox, two clean catches in one PR. Fix A, the relevance filter being fed the per-subquery string instead of the stable topic (so narrow subqueries dropped everything while broad ones leaked off-topic Claude markets), matched a garbled Polymarket footer I'd seen on real runs exactly. Fix B cleaned up the mid-article label truncation. Merged, shipping in 3.7.1. Small follow-up landed on top: the label fallback now truncates the cleaned string too.

PR #644, opened 2026-06-21
Fix last30days setup detection

natecl:
## Summary

Fixes setup and host-detection edge cases in the last30days skill:

- clarify that WebSearch loading is Claude-specific and Codex should fall back to the no-WebSearch engine path when the tool is unavailable
- require every expected cookie for a source before setup reports browser cookies as found
- make setup status text explicit that browser-cookie detection is best-effort and should be verified with `last30days.py --diagnose`
- avoid invoking `tr`/`sed` in the SessionStart hook when `EXCLUDE_SOURCES` is empty, which keeps the minimal-PATH hook test green

## Why

A local Codex invocation hit a stale first-run/setup path and then produced misleading coverage around X browser cookies. The installed skill could find a browser cookie hint, but source availability still depends on the runtime diagnostic path confirming the source is active.

## Validation

- `uv run pytest tests/test_setup_wizard.py`
- `HOME=$(mktemp -d /tmp/last30days-home.XXXXXX) uv run pytest tests/test_check_config_ytdlp_detection.py tests/test_setup_wizard.py`
- `HOME=$(mktemp -d /tmp/last30days-home.XXXXXX) uv run pytest`
  - `2026 passed, 4 skipped, 6 subtests passed`

### DIFF
--- a/hooks/scripts/check-config.sh
+++ b/hooks/scripts/check-config.sh
@@ -178,8 +178,12 @@ if [[ -n "$HAS_SCRAPECREATORS" ]]; then
   # whitespace) so the matching mirrors pipeline.py's .strip().lower() parsing.
   SC_ADD=3
   EXCLUDED="${ENV_EXCLUDE_SOURCES:-${EXCLUDE_SOURCES:-}}"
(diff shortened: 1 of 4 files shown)


natecl:
Closing per request; this was opened by mistake.

PR #659, opened 2026-06-22
fix: first-run setup wizard skipped on every new install

henkyermontero:
## Summary

- Every user's first `/last30days <topic>` silently skipped the setup wizard and ran degraded WebSearch-only synthesis — no browser cookie extraction, no yt-dlp, no Digg CLI.
- When Python 3.12+ is absent, the engine hard-exits but the model rationalized past the error and fell back to WebSearch-only mode with no user-visible warning.
- Reproduced on 2026-06-22 (Fredy Montero, Claude Sonnet 4.6, fresh macOS with Python 3.9.6).

## Root cause

The branching rule in **HOW TO INVOKE** (~line 241 of `SKILL.md`) read:

> "proceed to Step 0.5 / Step 0.55 / Step 0.75 / Research Execution below"

It said nothing about the wizard. The model read this, jumped to Step 0.5, and never encountered `## Step 0: First-Run Setup Wizard` at line ~339. The wizard code in `scripts/lib/setup_wizard.py` is correct — it was simply never called.

### `skills/last30days/SKILL.md`

1. **FIRST-RUN GATE** — inserted a mandatory shell command between STEP 0 (WebSearch) and the branching rule:
   ```bash
   cat ~/.config/last30days/.env 2>/dev/null | grep -c "SETUP_COMPLETE=true" || echo "FIRST_RUN_DETECTED"
   ```
   Observable output (`1` vs `FIRST_RUN_DETECTED`) cannot be rationalized past. Forces the model to check before any research begins.

2. **Branching rule update** — changed from routing directly to Step 0.5 to explicitly routing through `## Step 0: First-Run Setup Wizard` first, then continuing to Step 0.45/0.5/0.55/0.75.

3. **PYTHON VERSION GATE** — added a LAW-class block in the Runtime Preflight section: if `scripts/last30days.py` exits with a Python version error, the model must display a platform-specific install command (`brew install python@3.12` / `winget install Python.Python.3.12`) and **stop** — no WebSearch-only fallback.

4. **Named failure mode (2026-06-22)** — documented the root cause and the Fredy Montero reproduction for future debugging.

### `skills/last30days/scripts/last30days.py`

Improved `ensure_supported_python()` error message with platform-specific install commands:
```
last30days v3 requires Python 3.12+.
Detected Python 3.9.6.
Install with:
  Mac:     brew install python@3.12
  Windows: winget install Python.Python.3.12
  Linux:   sudo apt install python3.12  (or pyenv install 3.12)
Then rerun: python3.12 <path-to-script> setup
```

Existing test `test_ensure_supported_python_rejects_old_interpreter_with_actionable_error` passes — it checks for "last30days v3 requires Python 3.12+", "Detected Python 3.9.6", and "python3.12", all still present.

## What was NOT changed

- `scripts/lib/setup_wizard.py` — wizard code was already correct (yt-dlp via `brew install yt-dlp`, Digg CLI via `npx @mvanhorn/printing-press-library@0.1.16 install digg --cli-only`). It was just never called.
- `MIN_PYTHON` version lower (deferred) — code audit found no Python 3.12-specific runtime features in use (no `tomllib`, `asyncio.TaskGroup`, `ExceptionGroup`, `Path.walk()`, or `type` statements; all union annotations covered by `from __future__ import annotations`). Actual minimum is likely 3.10. However, `uv` is not available on the test machine to run `uv run --python 3.10 pytest`. Deferring this change until the test suite can be validated against 3.10.

## Test plan

- [ ] Contract tests in `tests/test_runtime_preflight_contract.py` pass (3/3 verified locally on Python 3.9 — pure text tests that read SKILL.md)
- [ ] `test_ensure_supported_python_rejects_old_interpreter_with_actionable_error` in `tests/test_cli_v3.py` — verified manually against new message shape; requires Python 3.12+ to run
- [ ] On a fresh machine with no `~/.config/last30days/.env`: invoke `/last30days <topic>` — model should run the first-run gate command, detect `FIRST_RUN_DETECTED`, and jump to the setup wizard before any research
- [ ] On a machine with `SETUP_COMPLETE=true` in `.env`: invoke `/last30days <topic>` — gate outputs `1`, model skips wizard and proceeds to research normally
- [ ] On a machine with Python 3.9 only: model should display the `brew install python@3.12` command and stop, not fall back to WebSearch-only synthesis

### DIFF
--- a/skills/last30days/SKILL.md
+++ b/skills/last30days/SKILL.md
@@ -230,7 +230,18 @@ ToolSearch select:WebSearch
 
 WebSearch is a **deferred tool** in Claude Code v2.1.114. The frontmatter of this file authorizes it (`allowed-tools: ... WebSearch`) but the runtime lists it as "schemas are NOT loaded." Calling WebSearch without `ToolSearch select:WebSearch` first will fail or do nothing. That friction is the documented cause of the second-most-common failure mode of this skill: the model sees "WebSearch is there but deferred," takes the low-friction path, skips Step 0.5 and 0.55, and runs the engine bare with only keyword search. The output looks fine but misses founder X timelines, GitHub repo activity, and subreddit-specific threads.
 
-Load WebSearch first. No exceptions. Then proceed to the branching rule below.
+Load WebSearch first. No exceptions. Then run the first-run gate below before anything else.
+
+**FIRST-RUN GATE — run this Bash command immediately after loading WebSearch, before reading the topic or doing any research:**
+
+```bash
+cat ~/.config/last30days/.env 2>/dev/null | grep -c "SETUP_COMPLETE=true" || echo "FIRST_RUN_DETECTED"
+```
+
+- Output is `1` → setup is complete. Continue to the branching rule below.
+- Output is `FIRST_RUN_DETECTED` (or `0`) → this is a first run. Jump immediately to `## Step 0: First-Run Setup Wizard` and complete it **before doing any topic research**. Do NOT proceed to Step 0.5, do NOT load WebSearch supplements, do NOT synthesize anything. The wizard installs yt-dlp (YouTube), the Digg CLI (via `npx`), and extracts browser cookies for X/Twitter and other sources. Skipping it produces a degraded WebSearch-only result that misrepresents the skill's capability to the user.
+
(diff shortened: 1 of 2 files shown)


mvanhorn:
Fredy, thank you for this - genuinely grateful for the help. You nailed the root cause: the branching rule in HOW TO INVOKE routed straight to Step 0.5 and orphaned the Step 0 wizard, so every brand-new install silently skipped onboarding and fell through to degraded WebSearch-only synthesis with no X/YouTube/TikTok. The observable first-run gate is exactly the right fix, and the Python-version gate is a great addition.

Merging as-is so your authorship lands on the commit. I'm following up with a tiny commit for the Greptile P1 (`grep -c || echo` emits `0` and `FIRST_RUN_DETECTED` on two lines for a fresh install) so the gate emits one token, plus the Linux install line - all crediting your work here.

PR #661, opened 2026-06-23
feat(setup): restore the v3.0.0 first-run NUX wizard

mvanhorn:
## What

Restores the original v3.0.0 first-run NUX wizard - the guided, modal-driven Claude Code onboarding that eroded out of `SKILL.md` over time - layered on top of the consent-driven foundation from #659 and #660.

`Step 0` now branches by host:

- **Claude Code Modal Flow** (hosts with `AskUserQuestion`): welcome message → Auto/Manual/Skip setup modal → cookie-consent modal → ScrapeCreators signup offer → TikTok/Instagram `INCLUDE_SOURCES` opt-in → first-topic picker.
- **Non-Modal Prose Flow** (OpenClaw, Codex, Cursor, Gemini CLI): the same work, done conversationally.

## Why

The original wizard was a guided flow Matt built for v3.0.0. It eroded into a flat prose sequence, and the routing bug #659 fixed had orphaned it entirely on new installs. This brings the polished flow back while keeping everything #659/#660 added.

## Notes

- **Additive, not a revert.** Everything from #659 (first-run gate) and #660 (consent-driven prose, automatic key-persist, macOS Full Disk Access remediation) is preserved. The source inventory is current, not the v3.0.0 set.
- **Digg** is threaded into the install messaging everywhere yt-dlp/YouTube appears (the install already existed; this is the copy). The ScrapeCreators-backed YouTube transcript fallback, Reddit backup, and YouTube comments are now named in the onboarding copy.
- ScrapeCreators credit count restored to **10,000 free calls**.
- The hard `ALWAYS execute Step 0 BEFORE Step 1` gate is restored to resist re-erosion.
- **Threads/Pinterest** are intentionally not surfaced in onboarding (power-user `INCLUDE_SOURCES` only).
- `tests/test_onboarding_contract.py` rewritten for the two-branch structure and to lock the flow against silent re-erosion. Full suite green.
- Original wizard captured at `docs/reference/old-nux-wizard-v3.0.0.md`.


mvanhorn:
Addressed all four Greptile findings in 4208b79:
- **Skip+Skip loop (real bug):** picking "Skip for now" at the setup choice now writes `SETUP_COMPLETE=true` and proceeds to the topic picker, so the first-run gate doesn't re-fire every run.
- **Step labels:** the body now has explicit Step 1/2/3 matching the sequence descriptor (it jumped 1 → 4).
- **Prose `BROWSER_CONSENT` bypass:** the Non-Modal Prose Flow now honors an existing `BROWSER_CONSENT=true` instead of re-asking.
- **Test symmetry:** added the prose Full Disk Access assertion plus a guard that the Skip path writes `SETUP_COMPLETE`.

mvanhorn:
Restored the masked-key retrieval hint on the `persisted: false` path in both flows (be130f1). All Greptile findings resolved; tests green. Merging.

PR #664, opened 2026-06-23
Honor plan's explicit sources at quick depth

Bexr66:
## Problem

With a valid `SCRAPECREATORS_API_KEY` and an explicit `--plan` listing `tiktok`/`instagram`/`reddit` in a subquery's `sources`, those sources returned 0 items at `--quick`. The quick-depth trimmer substituted global priority defaults (`jobs`, `youtube`) for the plan's explicit sources.

Root cause: the quick-depth branch built `preferred_sources = ranked_sources[:limit]` from the global priority table and ignored `subquery.sources`. The `requested_sources` rescue only fires for the `--search` flag, not for plan sources. So an external quick plan asking for `tiktok` could end up with `['jobs', 'youtube']`.

## Fix

The quick-depth trimmer now selects from the plan's own sources: priority-ranked plan sources first, then plan sources absent from the priority table like `instagram`, then applies the quick-depth limit to that planner-selected set. It falls back to the priority top-N only when the subquery lists no usable source.

The follow-up cleanup also removes unreachable default-depth/dead-else code from this block. Explicit `--search` sources are prioritized as user overrides within the quick-depth cap; they no longer append past the cap.

## Verification

- `_sanitize_plan(..., "quick")` on a plan with `[reddit, tiktok, instagram, youtube]` now keeps plan-selected sources instead of returning `[jobs, youtube]`.
- Requested sources are prioritized inside quick slots and remain capped by the quick-depth limit.
- `uv run pytest tests/test_planner_v3.py tests/test_internals_v3.py -k 'planner or TrimSubqueries or quick_mode or default'`
- `python3 -m py_compile skills/last30days/scripts/lib/planner.py`
- `git diff --check`

https://claude.ai/code/session_013b95ARRnJMYQjmJbUKQN75

### DIFF
--- a/skills/last30days/scripts/lib/planner.py
+++ b/skills/last30days/scripts/lib/planner.py
@@ -345,7 +345,7 @@ def _trim_subqueries_for_depth(
     limits = SOURCE_LIMITS.get(depth)
     if not limits:
         return subqueries
-    priority_table = QUICK_SOURCE_PRIORITY if depth == "quick" else SOURCE_PRIORITY
+    priority_table = QUICK_SOURCE_PRIORITY
     priority = priority_table.get(intent, priority_table["breaking_news"])
     limit = limits.get(intent, 3)
     ranked_sources = [source for source in priority if source in available_sources]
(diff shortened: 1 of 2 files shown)


tmchow:
> The comment inaccuracies around the requested_sources cap and the dead else-branch are worth a second look before merge.

Addressed in 3641369:
- Removed the unreachable default-depth/dead `else` branch from the quick-only trimming block.
- Collapsed the duplicate explanation into one comment.
- Removed the redundant `source in available_sources` guard from the plan-source loop.
- Removed the misleading “still capped” wording. Explicit `--search` sources remain user overrides when they are present in the plan, matching the existing quick-mode requested-source test.
- Updated the PR title/body so it no longer claims default-depth source trimming changed.

Validation: planner-focused tests passed, `planner.py` compiles, and `git diff --check` passed.

tmchow:
> `requested_sources` injection can exceed the quick-mode source cap

Fixed in 4dbaaf6. Quick mode now seeds requested sources first, fills any remaining quick slots from the plan-ranked source list, and stops once the quick-depth limit is reached. I also updated the regression test to assert an explicitly requested source wins a slot while the final source list remains capped.

Validation: planner-focused tests passed again, `planner.py` compiles, and `git diff --check` passed.

PR #671, opened 2026-06-24
fix: trust project config explicitly

tmchow:
## Summary

Folder-mode runs no longer inherit hidden `.claude/last30days.env` files from arbitrary directories. Project config now loads only when trust is granted from process env, global config, or explicit engine policy, and discovery stops at the git root so unrelated parent folders cannot redirect credentials.

Safe diagnose can still explain what happened: it reports ignored project config key names and blocked endpoint overrides without exposing secret values.

## Validation

- `uv run pytest tests/test_project_config_trust.py tests/test_security_boundaries.py`

---

Codex

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -69,14 +69,14 @@ Re-run onboarding by deleting `~/.config/last30days/.env`. The mechanical work l
 
 ## API keys (`.env`)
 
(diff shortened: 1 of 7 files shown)


tmchow:
Addressed the top-level review-summary item about config_exists() in a3bdade. It now uses the same project-config trust decision as get_config(), and tests cover untrusted project-only config, process-trusted project config, and global config.

PR #699, opened 2026-06-26
fix(preflight): find uv-managed Python off PATH; use resolved interpreter for setup

### LINKED ISSUE #695: Runtime Preflight fails for uv-managed Python 3.12+ off PATH; setup commands hardcode python3
**Environment:** macOS (Homebrew broken / unavailable), Python 3.12+ installed only as a uv-managed interpreter (`uv python install`), not symlinked onto `PATH` under its version name. System `python3` is 3.9.

## Bug 1 — Preflight can't find a valid uv-managed Python

The Runtime Preflight probes `python3.14 python3.13 python3.12 python3 python` on `PATH` only. When the sole 3.12+ interpreter is uv-managed and off-PATH, resolution fails and the skill aborts with "requires Python 3.12+", even though a valid Python is installed and `uv` could find it.

**Suggested fix:** after the PATH loop, fall back to uv:

```bash
if [ -z "${LAST30DAYS_PYTHON:-}" ] && command -v uv >/dev/null 2>&1; then
  uv_python="$(uv python find '>=3.12' 2>/dev/null)" && try_last30days_python "$uv_python"
fi
```

## Bug 2 — Setup commands bypass the resolved interpreter

The "Runtime Preflight" tells the agent to resolve `LAST30DAYS_PYTHON`, but the setup commands in both the Claude Code Modal Flow and the Non-Modal Prose Flow hardcode `python3 skills/last30days/scripts/last30days.py setup ...`. On a machine where `python3` is < 3.12, setup runs on the wrong interpreter while research execution (which correctly uses `${LAST30DAYS_PYTHON:-python3}`) does not. Inconsistent.

**Suggested fix:** use `"${LAST30DAYS_PYTHON:-python3}"` in the 7 setup invocations, matching the research-execution calls.

## Note / deeper issue

`LAST30DAYS_PYTHON` is never `export`ed, and a fresh shell per invocation means the variable doesn't survive across separate shell calls — `${LAST30DAYS_PYTHON:-python3}` then silently falls back to a possibly-broken `python3`. A robust fix would resolve the interpreter in the same shell block as each engine call (or persist the resolution).

SeanGearin:
## Problem

The Runtime Preflight in `skills/last30days/SKILL.md` resolves a Python 3.12+ interpreter into `LAST30DAYS_PYTHON`, but only probes `python3.14 … python3` on `PATH`. On a machine where the only 3.12+ interpreter is **uv-managed and not symlinked onto `PATH`** (Homebrew unavailable, system `python3` is 3.9), resolution fails and the skill aborts with `requires Python 3.12+` even though a valid interpreter is installed.

Separately, the `setup` invocations hardcode `python3 …/last30days.py setup`, while the research-execution calls use `${LAST30DAYS_PYTHON:-python3}`. On a box where the system `python3` is < 3.12, setup runs on the wrong interpreter while research runs on the right one.

## Fix

1. **uv fallback** — after the `PATH` loop and before the version-error exit, fall back to `uv` when it's available:
   ```bash
   if [ -z "${LAST30DAYS_PYTHON:-}" ] && command -v uv >/dev/null 2>&1; then
     uv_python="$(uv python find '>=3.12' 2>/dev/null)" && try_last30days_python "$uv_python"
   fi
   ```
   `try_last30days_python` still validates `>= 3.12`, and an empty result is a safe no-op.

2. **Consistent interpreter for setup** — the `setup` invocations now use `${LAST30DAYS_PYTHON:-python3}`, matching the research-execution calls.

## Test plan
- `tests/test_runtime_preflight_contract.py` passes (3/3).
- Confirmed no remaining hardcoded `python3 …/last30days.py setup` invocations; research-execution calls untouched.

## Out of scope (noted in the issue)
The issue also notes that `LAST30DAYS_PYTHON` is never `export`ed, so a fresh shell per invocation can leave it unset. That affects the research-execution calls equally and is orthogonal to these two fixes, so I've left it for a separate follow-up rather than widen this PR.

Fixes #695.

---
*Co-drafted with my AI coding agent and reviewed by me before submission.*


SeanGearin:
Good catch — fixed in 3166413. Quoted the interpreter expansion (`"${LAST30DAYS_PYTHON:-python3}"`) across all the setup invocations, plus the pre-existing `--preflight` call on the same flow, so every interpreter call in SKILL.md is space-safe and consistent. Contract test still green.

SeanGearin:
Closing this — `main`'s uv fallback already covers it, and better: it provisions a managed CPython 3.12 when none is found rather than only locating an existing one. Thanks for tightening that up!

PR #700, opened 2026-06-26
fix: --diagnose no longer falsely reports X unreachable when browser cookies are configured (#692)

### LINKED ISSUE #692: Bug: --diagnose uses plan_only cookie mode, causing X to appear unavailable even when Safari cookies are valid
## Summary

When the skill starts up via the `SessionStart` hook, it runs `--diagnose` to check source availability. The `--diagnose` flag intentionally uses `plan_only` browser-cookie mode (a privacy safeguard — it never reads actual cookie values). This causes `available_sources` to exclude `x` and `bird_authenticated` to report `false`, even when the user's Safari cookies are completely valid and X works fine on actual research runs.

The skill's own startup hook then surfaces this state as "X unreachable" to the user, which is a false negative.

## Steps to reproduce

1. Have `FROM_BROWSER=auto` set in `~/.config/last30days/.env` and be logged into x.com in Safari.
2. Start a new Claude Code session. The `SessionStart` hook runs `--diagnose`.
3. Observe that `available_sources` does **not** include `x`, and `bird_authenticated` is `false`.
4. Run an actual research query — X returns posts successfully (`source: browser`, `x_backend: bird`).

## Root cause

In `scripts/last30days.py`, `_config_policy_for_args` sets `browser_mode = "plan_only"` when `args.diagnose` is true:

```python
elif args.diagnose or args.preflight:
    browser_mode = "plan_only"
```

In `plan_only` mode, `extract_browser_credentials` is never called, so `AUTH_TOKEN`/`CT0` are never populated in the config. Consequently `x_backend_chain` finds no credentials and returns an empty list, causing X to be absent from `available_sources`.

Meanwhile, for actual research runs, `browser_mode = "read"` causes Safari cookies to be read dynamically at runtime — and X searches succeed.

## Confirmed working

On the affected machine:
- `extract_safari_cookies("x.com", ["auth_token", "ct0"])` returns valid tokens
- In `read` mode: `AUTH_TOKEN present: True (source: browser)`, `x_backend: bird`
- A test research query returns `X: 13 posts`

## Suggested fix

The `--diagnose` output (and the startup hook that consumes it) should distinguish between "X is unavailable because no credentials exist" and "X would be available via browser cookies, but we didn't read them in this diagnostic mode."

Options:

1. **Add a `browser_cookies_configured` field** to the diagnose JSON that is `true` when `FROM_BROWSER` is set to a non-`off` value — even in `plan_only` mode. The hook/skill can use this to infer that X is likely available without actually reading cookie values.

2. **Include `x` in `available_sources` when `FROM_BROWSER` is set** and the browser list is non-empty, even in `plan_only` mode — since the intent is clearly that X should be available.

3. **Document the limitation** in the diagnose output: add a note like `"x_note": "not checked in diagnose mode (plan_only); will be available at runtime via FROM_BROWSER"`.

Option 1 or 2 would fix the startup hook's false-negative display without compromising the cookie-read privacy invariant of `--diagnose`.

## Environment

- macOS, Safari, `FROM_BROWSER=auto`
- Plugin version: 3.8.1
- Claude Code with `SessionStart` hook running `--diagnose`

23241a6749:
**Problem**

`--diagnose` uses `plan_only` browser-cookie mode as a privacy safeguard — it never reads actual cookie values. This means `AUTH_TOKEN`/`CT0` are never populated in the config, causing `x_backend_chain()` to find no credentials and report X as unavailable, even though X works fine on actual research runs.

**Solution**

In `x_backend_chain()`, when in `plan_only` mode and `FROM_BROWSER` is set to Firefox or Safari (browsers that can provide X cookies), treat bird as potentially available. The runtime mode will re-check with live cookie values.

Closes #692.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,11 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-<<<<<<< HEAD
 - The query-plan invocation guidance now warns against wrapping the heredoc in `bash -lc '...'` / `zsh -lc '...'`, whose single quotes terminate at the first apostrophe in a ranking string and abort the engine run with `unmatched "` on Codex. The quoted `<<'PLAN_EOF'` heredoc is already apostrophe-safe; the `-lc` wrapper was the hazard.
-=======
 - `--diagnose` / `--preflight` no longer falsely reports X as unreachable when browser cookies are configured but cookie reading is skipped for privacy (`plan_only` mode). The X backend chain now treats bird as potentially available when `FROM_BROWSER` points to Firefox or Safari ([#692](https://github.com/mvanhorn/last30days-skill/issues/692))
->>>>>>> bd5591c (fix: --diagnose no longer falsely reports X unreachable when browser cookies are configured (#692))
 - Firefox profile detection on Linux now checks `$XDG_CONFIG_HOME/mozilla/firefox` (or its default `~/.config/mozilla/firefox`) in addition to `~/.mozilla/firefox`, fixing cookie extraction on distros that honour the XDG Base Directory Specification ([#667](https://github.com/mvanhorn/last30days-skill/issues/667))
 
 ## [3.8.1] - 2026-06-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


mvanhorn:
Thank you for this, @23241a6749 — you spotted and fixed #692 first, and your diagnosis was spot on. 🙏

I ended up merging #729 for the same issue, which builds on your fix: it also covers the Chrome case (`FROM_BROWSER=chrome`, not just Firefox/Safari), adds an explicit `x_pending_browser_auth` flag to the `--diagnose` output, and keeps `get_x_source` honest (it reports X as *pending* rather than spoofing a `bird` backend with no creds). It ships with a 14-case test suite.

You're credited as co-author on the merged CHANGELOG entry and commit, since you found and fixed this first. Closing this as superseded by #729 — but genuinely appreciate the catch and the clean writeup. Please keep them coming!

23241a6749:
Thanks so much! I really appreciate the acknowledgment and the co-author credit. It was a fun issue to investigate, and I'm glad the final implementation is even more robust. Looking forward to contributing more improvements!

PR #702, opened 2026-06-26
Add LinkedIn source via ScrapeCreators

ravstr:
Adds LinkedIn as a retrieval source, using the ScrapeCreators
`/v1/linkedin/search/posts` endpoint (Google-indexed LinkedIn content,
no auth required on ScrapeCreators' side beyond the API key).

Changes:
- New `lib/linkedin.py`: `search_linkedin()` calls the ScrapeCreators
  endpoint; `parse_linkedin_response()` normalizes the raw response into
  engine-compatible item dicts (id, text, url, author, date, engagement).
- `lib/pipeline.py`: registers `linkedin` as an available source (gated
  on `SCRAPECREATORS_API_KEY`, same as tiktok/instagram), adds it to
  `MOCK_AVAILABLE_SOURCES` and `MAX_SOURCE_FETCHES` (capped at 1
  fetch/run), and dispatches to it in `_retrieve_stream`.
- `lib/normalize.py`: adds `_normalize_linkedin` to map raw LinkedIn
  post dicts into the generic `SourceItem` model, registered in the
  source normalizer dispatch dict.
- `tests/test_linkedin.py`: unit tests for response parsing, field
  fallbacks, and the no-token/error short-circuit paths.

Requires `SCRAPECREATORS_API_KEY` to be set; falls back to skipping
(empty results) if absent, consistent with other ScrapeCreators-backed
sources.

Note on `DEPTH_CONFIG`: since `SCRAPECREATORS_API_KEY` is BYOK, the
default/deep windows use `last-month` with higher `max_results` (20/30)
rather than throttling to `last-week`/lower counts — each user's own
credit usage is their call to manage, not something to bake into the
upstream default.

Testing: added unit tests in `tests/test_linkedin.py` covering response
parsing, field-name fallbacks, and the no-token/error short-circuit
paths (all mocked, no live API calls). Ran the full existing test suite
locally with no regressions.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -99,6 +99,7 @@ The project-scoped file is useful for **intentional per-client setups**: drop a 
 | Instagram | `SCRAPECREATORS_API_KEY` + `INCLUDE_SOURCES` contains `instagram` | Instagram Reels | 10K free calls; raise `LAST30DAYS_TRANSCRIPT_TIMEOUT` (default 30s) if SC is slow on your network |
 | Threads | `SCRAPECREATORS_API_KEY` + `INCLUDE_SOURCES` contains `threads` | Threads items | 10K free calls |
 | Pinterest | `SCRAPECREATORS_API_KEY` + `INCLUDE_SOURCES` contains `pinterest` | Pinterest items | 10K free calls |
+| LinkedIn | `SCRAPECREATORS_API_KEY` + `INCLUDE_SOURCES` contains `linkedin` | LinkedIn posts | 10K free calls; power-user opt-in, not offered during first-run onboarding |
 | Bluesky | `BSKY_HANDLE` + `BSKY_APP_PASSWORD` | Bluesky items | yes (app password at bsky.app) |
 | TruthSocial | `TRUTHSOCIAL_TOKEN` | TruthSocial items | yes |
 | Web search | one of: `BRAVE_API_KEY`, `EXA_API_KEY`, `SERPER_API_KEY`, `PARALLEL_API_KEY` | `--auto-resolve` and Step 2 supplements | Brave has a free tier; native WebSearch on Claude Code / Codex / Gemini works as a fallback |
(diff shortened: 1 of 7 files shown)


ravstr:
Thanks for the review, i've addressed all three points below:

1. INCLUDE_SOURCES gate. Yes this needed an explicit opt-in. I checked the existing precedent in `available_sources()`: TikTok/Instagram activate on `SCRAPECREATORS_API_KEY` alone (they're offered during SKILL.md's Step 0 onboarding flow), while Threads/Pinterest/Perplexity are power-user-only and require explicit
`INCLUDE_SOURCES` opt-in since they're not surfaced during onboarding. LinkedIn fits the latter bucket, so I gated it the same way Perplexity is gated - `INCLUDE_SOURCES=linkedin` or `--sources linkedin` now required; existing `SCRAPECREATORS_API_KEY` holders won't pick it up silently. Added `TestLinkedinAvailability` covering this.

2. Date-range filtering. There is a global date filter applied to every source in `normalize.py`, so results weren't entirely unfiltered, but you're right that LinkedIn was missing the source-level filter Instagram has (with its graceful "keep all if filtering would zero out the results" fallback for when SC doesn't return a parseable date). Added the same pattern to `parse_linkedin_response()`, plus `TestDateRangeFiltering` coverage.

3. Docs. Added LinkedIn to CONFIGURATION.md's source table, SKILL.md's tags list, the power-user opt-in note (alongside Threads/Pinterest), and the `ACTIVE_SOURCES_LIST` display-name mapping.

Also dropped the unused `sys` import you flagged in linkedin.py.

Full test suite passes locally (110/110 on the new/touched tests, no regressions elsewhere).

mvanhorn:
Thanks for this, @ravstr — great addition. I pushed one fix commit on top (maintainer edit) before merging.

Heads up on what it changes and why:

- **Parser field mapping (the blocker).** The live `/v1/linkedin/search/posts` post object carries the body in `description` and the timestamp in `datePublished`. The parser was keyed on `text`/`date`, so every post hit the `if not text: continue` guard and got dropped — verified against the real API, 10 raw posts → 0 items. Added the real field names to the existing fallback chains and realigned the test fixtures to the actual response shape (with a regression guard for exactly this case, since the old mock shape kept the tests green while the real path was dead).

- **Articles as high signal.** Articles never come back from post search (every result is a `/posts/` URL); LinkedIn long-form lives only on the profile endpoint's `articles[]`. Added a bounded enrichment lane: on person topics (a returned post author matches the topic) we make one `/v1/linkedin/profile` call and surface that author's articles as high-signal items (relevance 0.9, ranked above posts). Keyword topics make no extra call.

Your opt-in gating (`INCLUDE_SOURCES=linkedin`, off by default) is unchanged and now pinned by a regression test. Full suite green (2299 passed). Merging shortly — thanks again!

PR #708, opened 2026-06-27
fix: register FUN_LEVEL in env.py so .env config is not silently ignored

### LINKED ISSUE #707: Bug: FUN_LEVEL .env config silently ignored (not registered in env.py keys)
## Summary

`FUN_LEVEL` (controls how many "best takes" / funny comments appear in output, values: `low|medium|high`) is missing from the `keys` tuple in `scripts/lib/env.py`. The `get_config()` function only populates the config dict with keys from this tuple, so any value set in the `.env` file is silently omitted.

## Impact

Users who set `FUN_LEVEL=high` in `~/.config/last30days/.env` always silently get `"medium"` behavior instead. The `--fun-level` CLI flag and the process environment variable `FUN_LEVEL` are unaffected (those bypass `env.py`), so only `.env` configuration is broken.

## Root cause

`scripts/lib/env.py` line ~460: the `keys` tuple list does not include `('FUN_LEVEL', 'medium')`.

## Fix

Add `('FUN_LEVEL', 'medium')` to the `keys` list in `env.py`.

## Environment

23241a6749:
Fixes #707.

`FUN_LEVEL` was missing from the config keys tuple in `env.py`, so `get_config()` would never load it from the `.env` file. Users setting `FUN_LEVEL=high` in `.env` always silently got `"medium"` behavior.

The `--fun-level` CLI flag and process env var `FUN_LEVEL` were unaffected (both bypass `env.py`).

### DIFF
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -502,6 +502,7 @@ def get_config(policy: ConfigLoadPolicy | None = None) -> dict[str, Any]:
         ('INCLUDE_SOURCES', ''),
         ('EXCLUDE_SOURCES', ''),
         ('LAST30DAYS_DEFAULT_SEARCH', ''),
+        ('FUN_LEVEL', 'medium'),
         ('LAST30DAYS_YOUTUBE_SSH_HOST', None),
         ('LAST30DAYS_TRANSCRIPT_TIMEOUT', None),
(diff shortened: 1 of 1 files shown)


mvanhorn:
Rebased and dropped the CHANGELOG hunk, then merged. Your registration line was exactly right, including the medium default matching the consumer.

PR #709, opened 2026-06-27
feat: Add arXiv + Techmeme (default-on) and Trustpilot (opt-in) sources

mvanhorn:
Adds three Printing Press CLIs as last30days sources, built on the Digg adapter pattern. arXiv and Techmeme are default-on (zero-auth, auto-installed); Trustpilot is opt-in.

## Sources

- **arXiv (📄)** — research papers. Default-on, zero-auth. Quoted-phrase + relevance-sorted query with a 365-day recency cutoff, which also keeps it quiet on non-research topics (a stale keyword match drops out).
- **Techmeme (📰)** — tech-news headlines. Default-on, zero-auth. Syncs a local cache once per run (lock + TTL guarded), then searches; filters publication-name header rows.
- **Trustpilot (⭐)** — brand/company review sentiment. **Opt-in** via `INCLUDE_SOURCES=trustpilot`; off by default because it can do a one-time headless-Chrome WAF-cookie harvest. Gated to company/brand topics, browser opt-out via `LAST30DAYS_TRUSTPILOT_NO_BROWSER`, degrades to empty on any failure.

## Default-on mechanism

Mirrors Digg/yt-dlp: the setup wizard auto-installs `arxiv-pp-cli` and `techmeme-pp-cli` via `npx @mvanhorn/printing-press-library install <slug> --cli-only`, and `available_sources()` activates each when its binary is on PATH. Trustpilot is **not** auto-installed and is gated behind `INCLUDE_SOURCES` (same pattern as perplexity/linkedin), so absent an explicit install + opt-in it never runs and never touches a browser.

## Integration

Each source follows the Digg pattern across `pipeline.py` (gate, dispatch, mock fixture), `normalize.py`, `planner.py`, `signals.py`, `render.py` (emoji), `ui.py`, plus `setup_wizard.py` for installs. Emoji (📄/📰/⭐) checked against the existing set for collisions.

## Validation

| Check | Result |
|---|---|
| Full test suite | PASS |
| Coverage | 83.76% (floor 60%; new modules 62-75%) |
| Live eval (real CLIs) | arXiv/Techmeme return on-topic results; controls stay quiet |
| Code review | 5-reviewer pass; P1/P2 findings fixed |

~70 new tests across the three adapters, the wizard installer, and a diverse-topic eval harness (deterministic fire/quiet matrix + opt-in `LAST30DAYS_EVAL_LIVE` mode).

## Notes

- The Techmeme `--agent` path depends on printing-press-library PR #1383 (the `compactFields` fix); this adapter uses `search --json` so it works on any binary version regardless.
- Accepted residual (follow-up): adapter-level URL-scheme guard parity with Digg — the actual XSS surface (HTML output) is already centrally guarded by `html_render._is_safe_link_url`, and these sources emit URLs from trusted first-party CLIs scraping known domains.


mvanhorn:
Addressed the Greptile P1 in f2a925e: the brand-shape gate now excludes single-word programming languages, frameworks, runtimes, OSes, and dev tools via a `_TECH_TOKENS` set, so a bare `Python`/`React`/`Docker`/`Rust`/`Linux` query no longer passes as a brand (no Chrome harvest, no wrong-company data). Explicit company intent still works through the domain branch (e.g. `docker.com`). Added parametrized tests covering the flagged tech names plus the domain escape hatch.

PR #715, opened 2026-06-29
fix: open .env with explicit utf-8 encoding for Windows compat

### LINKED ISSUE #714: Bug: load_env_file() opens .env without utf-8 encoding (Windows UnicodeDecodeError)
## Summary

`load_env_file()` at `scripts/lib/env.py:136` opens the `.env` config file without specifying `encoding='utf-8'`. On Windows, Python's default encoding is the system locale (e.g. `cp1252`), not UTF-8.

The `setup_wizard.py` writes `.env` files using `encoding="utf-8"`, creating a read/write encoding mismatch. If a `.env` file contains any non-ASCII character (accented characters, emoji, CJK, or API tokens with special bytes), `load_env_file()` raises `UnicodeDecodeError` on Windows with non-UTF-8 system locale.

## Impact

Medium. Windows users with non-ASCII values in their `.env` file (e.g., a display name, a non-English path, or emoji in comments) will see the app fail to start with `UnicodeDecodeError`. Users with purely ASCII `.env` files are unaffected.

## Suggested fix

```diff
- with open(path, 'r') as f:
+ with open(path, 'r', encoding='utf-8') as f:
```

## Environment

- Windows 10/11 with non-UTF-8 system locale

23241a6749:
Fixes #714.

`load_env_file()` at `env.py:136` opened `.env` without specifying `encoding='utf-8'`. On Windows with non-UTF-8 system locale, any non-ASCII character in the file would raise `UnicodeDecodeError`. The `setup_wizard.py` writes `.env` with `encoding='utf-8'`, creating a read/write mismatch.

### DIFF
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -172,7 +172,7 @@ def load_env_file(path: Path) -> dict[str, str]:
         return env
     _check_file_permissions(path)
 
-    with open(path, 'r') as f:
+    with open(path, 'r', encoding='utf-8-sig') as f:
         for line in f:
             line = line.strip()
             if not line or line.startswith('#'):


mvanhorn:
Rebased your branch onto main and dropped the stale CHANGELOG hunk so this could land as-is, then merged. You reported #714 and filed first, and utf-8-sig is the right call for the Notepad BOM case. A regression test from #771 and a locale fallback from #780 are stacking on top.

PR #718, opened 2026-06-30
Fix X search on Windows + Node 24 (UTF-8 decode + tolerate Bird exit-crash)

spanner20:
## Problem

On **Windows 11 / Node v24.14.0 / Python 3.12**, the bundled Bird X backend returned **0 results** even with valid `AUTH_TOKEN`/`CT0` cookies (auth check passed). Two separate Windows-specific bugs:

**1. UTF-8 decode crash.** `_run_bird_search` spawns the Node CLI with `subprocess.Popen(text=True)` and no `encoding=`, so on Windows the child stdout is decoded as **cp1252**. Any tweet with emoji or non-Latin text (CJK / Arabic / etc.) raises `UnicodeDecodeError: 'charmap' codec can't decode byte 0x81` in the reader thread → search returns 0 results.

**2. Valid results discarded on a benign exit-crash.** The vendored `bird-search.mjs` uses native `fetch` (undici). On Windows + Node 24, calling `process.exit()` while undici keep-alive sockets are still closing trips a libuv assertion:

```
Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76
```

The process exits **non-zero AFTER** it has already written a **complete, valid JSON result** to stdout. The wrapper treated any non-zero exit as failure and discarded the good output.

## Fix (`scripts/lib/bird_x.py`, search path only)

- Pin `encoding="utf-8", errors="replace"` on the `Popen` call so non-ASCII tweet text never crashes the reader thread.
- Trust parseable stdout regardless of exit code; only treat a non-zero exit as a real failure when stdout is **empty or not valid JSON**. This tolerates the benign Windows/Node 24 exit-crash without masking genuine failures.

No behavior change on Linux/macOS (python emits LF / exits 0 there).

## Verification

Windows 11 / Node v24.14.0 — a `Claude Code` search now returns live tweets including CJK/Arabic text (previously 0 results). Auth check was already passing; this fixes the search path only.

### DIFF
--- a/scripts/lib/bird_x.py
+++ b/scripts/lib/bird_x.py
@@ -177,6 +177,8 @@ def _run_bird_search(query: str, count: int, timeout: int) -> Dict[str, Any]:
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             text=True,
+            encoding="utf-8",
+            errors="replace",
             preexec_fn=preexec,
             env=_subprocess_env(),
         )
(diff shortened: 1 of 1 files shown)


sdhilip200:
I opened a small supplemental PR for the Phase 2 handle-search gap from the review: https://github.com/spanner20/last30days-skill/pull/1

It applies the Windows/Node 24 Bird subprocess handling to `search_handles` as well: explicit UTF-8 replacement decoding, plus parsing valid JSON stdout before treating a non-zero exit as fatal. Added regression coverage for both pieces.

Verified with:
- `uv run pytest tests/test_bird_x.py -k 'HandleSearchRuntime'` -> `2 passed, 7 deselected`
- `uv run pytest tests/test_bird_x.py tests/test_pipeline_v3.py` -> `45 passed`
- `git diff --check`

I also ran the full suite on this PR branch; it currently reports `15 failed, 1024 passed` due to failures outside the touched Bird files, including missing `scripts/generate-synthesis-inputs.py` and store/watchlist/resolve expectations. `bash scripts/sync.sh` also fails because `variants/open/` is absent on this branch, so I left those branch-structure issues out of the supplemental fix.

mvanhorn:
Closing: both parts of this are now on main - the exit-code tolerance landed via #813 and the UTF-8 decode handling was covered by the subproc changes earlier. Thanks for flagging the Windows/Node 24 combination, it drove the fix getting prioritized.

PR #724, opened 2026-06-30
env: register GITHUB_TOKEN so .env value is not silently ignored

23241a6749:
## Problem

`GITHUB_TOKEN` is missing from the `env.py` keys tuple. Users who set `GITHUB_TOKEN` in their `.env` file find it silently ignored:

- `config.get('GITHUB_TOKEN')` returns `None` (it was never loaded from `.env`)
- `resolve_token()` in `github.py` falls through to `os.environ.get('GITHUB_TOKEN')` — also `None`, since `.env` values are not propagated to `os.environ` for unregistered keys
- Falls back to `gh auth token` subprocess or returns `None`, resulting in unauthenticated GitHub API usage (severe rate limits)

## Fix

Register `('GITHUB_TOKEN', None)` in the `env.py` keys tuple (one line). Same pattern as every other env var in the list.

## Impact

Users who explicitly configure `GITHUB_TOKEN` in `.env` will now have it picked up correctly. No behavior change for users relying on the `gh` CLI fallback.


mvanhorn:
Closing in favor of #782, but not because this was wrong in spirit: registering the token as a config key without also teaching lib/github.py to read config would let doctor claim auth the fetcher won't use. What your PR is really pointing at, first-class GITHUB_TOKEN support via .env and keychain end-to-end, is a feature I'd take. If you want it, file the issue and it's yours; your setup-script additions would carry straight over.

PR #732, opened 2026-07-01
env: register LAST30DAYS_REPORT_CACHE_TTL_SECONDS so .env value is picked up

23241a6749:
## Problem

`LAST30DAYS_REPORT_CACHE_TTL_SECONDS` was missing from the env.py keys tuple. Users who set this value in their `.env` file found it silently ignored:

- `_report_cache_ttl_seconds()` at `last30days.py:662` reads from `os.environ` first, then falls back to `config.get()`
- `config.get()` always returned `None` because the key was never loaded from `.env` into the config dict
- The default of 3600s was used regardless of the user's `.env` setting

This is the same bug pattern as FUN_LEVEL (#707) and GITHUB_TOKEN — a missing entry in the env.py keys tuple.

## Fix

Register `('LAST30DAYS_REPORT_CACHE_TTL_SECONDS', None)` in the env.py keys tuple (one line). Same pattern as every other env var in the list.

The flag is already fully documented in CONFIGURATION.md and the HTML reference — no doc changes needed.

## Impact

Users who explicitly configure `LAST30DAYS_REPORT_CACHE_TTL_SECONDS` in `.env` will now have it picked up correctly. No behavior change for users who rely on the `os.environ` path or the default.


mvanhorn:
Merging, thanks. None default is fine since the consumer falls back to 3600 on its own.

Issue #384, opened 2026-05-13
RFE: per-source return shape should signal failure mode (rate-limited / unreachable / auth-failed) vs no-results


lustrousgorilla:
## Background

The four upstream PRs filed today (#380, #381, #382, #383) all share a single root cause beyond their individual mechanics: every failure mode in the per-source `lib/<source>.py` modules is currently laundered to the orchestrator as `{"items": []}` (sometimes with an `"error"` key attached, sometimes not). The orchestrator and downstream callers have no reliable way to distinguish:

- **Genuine no-results** — the query ran successfully, just nothing matched.
- **Rate-limit / interstitial / transient anti-bot** — retry might succeed; the user should be informed the source was effectively skipped.
- **Authentication failure** — the API key is missing, expired, or revoked; the user should be told to fix it, not "no signal in HN this month."
- **DNS / unreachable / network** — the source is degraded; the user should know the result is partial.
- **Schema drift** — the upstream API changed shape; the wrapper is broken and the user should know that, not see "0 results."

All four PRs filed today were classic instances of this contract gap. PRs #380 and #381 are deterministic bugs (an HTTP 400 and an over-strict post-filter) — but the reason they went *unnoticed* long enough to surface during a community-signal pass is that they returned `[]` instead of raising. PRs #382 and #383 add retry logic for transient cases — but a user who runs `/last30days` while their network is degraded still has no way to know that "Reddit returned nothing" means "Reddit was unreachable for the entire 3-attempt budget" versus "Reddit had nothing on the topic."

## Proposal

Extend the per-source return shape with a `status` enum and an optional `diagnostic` string, while preserving backwards compatibility with the existing `items`/`error` keys. Concretely:

```python
{
    "status": "ok" | "empty" | "rate_limited" | "auth_failed" | "unreachable" | "schema_error" | "timeout" | "error",
    "items": [...],
    "diagnostic": "Twitter served HTML anti-bot interstitial after 2 retries",
    "error": "..."  # legacy field, kept populated for existing consumers
}
```

The status enum:

- **`ok`** — request succeeded, results returned (may still be `items: []` if the query genuinely matched nothing AND the source distinguishes "ran cleanly" from "didn't run"; otherwise prefer `empty` for the no-results case).
- **`empty`** — request succeeded, no items matched. Distinct from `ok` only when the source explicitly wants to signal "completed cleanly with zero hits."
[...]

PR #753, opened 2026-07-04
feat(doctor): unified doctor health command with fix prescriptions

mvanhorn:
## Summary

`/last30days doctor` gives users and agents one answer to "what's broken, what's serving, and what do I run to fix it." Before this, health knowledge was split across `--diagnose`, `--preflight`, `lib/health.py`, and post-run nudges, and none of them could say which backend was actually serving or name the exact fix. The result was the silent-degradation issue class: a stale yt-dlp shim reported as available, X quietly failing over across four backends with no visibility, sources dropping out with nothing to check.

What's now possible:

- One command reports every source's tier (ok / warn / off / error), the specific probe state behind it, and a copy-pasteable fix for anything non-ok. `--json` is the machine contract; `--cached` serves a fingerprinted 15-minute cache so the agent's pre-research check costs a file read.
- A binary that resolves on PATH but cannot exec now reads `broken` with a reinstall prescription, not "available" (the stale-shim false negative). An on-disk-but-off-PATH binary reads `missing` with a PATH fix, including the Windows PrintingPress install dir.
- Chained sources render truthful routing: a predicted `will use: bird` for X (prediction semantics, not a serving claim), and Reddit rendered honestly as "public keyless (default); ScrapeCreators backfill below floor" because its routing is per-query and cannot be probed into a single winner.
- Fix prescriptions live in one registry with both natural-language and CLI forms, shared by doctor and quality nudges so the two surfaces cannot drift.

## Design decisions

- `doctor` is a topic-word command mirroring `setup`, so the slash form works on every harness.
- Legacy `--diagnose`/`--preflight` JSON shapes are frozen by characterization tests for their two real consumers (the Go MCP preflight passthrough and SKILL.md's `available_sources` read). New data appears only in `doctor --json`.
- Doctor makes zero network calls, and forbid-network test harnesses prove it: xurl is probed via its local token store (never `xurl whoami`), paid lanes by key presence only, and the safe/diagnose path cannot reach the network through any of its three former vectors.
- The cache carries a schema version and a config fingerprint (key-presence booleans and pin values, never secrets); credential or pin changes invalidate it, `generated_at`/`from_cache` make staleness visible, and a drifted cache falls through to a live run instead of crashing.
- Per-chain probe budgets are additive by design (documented); cross-source probes run concurrently, bounding a full doctor run at roughly the slowest single probe.

## Test plan

Around 150 new tests across seven files: probe taxonomy (including a real dead-shebang shim integration test), backend descriptor parity against the pipeline's actual X selection, prescription completeness lint, doctor tier rollup and per-source exception isolation, cache corruption/drift/fingerprint/TTL scenarios, forbid-network harnesses on the whole doctor path, Windows off-PATH regression, and SKILL.md/CONFIGURATION.md contract locks. Full suite passes with the coverage gate at 85.21% against the 84% floor. Live smokes: fresh-config `doctor` and `doctor --json` (18 sources, 0.42s), byte-identical `--cached` serve, `check-config.sh` clean.

Review: simplification pass (10 findings applied), full multi-persona code review plus an independent cross-model adversarial pass and 11 per-finding validators; 15 of 16 validated findings fixed in the three `fix(review)` commits, 1 resolved by documenting the additive chain budget.

## Known residuals

Accepted advisory items, none actionable-blocking:

- The cache secret-refusal scan matches raw substrings and can miss JSON-escaped secret values (defense-in-depth on top of booleans-only output).
- A future-timestamped cache never expires until wall clock passes it (no clock-skew tolerance; explicit `doctor` always runs live).
- The Printing Press npm version pin exists in both `health.py` and `setup_wizard.py` constants, tied by comment only.
- No MCP `doctor` tool yet; deferred to an MCP-focused pass per the plan (the MCP bundle has separate version-drift work pending).

- Watch incoming GitHub issues for doctor misreports, especially Windows off-PATH detection and xurl local-auth detection against older `~/.xurl` formats.
- Healthy signal: fewer "source silently missing" issues of the #692 class; users self-serving fixes from prescriptions.
- Validation: run `/last30days doctor` on a fresh install per host. Claude Code and Codex smokes done; OpenClaw/Gemini/Hermes worth a spot check in the first week.
- Rollback: revert this PR. No data migrations; the only persisted artifact is `doctor-cache.json`, which self-heals (schema-stamped, falls through live).
- Owner: mvanhorn; validation window: first week post-merge.

## Merge sequencing

This branch is based on 9a625dd, before the uncommitted v3.9.0 source wave. When v3.9.0 lands, rebase and re-record the two `--diagnose`/`--preflight` characterization snapshots in `tests/test_diagnose_compat.py`; the test comments carry the instruction.

## Related

Related: #692, #384. Both reproduction scenarios are covered by tests in this PR; verify on a fresh install and comment with the shipped behavior before closing.

---

Claude Code


mvanhorn:
Rebased onto main (v3.10.0). Changes since the initial Greptile pass:

- Dropped the stale local hosted-API commit (9a625dd) that rode along in the original push; main's #747 implementation is the canonical one, so this PR is now doctor-only (the Greptile note on hosted.py belongs to #747's code, not this PR).
- Re-recorded the diagnose/preflight characterization snapshots against the committed v3.10.0 baseline (new sources, x_pending_browser_auth, PP CLI external commands), exactly the re-record path the test comments documented.
- Extended the no-network gate through main's new x_pending precompute in diagnose(), keeping the doctor path provably offline on the new baseline.
- Fixed the P1: quality_nudge's alt_cli destructuring is now arity-tolerant (degrades wording instead of crashing), with a pin test keeping both YouTube entries at two platform alternates.
- Addressed the P2: an autouse conftest fixture resets the probe/availability memo caches around every test.

Full suite green with coverage at 84.96% vs the 84% floor. Two pre-existing failures in tests/test_github_device_auth.py reproduce identically on clean origin/main and are unrelated to this branch.

PR #780, opened 2026-07-08
fix(env): locale-decoder fallback for non-UTF-8 .env files

### LINKED ISSUE #714: Bug: load_env_file() opens .env without utf-8 encoding (Windows UnicodeDecodeError)

SeanGearin:
Adds a locale-decoder fallback on top of the utf-8-sig read from #715: a genuinely locale-encoded `.env` (e.g. cp1252) that loaded before keeps loading instead of crashing. If it decodes as neither utf-8 nor the locale encoding, `UnicodeDecodeError` surfaces rather than corrupting keys/secrets with replacement characters.

Tests cover the UTF-8 BOM case and the locale-fallback case (deterministic via a forced cp1252 locale).

(Trimmed by the maintainer: the core utf-8-sig read landed via #715. Closes #714 together with it.)

### DIFF
--- a/skills/last30days/scripts/lib/env.py
+++ b/skills/last30days/scripts/lib/env.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import datetime
 import json
+import locale
 import os
 import sys
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


mvanhorn:
Edited your branch: the core utf-8-sig fix landed via #715 (first-filed by the issue reporter), so I trimmed this to your unique delta, the locale fallback and both tests, and merged. Also corrected the PR body, which said the fallback uses errors='replace' while the code deliberately lets UnicodeDecodeError surface. The code was right, the description wasn't.

PR #775, opened 2026-07-08
fix: wrap open() calls in 'with' to close file descriptors promptly (#774)

23241a6749:
## Summary

Two call sites used `open(plan_str).read()` without a `with` statement, leaving the file descriptor open until garbage collection:

1. **`parse_competitors_plan()` at `last30days.py:440`** — reads a `--competitors-plan` JSON file.
2. **`main()` at `last30days.py:1174`** — reads a `--plan` JSON file.

While CPython's refcounting closes these promptly, PyPy and other implementations may defer finalization. In a long-running or descriptor-constrained environment, leaked handles can accumulate.

Both sites already had proper `try/except` wrapping for `OSError` / `UnicodeDecodeError`, so this is purely a resource-management fix with no behavior change.

## Changes

| File | Change |
|---|---|
| `last30days.py:440` | `open(...).read()` → `with open(...) as f: f.read()` |
| `last30days.py:1174` | `open(...).read()` → `with open(...) as f: f.read()` |

Fixes #774.

PR #781, opened 2026-07-08
test(cli): guard against bare open().read() regressions in last30days.py

SeanGearin:
Adds an AST test asserting no `open(...).read()` appears outside a `with` block in `skills/last30days/scripts/last30days.py`, so the fd-leak pattern fixed in #775 can't regress.

(Original fix hunks landed via #775; this PR was trimmed by the maintainer to the regression test.)

### DIFF
--- a/tests/test_no_bare_open_read.py
+++ b/tests/test_no_bare_open_read.py
@@ -1,0 +1,37 @@
+"""Prevent direct ``open(...).read()`` calls in the CLI entrypoint."""
+
+from __future__ import annotations
+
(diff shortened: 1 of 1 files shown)


mvanhorn:
Heads up that I edited your branch (edits-by-maintainers is on): the two with-block fixes landed via #775, which was filed first with the identical change, so I trimmed this down to your AST regression test and merged it. The test is the part nobody else brought - it makes the whole pattern unregressable. Thanks.

PR #782, opened 2026-07-08
fix(doctor): detect GITHUB_TOKEN from process env

SeanGearin:
`doctor`'s GitHub check reads `config.get("GITHUB_TOKEN")`, but `GITHUB_TOKEN` isn't a registered config key, so it's always `None` — while the actual fetcher (`lib/github.py`) reads it from the environment via `env.read_secret_env`. As a result, a user with `GITHUB_TOKEN` exported (and no `gh` CLI) is reported as unauthenticated even though GitHub calls authenticate fine.

Fix: `doctor` detects the token the same way the fetcher does. Adds two tests (env token without gh → authenticated; neither → unauthenticated).

### DIFF
--- a/skills/last30days/scripts/lib/doctor.py
+++ b/skills/last30days/scripts/lib/doctor.py
@@ -303,7 +303,7 @@ def _polymarket_record(config):
 
 
 def _github_record(config):
-    authed = bool(config.get("GITHUB_TOKEN") or shutil.which("gh"))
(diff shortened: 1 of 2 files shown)


mvanhorn:
Merging. For the record on why this one over #724 and #777: the fetcher reads GITHUB_TOKEN from process env only (read_secret_env), so registering it as a config key would make doctor report authenticated for a .env-stored token the fetcher never sees. Mirroring the fetcher is the only version where doctor tells the truth.

PR #783, opened 2026-07-08
test(env): documented env keys must be registered in get_config

SeanGearin:
Adds a completeness test asserting every env key documented in `SKILL.md`/`README.md` is registered in `get_config()`, so the silently-ignored-key class (#707, #732, #724) can't recur. Also round-trips `FUN_LEVEL` and `LAST30DAYS_REPORT_CACHE_TTL_SECONDS` from a `.env` file.

(Trimmed by the maintainer: the two registration lines landed via #708 and #732.)

### DIFF
--- a/tests/test_env_doc_contract.py
+++ b/tests/test_env_doc_contract.py
@@ -1,0 +1,82 @@
+from __future__ import annotations
+
+import re
+from pathlib import Path
+from unittest import mock
+
+from lib import env
+
+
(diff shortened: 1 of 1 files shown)


mvanhorn:
Edited your branch: FUN_LEVEL and TTL registration landed via #708 and #732, both first-filed, so I trimmed this to the doc-contract completeness test and merged. That test is the real prize here, it kills the entire silently-ignored-key class.

PR #757, opened 2026-07-05
fix: prevent save_output from silently overwriting date-stamped files (#755)

23241a6749:
## Problem

Fixes #755

`save_output()` at `last30days.py:139-141` had a one-shot collision guard. When the base filename existed it fell back to a date-stamped name — but then called `write_text` unconditionally with no second existence check:

1. **First run:** creates `topic-raw.md`
2. **Second run:** base exists → creates `topic-raw-2026-07-05.md`
3. **Third run:** base exists → creates `topic-raw-2026-07-05.md` again → **silently overwrites the second run**

No warning, no error, no indication the file was replaced.

## Fix

Replace the one-shot fallback with a loop that appends a counter suffix until a unique filename is found. After the fix:

1. First run: `topic-raw.md`
2. Second run: `topic-raw-2026-07-05.md`
3. Third run: `topic-raw-2026-07-05-1.md`
4. Fourth run: `topic-raw-2026-07-05-2.md`

## Impact

No more silent data loss on repeated runs within the same calendar day. All existing tests pass unchanged (they only exercise the single-write path).

Issue #755, opened 2026-07-04
[bug] save_output silently overwrites the date-stamped file on the third run of the same day


bastitva0-blip:
### Summary

<h2>Summary</h2>

<p>
Running <code>--save-dir</code> three or more times in one calendar day causes the third run to
silently overwrite the second run's saved output file. No warning, no error, no indication the
file was replaced.
</p>

<h2>Root Cause</h2>

<p>
<code>save_output</code> in <code>scripts/last30days.py</code> lines 139–141 has a one-shot
collision guard. When the base filename exists it falls back to a date-stamped name — but then
calls <code>write_text</code> unconditionally with no second existence check.
</p>

<pre><code>out_path = path / f"{slug}-{raw_label}{suffix_part}.{extension}"
if out_path.exists():
    out_path = path / f"{slug}-{raw_label}{suffix_part}-{datetime.now().strftime('%Y-%m-%d')}.{extension}"
# no check here — write_text follows unconditionally
out_path.write_text(content, encoding="utf-8")</code></pre>

### Steps to Reproduce

<h2>Steps to Reproduce</h2>

<ol>
  <li>Run <code>last30days &lt;topic&gt; --save-dir ~/some/dir</code> — creates <code>&lt;topic&gt;-raw.md</code></li>
  <li>Run the same command again — base file exists, creates <code>&lt;topic&gt;-raw-2026-07-05.md</code></li>
  <li>Run the same command a third time — both files exist, silently overwrites <code>&lt;topic&gt;-raw-2026-07-05.md</code></li>
</ol>

### Expected Behavior

<h2>Expected Behavior</h2>

<p>
The third run should write to a new unique filename (e.g. <code>&lt;topic&gt;-raw-2026-07-05-1.md</code>)
rather than overwriting the existing dated file.
</p>
<details>
<summary>Suggested fix</summary>

<pre><code>out_path = path / f"{slug}-{raw_label}{suffix_part}.{extension}"
if out_path.exists():
    date_str = datetime.now().strftime('%Y-%m-%d')
    candidate = path / f"{slug}-{raw_label}{suffix_part}-{date_str}.{extension}"
    counter = 1
    while candidate.exists():
        candidate = path / f"{slug}-{raw_label}{suffix_part}-{date_str}-{counter}.{extension}"
        counter += 1
    out_path = candidate
out_path.write_text(content, encoding="utf-8")</code></pre>
</details>

### Error / Traceback

```text
None. The overwrite is completely silent.
[...]

PR #784, opened 2026-07-08
fix(hosted): don't overwrite same-day saved reports

SeanGearin:
Ports the exclusive-create collision handling that landed in #757 to the hosted save path (`lib/hosted.py:_save_output`), which had the same same-day overwrite bug. Covers both the local and hosted paths with tests (adapted to #757's `-1`, `-2` counter naming).

(Reworked by the maintainer on top of #757; the local-path half of the original PR landed there first.)


mvanhorn:
Reworked this on your branch: #757 landed first for #755 and its exclusive-create is stronger where the two overlap, so the local path was done. Your hosted.py coverage was the half nobody else had, so I ported the same O_EXCL approach there and kept your test (adapted to #757's -1, -2 counter naming). Merged, thanks.

SeanGearin:
the O_EXCL port to hosted.py is cleaner than what I had - thanks for taking the time

PR #821, opened 2026-07-13
fix(http): sanitize non-ASCII in URL before urllib.request to prevent latin-1 encode crash (fixes #817)

### LINKED ISSUE #817: UnicodeEncodeError ('latin-1' codec) on Arabic-script topics — TikTok/Instagram (ScrapeCreators) requests fail
### Summary

Running /last30days with an Arabic topic (and even an English topic resolved from an Arabic context) causes the TikTok and Instagram sources to fail with UnicodeEncodeError: 'latin-1' codec can't encode character '\u0627'. The Arabic query string appears to be passed into the HTTP request (URL/query or header) without proper URL-encoding (percent-encoding / UTF-8), so Python's http layer falls back to latin-1 and crashes. All non-Latin-script topics (Arabic, and likely CJK, Cyrillic, etc.) lose TikTok/Instagram coverage entirely, which are key sources for non-English markets.

### Steps to Reproduce

1. Fresh install of the skill (v3.14.0) in Claude Code on macOS, manual setup.
2. SCRAPECREATORS_API_KEY configured in ~/.config/last30days/.env.
3. Run: /last30days اسعار التمريض المنزلي السعودية
   (topic = "home nursing prices Saudi Arabia" in Arabic)
4. TikTok and Instagram sources both fail with the same UnicodeEncodeError.
5. Re-running with the English topic "home nursing prices Saudi Arabia" still failed the same way in my run (the engine seems to include the original Arabic string somewhere in the request pipeline).

### Expected Behavior

Arabic (and any non-Latin) query strings should be UTF-8 percent-encoded before being placed in the request URL/params (e.g. urllib.parse.quote / params= with requests), so TikTok/Instagram searches run normally and return Arabic-language content.

### Error / Traceback

```text
Instagram: UnicodeEncodeError: 'latin-1' codec can't encode character '\u0627' in position 28: ordinal not in range(256)
TikTok: UnicodeEncodeError: 'latin-1' codec can't encode character '\u0627' in position 28: ordinal not in range(256)
(Same error on both sources, position 28, character U+0627 ARABIC LETTER ALEF. Reported in the raw report's "Source Errors" section, run doctor shows the same.)
```

### Install Method

Claude Code plugin

### OS

macOS (Apple Silicon MacBook Air), zsh, Claude Code CLI (native installer)

23241a6749:
Fixes #817

## Problem

Non-Latin-script topics (Arabic, CJK, Cyrillic, etc.) cause TikTok and Instagram ScrapeCreators requests to crash with:

> ```
> UnicodeEncodeError: 'latin-1' codec can't encode character '\u0627' in position 28: ordinal not in range(256)
> ```

## Fix

Added a `urllib.parse.quote` pass on the URL string after `urlencode` construction in `http.request()`, using a safe set that preserves all valid URL structural characters. This guarantees the URL is always 100% ASCII-safe before being passed to `urllib.request.Request`.

`safe='/:@!$&\'()*+,;=-._~%?#[]=+'`

## Testing

- `test_http_v3.py`: 14/14 passed
- `test_grounding_v3.py`: 28/28 passed
- `test_env_v3.py`: 13/13 passed

### DIFF
--- a/skills/last30days/scripts/lib/http.py
+++ b/skills/last30days/scripts/lib/http.py
@@ -13,7 +13,7 @@ from contextlib import contextmanager
 from contextvars import ContextVar, copy_context
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
-from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
+from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit, quote
 
 from . import health
 from . import log as _log
@@ -556,6 +556,9 @@ def request(
         if filtered:
             separator = "&" if ("?" in url) else "?"
             url = f"{url}{separator}{urlencode(filtered)}"
+    # Encode any non-ASCII characters to prevent UnicodeEncodeError from
+    # http.client.HTTPConnection.putrequest (which uses latin-1 internally).
+    url = quote(url, safe='/:@!$&\'()*+,;=-._~%?#[]=+')
 
     fixture_request = _fixture_request(method, url, json_data, raw)
     fixture_redactions = _fixture_redactions(url, headers, json_data)


mvanhorn:
Closing this one: #822 (just merged) carried a component-wise version of the same fix, encoding path/query/fragment while leaving the hostname alone, which handles non-ASCII domains better than the whole-URL quote here. Thanks for running down #817, the fix that landed builds directly on your diagnosis.

PR #824, opened 2026-07-15
fix: cap topic slug so long topics don't crash save_output (ENAMETOOLONG)

Mekiwi:
## Problem

A research run whose topic is a long natural-language question (a full pasted prompt, not a short phrase) crashes at save time:

```
OSError: [Errno 36] File name too long:
  '…/<run-dir>/i-m-making-a-video-in-davinci-resolve-i-need-some-transitions-…-a-codex-subscription-raw.md'
```

`save_output` builds the artifact filename as `slugify(topic) + "-raw.md"`, and `slugify` never caps length. A ~290-char topic → a ~297-byte single filename component, past Linux/macOS `NAME_MAX` (255 bytes).

The nasty part: in `_render_save_and_print`, `save_output()` runs **before** `print(rendered)`. So a fully successful run (all sources returned, report rendered) is thrown away on the way to disk and surfaces to the caller as a hard failure with no answer.

Repro:
```bash
python3 skills/last30days/scripts/last30days.py "$(python3 -c 'print("a "*150)')" --emit=compact --save-dir /tmp/x
```

## Fix

Cap the filename slug at 80 chars via `_cap_filename_slug` (the slug is ASCII `[a-z0-9-]`, so the char cap is also a byte cap, leaving headroom for the `-raw`/date/`-NN`/extension suffixes). The save directory already disambiguates runs, so the slug is purely a human label — truncating it loses nothing functional.

## Test

Adds `test_save_output_caps_long_topic_filename` (in `tests/test_cli_v3.py`) — asserts a long topic produces a `<=255`-byte `*-raw.md` that actually writes. It errors with `ENAMETOOLONG` on `main` without the fix.

## Note

`test_cli_v3.CliV3Tests.test_main_writes_rendered_output_to_explicit_file` fails on a clean `main` checkout in my environment (`emit_output` called twice) — it is **pre-existing and unrelated** to this change (this PR does not touch that path).


mvanhorn:
Thanks for catching this and for the clean regression test. I merged #786 for the same bug since it patches slugify itself (covers every filename call site and hash-suffixes to keep two long topics distinct), so I'm closing this one. The diagnosis here was spot on, and your #825 is still on my list.

PR #825, opened 2026-07-15
fix: keyless web search on DuckDuckGo-blocked IPs + stop reddit enrichment poisoning web results

Mekiwi:
Two independent bugs made the **keyless web-search floor return nothing** on datacenter/VPS hosts (found while debugging a self-hosted deployment where `grounding` consistently reported `Web: 0 items — HTTP 403: Blocked`).

## Bug 1 — DuckDuckGo anomaly-blocks the IP, and it was the only HTML rung

`html.duckduckgo.com/html/` returns an HTTP **202 anomaly/challenge page** ("unfortunately, bots use DuckDuckGo too") for datacenter IPs — every method (GET/POST) and endpoint (`/html/`, `/lite/`). With no result anchors parsed, `keyless_search` fell straight to `keyless-search-unavailable` (public SearXNG instances 403/429 from the same IPs or disable the JSON API).

**Fix:** add **Startpage** as a second keyless rung — `ddg → startpage → searxng`. Startpage returns organic results to a plain browser-UA GET where DuckDuckGo refuses. Also hardened `_strip_html` to strip `<style>`/`<script>` *contents* (Startpage inlines emotion CSS inside the title anchor, which would otherwise leak into titles/snippets).

## Bug 2 — a reddit enrichment 403 discarded otherwise-good web results

Once results came back, any `reddit.com/comments/` URL among them triggered `_enrich_reddit_items`, which fetches the thread JSON. That fetch **403s on a datacenter IP**. Although `_enrich_reddit_items` swallows the error, the `http` layer had already recorded the 403 into the active `capture_failures` sink, and `_resolve_stream_outcome` then prioritized that captured failure over the returned items — reporting the whole web source as failed (`0 items, HTTP 403`) and throwing away the good results.

**Fix:** wrap the secondary enrichment in its own `http.capture_failures()` sink so its failures can't be attributed to the web/grounding source outcome.

## Tests

- `test_web_search_keyless.py`: Startpage used when DDG empty, DDG still preferred when it has results, `_strip_html` drops style blocks.
- `test_grounding_v3.py`: `RedditEnrichmentIsolationTests` — an enrichment 403 leaves the source's failure sink clean and keeps the items (verified failing on the pre-fix code as a negative control).

All existing `test_grounding_v3` (29) and `test_web_search_keyless` (now 9) pass.

## Note

Independent, pre-existing test failure unrelated to this change: `test_cli_v3.CliV3Tests.test_main_writes_rendered_output_to_explicit_file` (`emit_output` called twice) fails on a clean `main` in my environment — not touched here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Keyless web search now works on hosts where DuckDuckGo anomaly-blocks the egress IP (a 202 challenge page with no results — common on datacenter/VPS IPs). Added Startpage as a second keyless rung (DuckDuckGo → Startpage → configured SearXNG), so the web floor still returns results there. Also hardened `_strip_html` to drop `<style>`/`<script>` contents so inline CSS can't leak into a title or snippet.
+- Web/grounding results are no longer discarded when one of them is a reddit.com URL whose enrichment fetch fails. Reddit enrichment is a best-effort secondary fetch; its HTTP failures (e.g. a 403 on a datacenter IP) were being attributed to the whole web source, which then reported "0 items — HTTP 403" despite having retrieved good results. Its failures are now isolated from the source's outcome.
+
 ## [3.14.0] - 2026-07-12
 
 ### Added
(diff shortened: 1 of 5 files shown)


mvanhorn:
Merged via cherry-pick onto main as 2626d0f with your commit and authorship preserved as-is. The only conflict was CHANGELOG placement after today's other merges, which I resolved on my side. Both fixes are live: the Startpage fallback rung and the reddit-enrichment failure isolation. Full test suite ran green before push. Thanks for both of these, closing since the commit is on main.

PR #840, opened 2026-07-19
fix(bird_x): normalize retries and preserve clean empty results

spiky02plateau:
## Summary

Bird's retry ladder can inherit quote/group punctuation from a plan query. When core-subject extraction leaves a partial wrapper, fallback queries can become syntactically invalid; a clean initial zero-result response can then be overwritten by the failed retry and reported as a backend failure.

This keeps Bird fallback queries lexical and preserves a valid empty result when only an optional recall-expansion retry fails.

## Changes

- Strip quote/group wrappers from extracted Bird query tokens before building the initial and fallback queries.
- Remember the latest response that completed without an error.
- Preserve that clean empty response if a later optional retry fails.
- Preserve the existing failure behavior when every attempt fails.
- Add behavior-level regression coverage for malformed wrappers and both empty/error outcomes.

## Testing

- [x] Ran `python -m pytest -q --tb=short` — 3,167 passed, 5 skipped, 23 subtests passed.
- [x] Ran `git diff origin/main...HEAD --check`.

## Related Issues

Related to #384. This complements the typed failure-vs-no-results semantics from #797; it prevents a successful empty Bird search from being overwritten by failure in an optional retry.

### DIFF
--- a/skills/last30days/scripts/lib/bird_x.py
+++ b/skills/last30days/scripts/lib/bird_x.py
@@ -111,6 +111,16 @@ def _extract_core_subject(topic: str) -> str:
     return extract_core_subject(topic, max_words=5, strip_suffixes=True)
 
 
+def _plain_query_tokens(text: str) -> list[str]:
+    """Return lexical tokens without Bird query grouping syntax."""
+    separators = str.maketrans({char: " " for char in '\"“”()[]{}'})
+    return [
+        clean
+        for token in text.translate(separators).split()
+        if (clean := token.strip("'‘’"))
+    ]
+
+
 def is_bird_installed() -> bool:
     """Check if vendored Bird search module is available.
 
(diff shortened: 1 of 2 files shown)


spiky02plateau:
Addressed both review notes in 1ec97ec: renamed the state to `last_clean_response` and mocked `lib.query.extract_compound_terms` so the OR-group path is deterministic. Full suite remains green: 3,167 passed, 5 skipped, 23 subtests.

spiky02plateau:
Late spec review found one additional edge case: embedded unmatched grouping syntax such as `foo(bar` survived boundary-only stripping. Addressed in 89e699a by treating grouping/double-quote syntax as token separators everywhere while preserving internal apostrophes. The direct repro and full suite are green (3,167 passed, 5 skipped, 23 subtests).

PR #850, opened 2026-07-20
Fix stale footer path after collision-safe save fallback

### LINKED ISSUE #833: [bug] Footer 'Raw results saved to' points at the stale base file when save_output falls back to the date-suffixed candidate
### Summary

When the base raw filename already exists, `save_output` correctly falls back to the date-suffixed candidate (the O_EXCL candidate chain from the #755 fix, 3.12.0) - but the compact footer still prints the base path. `compute_save_path_display` predicts only the base name and the footer is rendered before `save_output` runs, so the prediction can never see the collision. Following the footer lands you in the older (possibly empty) file.

### Steps to Reproduce

1. Run `python3 skills/last30days/scripts/last30days.py "some topic" --emit=compact` so that `some-topic-raw.md` exists in the save dir.
2. Run the same topic again the same day.
3. The engine writes `some-topic-raw-YYYY-MM-DD.md` (O_EXCL fallback works, no overwrite), but the run's footer still prints `Raw results saved to ~/.../some-topic-raw.md`.

Observed on v3.16.0 (2026-07-16, run with a custom `--suffix`): the save log wrote `<slug>-raw-<suffix>-2026-07-16.md` while the pass-through footer kept pointing at the older, empty un-suffixed file.

Code refs at v3.16.0:
- `scripts/last30days.py` `save_output` (~L159): candidate chain base -> `-YYYY-MM-DD` -> `-YYYY-MM-DD-N`, O_EXCL, returns the actually written Path.
- `scripts/last30days.py` `compute_save_path_display` (~L392): builds only the base candidate, no collision probing.
- Footer is computed (~L1486) before `save_output` is called (~L1527), so the display value is a prediction, not the real path.

Suggested fix direction: feed the footer from `save_output`'s returned Path (reorder render/save), or make the display helper walk the same candidate chain (racy under concurrency, the return-value route seems cleaner).

Related: #755 (fixed in 3.12.0) - this is a follow-on: the file is no longer overwritten, but the printed pointer went stale instead.

Thanks for your efforts @mvanhorn - highly appreciated!

### Expected Behavior

The footer's `Raw results saved to ...` line shows the path that was actually written in this run (the date-suffixed candidate when the collision fallback fired).

### Error / Traceback

```text
n/a - no crash. Symptom is a wrong path in the footer output.
```

### Install Method

Other

### OS

macOS 26.5

terry-li-hm:
## What changed

- Render the report footer only after `save_output` has reserved the actual collision-safe path with `O_EXCL`.
- Use that allocated path in both the saved report and stdout.
- Remove the reserved candidate if deferred rendering or writing fails.
- Add regression coverage for base and same-day filename collisions, unchanged existing files, cleanup on render failure, comparison output, and stdout consistency.

## Root cause

The footer path was predicted before `save_output` ran, so it could only name the base candidate and could not observe the dated fallback actually selected after a collision.

## Validation

- `.venv/bin/python -m pytest tests/test_cli_v3.py tests/test_render_footer.py -q`
- `.venv/bin/python -m pytest -q --tb=short`
- `uvx ruff check skills/last30days/scripts/last30days.py tests/test_cli_v3.py`

Closes #833.

### DIFF
--- a/changelog.d/850.fixed.md
+++ b/changelog.d/850.fixed.md
@@ -1,0 +1,1 @@
+Report footer path now matches the collision-safe path actually reserved by save_output.
--- a/skills/last30days/scripts/last30days.py
+++ b/skills/last30days/scripts/last30days.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# fmt: off
 # ruff: noqa: E402
 """last30days CLI."""
 
@@ -15,6 +16,7 @@ import signal
 import sqlite3
 import sys
 import threading
+from collections.abc import Callable
 from pathlib import Path
 
 MIN_PYTHON = (3, 12)
@@ -222,6 +224,7 @@ def save_output(
     json_profile: str = "agent",
     register: str = "default",
     private: bool | None = None,
+    render_fn: Callable[[Path], str] | None = None,
 ) -> Path:
     from datetime import datetime
     path = Path(save_dir).expanduser().resolve()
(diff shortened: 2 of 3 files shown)


tmchow:
@terry-li-hm

### Action needed: rebase onto `main` for the new towncrier release process

`main` just landed [#872](https://github.com/mvanhorn/last30days-skill/pull/872) — feature/fix PRs no longer edit `CHANGELOG.md` or bump version manifests. CI (`changelog-guard`) will fail PRs that still do the old thing.

**Docs:** [`changelog.d/README.md`](https://github.com/mvanhorn/last30days-skill/blob/main/changelog.d/README.md) · [`CONTRIBUTING.md`](https://github.com/mvanhorn/last30days-skill/blob/main/CONTRIBUTING.md)
#### 1. Rebase onto latest `main`

```bash
git fetch origin
git rebase origin/main
# resolve any conflicts, then:
git push --force-with-lease
```
#### Add a changelog fragment (required for engine/skill changes)

```bash
cat > changelog.d/850.fixed.md <<'EOF'
Saved-report footer path now matches the collision-safe save fallback path.
EOF
git add changelog.d/850.fixed.md
```

Or, if this PR truly has nothing for release notes: check **Skip changelog** in the PR template and add the `skip-changelog` label.

tmchow:
Maintainer update: rebased onto `origin/main` and added `changelog.d/850.fixed.md` for the towncrier release process (#872). Prior review feedback was already addressed (or none outstanding). No action needed from you.

PR #884, opened 2026-07-25
fix(signals): stop rescuing an all-off-topic social batch into the ranking

bekonyn:
## Summary

`prune_low_relevance` ended in `return filtered or items`, so a batch where *every* item failed the relevance floor was handed back whole. The floor therefore never applied to the case it matters most: a source whose entire yield is off topic.

Observed on a Power BI research run: all three X items were crypto promo (`@amnyamcrypto`, `@0x_zozo`, …) matching the single generic token "developers" in a 15-token query. All three were rescued, and the freshness plus engagement terms in `local_rank_score` floated one of them to #2 overall — above the run's genuinely on-topic Reddit thread — after which the report listed all three handles under `Top voices`.

Batches consisting only of social sources with no relevance gate of their own now prune to empty, which surfaces as `no-results` in `source_status` rather than as ranked spam. Two exemptions:

- **Reddit** keeps the rescue — `reddit.py` / `reddit_keyless.py` already apply `RELEVANCE_FLOOR` upstream, and the existing `test_prune_low_relevance_falls_back_when_all_are_weak` behavior is preserved.
- **Pinned lanes** (an explicitly named handle) opt back in via the new `rescue_ungated_social` argument: those posts are wanted whether or not the topic words appear in them. Without this, `_run_supplemental_searches` dropped handle-lane results whose text did not lexically match the topic.

## Testing

- [x] `uv run pytest` — 43 pre-existing failures on this Windows host, identical set before and after; no new failures.

`test_prune_drops_all_weak_x_batch` reproduces the three-spam-post batch; `test_prune_rescues_weak_x_batch_when_mixed_with_other_sources` pins the scoping so a mixed batch still keeps the net.

## Changelog

### AI review

Written with Claude Code. Main risk checked: zeroing a source that previously always returned something. The first draft applied the rule to every social source and broke four `test_pipeline_v3` X-lane tests — that surfaced the pinned-handle case, which is why the `rescue_ungated_social` opt-in exists rather than a blanket rule. The mixed-batch subset check (`sources_present <= _RESCUE_EXEMPT_SOURCES`) keeps the net for any batch containing a non-social source. Full-suite output was diffed against `main` on the same host to separate real regressions from the pre-existing Windows failures.

### Security

N/A — pure ranking logic, no input handling, execution, paths, auth, or dependencies.

## Notes

Worth flagging separately: `prune_low_relevance` is called per-source from `_normalize_score_dedupe`, so `sources_present` is always a single source and the `sole_source` branch in `_passes_engagement_floor` is therefore always true — the TikTok/Instagram 1000-view floor never actually fires in that path. Not touched here; happy to file it as its own issue.

- [x] None

## Related issues

N/A — found while running a research brief.


mvanhorn:
Speaking as Matt's firstmate:

**Triage:** Corrective (stop rescuing all-off-topic social batches) — blocked

- **Class:** corrective ranking/relevance-floor fix (changes default rescue behavior for wholly weak ungated social batches)
- **CI history:** checks were green at last review, but PR is now `CONFLICTING` / `DIRTY` vs main
- **Greptile:** 3/5 — pinned TikTok/Instagram explicit lanes can lose all-weak batches because `pinned` is not propagated the way supplemental X handles are

Conflict resolution and the pinned-lane fix are hard work → factory ship path only (not standing triage). Please rebase onto current main and address Greptile, or say if you want this flagged for factory ship. No merge proposal while conflicting and below Greptile 5/5.

PR #885, opened 2026-07-25
fix(rank): keep out-of-window evidence out of the top ranks

bekonyn:
## Summary

`date_confidence` was display-only. Items dated outside the run's window rendered a `[date:low]` tag and then competed for rank on equal terms — in a 2026-07 brief that put a 2025-10-15 video at cluster **#1** and a 2025-12-18 one at **#5**. Both were good videos; neither belongs at the top of a brief whose whole promise is the last 30 days.

Candidates whose *every* dated item falls outside the window are now demoted:

- `fusion._candidate_sort_key` orders the fused pool, and
- `rerank._final_score` — the score the report's final order actually comes from, where a strong LLM rerank score previously let a stale item climb straight back past the fusion ordering.

Scaling `final_score` rather than subtracting a constant preserves the ordering *among* older items, so they remain readable as a group underneath the in-window evidence rather than being flattened into noise. They are demoted, not dropped, matching the existing `[date:low]` disclosure rather than replacing it.

Candidates with **no** dated item are exempt: an unknown date is a coverage gap, not a stale item, and demoting them would bury thin-metadata sources for a problem they do not have.

Also wires the freshness verdict into the pass-through footer. `_assess_data_freshness` already produced "only 14 of 30 dated items are from the last 7 days" on the run that prompted this — correct, mid-report, and acted on by nobody. It now appears in the footer block readers actually scan, and stays silent when the evidence is recent or when the footer would otherwise be empty.

## Testing

- [x] `uv run pytest` — 43 pre-existing failures on this Windows host, identical set before and after; no new failures.

`TestOutOfWindowDemotion` (stale candidate stronger on every other signal still loses to an in-window one; undated candidates unaffected), `OutOfWindowSortTests` for the fusion ordering, and two footer tests covering the freshness line's presence and absence.

## Changelog

### AI review

Written with Claude Code. Risks checked: the undated-item case (the first cut keyed only on `date_confidence != "high"`, which would have demoted every item with no date — hence the explicit `dated` guard and its test); double demotion alongside the existing entity-miss penalty (they compose, and an item that is both stale and off-entity deserves both); and the footer's documented "returns empty when there is nothing to report" contract, which an early draft broke by making an otherwise-empty run announce its own emptiness — caught by `test_footer_empty_with_no_save_path_returns_nothing` and fixed by gating on a non-empty body.

`OUT_OF_WINDOW_FINAL_MULTIPLIER = 0.35` is a judgment call, picked to sit alongside the existing `0.3` low-rerank multiplier. If you would rather have out-of-window items hard-excluded from the ranked clusters and listed in their own "Older but relevant" block, that is a small change from here — I went with demotion as the less destructive default.

### Security

N/A — ranking and rendering only. No input handling, execution, paths, auth, secrets, or dependencies.

## Notes

`schema.candidate_out_of_window` is new and shared by both call sites, so the definition of "out of window" stays in one place.

- [x] None

## Related issues

N/A — found while running a research brief. Companion to #883 and #884 from the same run.


mvanhorn:
Thanks for this PR! We cherry-picked your recency-contract ranking fix into #993, plus follow-ups so clustering and date-window membership keep stale evidence out of the top ranks. Really appreciate the contribution.

Closing this in favor of #993. That replacement is CI-green and Greptile 5/5; it is not on main yet.

PR #898, opened 2026-07-27
feat(transcribe): add self-hosted Whisper provider (LAST30DAYS_WHISPER_URL)

josephkehan-prog:
## Summary

Caption-free transcription can now use a self-hosted, OpenAI-compatible endpoint. `LAST30DAYS_WHISPER_URL` is tried ahead of Groq and OpenAI: no key, no per-minute cost, and the audio never leaves the machine. Previously `_PROVIDER_ENDPOINTS` hardcoded the two hosted URLs and `OPENAI_BASE_URL` was honoured only by `providers.py` (the Responses path), so a local whisper-server was unreachable without editing the module.

Behaviour is unchanged when the variable is unset, and a local server that is down or erroring falls through to the hosted backstops through the existing per-chunk provider loop.

## Testing

- [x] `uv run pytest` — 3491 passed, 5 skipped, 41 subtests passed

New `TestLocalProvider` covers provider ordering (local → groq → openai), absence without the URL, endpoint/model resolution including the default and a pinned model, hosted entries staying untouched, `is_available` opening on the URL alone, and both header paths (omitted when unauthenticated, `Bearer` when a key is set).

Verified end to end outside the suite against whisper.cpp's `whisper-server` (`ggml-base.en`, `--inference-path /v1/audio/transcriptions`): real transcript text returned through the engine's own `_post_audio` with no key and no network egress.

## Changelog

- [x] Added `changelog.d/+local-whisper-provider.added.md`
- [ ] Skip changelog

## Agent disclosure

Opened by Claude Code (Opus 5) driven by the repo owner of the fork.

### AI review

Risks checked: (1) **Silent behaviour change for existing users** — verified `transcription_providers({})` still returns `[]` and the hosted endpoints resolve exactly as before, with an explicit test for each. (2) **Signature change blast radius** — `_post_audio` gained a keyword argument; grepped for every call site (one, in `_transcribe_chunk`) and found the in-repo test stub `post(provider, path, key, timeout)` was positionally coupled. The full suite caught it as a real failure before this PR was opened, and the stub is updated in the same commit rather than the assertion being loosened. (3) **Fallback integrity** — a missing or empty `LAST30DAYS_WHISPER_URL` makes `_post_audio` return `None`, so the existing per-chunk loop moves to the next provider instead of raising.

### Security

The new value is a URL read from the same config layer as every other setting; it is passed to `urllib.request.Request` unmodified and never logged, and no shell or path handling is involved. The header change reduces exposure rather than widening it: an empty bearer is no longer transmitted, and `LAST30DAYS_WHISPER_KEY` is optional and treated like the existing provider keys. Because the endpoint is operator-supplied it can point anywhere, which is the same trust model as `XIAOHONGSHU_API_BASE` and `OPENAI_BASE_URL` today. Tests use dummy values and a loopback URL; no real keys or fixtures added.

## Notes

The endpoint is the **full path** (matching how the hosted entries are written), so LocalAI, vLLM, and gateway setups work without special-casing. `CONFIGURATION.md` gains a "Self-hosted transcription" section with a runnable whisper.cpp example, and the source table row lists the new variable alongside the existing keys.

- [x] None

## Related issues

N/A


mvanhorn:
Speaking as Matt's firstmate:
Classified **opt-in** (self-hosted Whisper via `LAST30DAYS_WHISPER_URL`). Greptile **4/5** (preflight omits Whisper URL override). `mergeable_state=dirty` — needs author rebase + Greptile fix. Proposing stale-close (14d+ waiting-author, not CLEAN). Hard rebase is factory ship path only; triage will not resolve conflicts.

PR #913, opened 2026-07-30
feat(perplexity): migrate synthesis from Sonar to the Agent API

brendo1025:
## Summary

Moves direct Perplexity synthesis from Sonar chat completions to the Agent API (`/v1/agent`), which Perplexity now recommends for all new and existing integrations. Sonar model names stay as the config surface and map to presets, so existing `LAST30DAYS_PERPLEXITY_MODEL` values keep working; `LAST30DAYS_PERPLEXITY_PRESET` is new and selects a preset directly, including `xhigh`. The OpenRouter fallback is untouched and still speaks Sonar.

The measured win is on Deep Research: **~$0.32/query vs ~$0.90 on `sonar-deep-research`** in a live run, with equivalent citation coverage.

### What changed

**Request** — `preset` + `input` replace `model` + `messages`. Date, domain, and recency filters move onto the `web_search` tool's `filters`; `search_context_size` sits on the tool itself; `reasoning_effort` becomes `reasoning.effort`.

**Response** — the typed `output` array replaces `choices`. Two things worth flagging for review, because they are easy to get wrong:

1. The agentic loop emits **one `search_results` item per `web_search` invocation**, not one per response. A live `low` run produced 6 of them (80 results, 62 unique URLs). Reading `output[0]` would silently drop most sources, so results are merged across every item and deduped by URL.
2. The flat top-level `citations` URL array **no longer exists**. Citations come only from `search_results[]`.

**Deep Research** — background runs (`background: true`, poll `GET /v1/agent/{id}`, terminal statuses `completed` / `failed` / `cancelled` / `incomplete`) replace the async Sonar endpoint. The old `_poll_async_sonar` is removed.

**Artifacts** — now carry `preset`, the model the preset actually selected (presets re-tune over time, so this is recorded rather than assumed), and the API-reported `usage.cost.total_cost` as `totalCostUsd`. Endpoint labels are `agent` / `agent-background`.

**Dropped with no Agent API equivalent** — `LAST30DAYS_PERPLEXITY_SEARCH_MODE` (`academic` / `sec`) and `LAST30DAYS_PERPLEXITY_LANGUAGE_FILTER`. Both are now logged and ignored rather than silently dropped, so a configured value does not fail quietly.

## Testing

- [x] `uv run pytest`

`tests/test_perplexity.py` rewritten for the Agent API shapes: 24 tests, all passing. New coverage for preset mapping and override, multi-step `search_results` merge/dedupe, `search_context_size` and `reasoning.effort` placement, dropped-filter handling, cost capture, background-run submit/poll/terminal-on-submit, and `cancelled` / `incomplete` as terminal failures.

Full suite on this machine: **36 failures before and after this change, identical sets** — all pre-existing Windows environment issues (file-mode `0600` assertions, `killpg`, POSIX path checks, `npx` CLI installs), none related to this PR. Verified by diffing the failure sets against pristine `main`.

Live verification against the real API, on top of the mocked tests:

| Path | Result |
|---|---|
| `low` (default) | 81 citations, 81 unique URLs, 81/81 dates inside the requested window, $0.051 |
| `high` background | `queued` → `in_progress` → `completed` over 8 polls, 60 citations, 60/60 in window, $0.355 |

The date-filter behavior was the main risk and was verified explicitly: the docs only show `web_search` filters alongside `model=`, never alongside `preset=`, and state that `tools: []` does not clear a preset's tools. Since this skill is entirely date-window driven, I probed a live `preset: "low"` run with an explicit filtered `web_search` before writing any code — 80/80 results fell inside the window and the request echoed `tools` back verbatim. Preset and explicit filters compose.

## Changelog

- [x] Added `changelog.d/+perplexity-agent-api.changed.md`
- [ ] Skip changelog

No `CHANGELOG.md` edits and no version/manifest bumps.

### AI review

Written with Claude Code. Risks checked, and what came of each:

- **Filters silently ignored under a preset** — the failure mode that would quietly break every date-windowed run while still returning plausible results. Probed live before implementing; confirmed honored. This gated the whole change.
- **Dropping sources from the multi-step loop** — caught by inspecting a real response rather than the docs, which show a single `search_results` item. Drove the merge-and-dedupe implementation and a dedicated test.
- **Cost claims** — not taken from Perplexity's marketing. Measured from `usage.cost.total_cost` on live runs. Worth noting the non-deep path measures ~$0.05, above the $0.03/query figure Perplexity advertises, because date-windowed prompts provoke 6+ search invocations. The Deep Research saving is real and large; the everyday-query saving is smaller than advertised.
- **Regression scope** — verified by diffing full-suite failure sets before and after, not by reading the summary line.
- **Pinning a model in tests** — avoided. The probe ran on `openai/gpt-5.6-luna` and a later run on `openai/gpt-5.6-sol`; presets re-tune, so the artifact records the model rather than asserting one.

### Security

No auth, secrets, path, or subprocess handling changed. Requests still use the existing `http` helper and `Authorization: Bearer` from the already-resolved `PERPLEXITY_API_KEY`; no new dependencies. Tests use dummy keys only. `N/A` otherwise.

## Notes

- Sonar Chat Completions is **not deprecated** — Perplexity's migration guide says it "remains supported" with no announced EOL. This is a cost and accuracy change, not a break-fix, so it can be timed however suits the project.
- **The OpenRouter fallback path is not live-tested** — I have no `OPENROUTER_API_KEY`. Its code path is unchanged and still covered by mocked tests, but a maintainer with a key may want to confirm it end to end.
- `academic` search mode has no clean replacement. The docs suggest approximating it with domain filters plus prompting; I did not attempt that here, preferring an explicit log line over a silent behavior change.

- [x] None
- [ ] Yes — disclosure:

No affiliation with Perplexity. I use this skill and hit the migration notice as a user.

## Related issues

N/A


tmchow:
Not worth it rn.  Sonar is still supported

PR #918, opened 2026-07-31
fix: accept an API root in *_BASE_URL overrides

Opgeek-1:
## Summary

`OPENAI_BASE_URL`, `XAI_BASE_URL`, and `OPENROUTER_BASE_URL` were used as full endpoint URLs, but by the convention every OpenAI-compatible provider documents these name the API root (`https://host/v1`) and the client appends the endpoint path. A value copied from a provider's setup guide POSTed to the API root and failed. `resolve_endpoint` now accepts either form.

## Testing

- [x] `uv run pytest`

Added `ResolveEndpointTests` to `tests/test_providers_v3.py` — 7 cases covering unset, blank, API root, trailing slash, full endpoint URL (the previous behaviour), and the OpenRouter `/chat/completions` path. The 30 existing tests in that file still pass.

## Changelog

- [x] Added `changelog.d/+base-url-api-root.fixed.md` (type: `fixed`)
- [ ] Skip changelog — chore/internal only

### AI review

Written with Claude Code (Opus 5). Risks checked: backwards compatibility for anyone already setting the full endpoint URL (covered by a test), trailing-slash and empty-string handling, and whether these three call sites are the only readers of the variables — `lib/env.py:504` and `lib/permission_preflight.py:12` pass them through without interpreting them, and `last30days.py:2406` re-exports them into `os.environ`, so no other behaviour changes.

### Security

N/A — no change to auth, secrets, command execution, or path handling. These variables were already user-controlled and are still only used as a request URL.

## Notes

The mapping lives in `_ENDPOINT_PATHS`, keyed by the default URL, so a new client needs its endpoint path registered there alongside its default constant.

- [ ] None
- [x] Yes — disclosure: I work at WorldRouter (WorldClaw), an OpenAI-compatible inference provider. This PR does not add, name, or route to WorldRouter — it fixes the generic environment-variable contract. Our users hit this failure, which is how I found it.

## Related issues

N/A

### DIFF
--- a/changelog.d/+base-url-api-root.fixed.md
+++ b/changelog.d/+base-url-api-root.fixed.md
@@ -1,0 +1,1 @@
+`OPENAI_BASE_URL`, `XAI_BASE_URL`, and `OPENROUTER_BASE_URL` now accept an API root (`https://host/v1`) in addition to a full endpoint URL. Values copied from a provider's setup guide previously POSTed to the API root and failed.
--- a/skills/last30days/scripts/lib/providers.py
+++ b/skills/last30days/scripts/lib/providers.py
@@ -26,6 +26,34 @@ OPENROUTER_URL = "https://openrouter.ai/api/v1/chat/completions"
 OPENROUTER_DEFAULT = "google/gemini-3.1-flash-lite-preview"
 
 
+_ENDPOINT_PATHS = {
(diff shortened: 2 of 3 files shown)


mvanhorn:
Speaking as Matt's firstmate:

PR #920, opened 2026-08-01
docs: correct meaning-changing errors in the localized READMEs

tmchow:
## Summary

The five localized READMEs merged in #915 were machine-drafted and reviewed for structure, not for meaning. Several errors change what the docs claim rather than how they read. The product name `last30days` was translated as a date range, so the Grok section tells German readers that installations "take 30 days" and Japanese readers that they are "valid for 30 days". Japanese star counts are corrupted (351K renders as `351, GitHub 0星`) and a stray `44` sits inside a quoted tweet. A dropped negation makes the `--hiring-signals` paragraph claim in Spanish, Portuguese, and Japanese that the report predicts what the roadmap will ship — the opposite of what it does. The repo name is mangled inside @yieldhunter95's quote in all five, citing a repository that does not exist.

This rewrites FR/DE/ES/PT-BR/JA against the English source, preserving the structure the contract tests enforce. Patching line-by-line was not viable: roughly a third of the prose sentences needed touching, and the register problems (German alternating du/Sie, Japanese dropping out of です・ます mid-document) are not fixable in isolation.

Also fixed: recurring calques that read as howlers to a native speaker — papers→newspapers, grounded→electrically earthed, cookie→`Keks`, markdown→`descuento`, store→shop, harness→wiring harness, "hosts"→TV presenters, "run"→game round. And the TweetClaw paragraph, which was garbled in all five because the source line breaks were parsed as sentence boundaries.

One correction the Chinese README already had right: `From your library` stays untranslated. `library_index.py:28` matches that heading literally, so a localized string would not match what the renderer emits.

`README.zh-CN.md` is a separate commit. It predates #915, is in good shape, and needed only three wording fixes (`当地居民` for "locals", not `本地游客`; `执行层/已上线`; `以编辑为先`).

## Testing

- [x] `uv run pytest` — 3,581 passed, 5 skipped, 48 subtests passed

No new tests. `tests/test_readme_translations.py` (added by #915) already locks the structural contract — nav line, code fences, commands, external links, table shape, ordered-list count — and passes on all six files. It cannot catch this class of defect: every error here was structurally valid prose. Verified separately by grep that the mechanical classes are gone (zero markdown-boundary glue, zero stray punctuation spacing, zero product-name translations, zero mangled repo names); remaining hits were false positives (`Diggs` genitive, `CLIs` plural, `los últimos 30 días` as a genuine phrase).

## Changelog

- [x] Skip changelog — documentation only (also add the `skip-changelog` label)

Matches #915's own changelog decision. The guard requires a fragment only for `skills/last30days/scripts/*`, `SKILL.md`, or `mcp/*`; this touches root READMEs only.

### AI review

Claude reviewed all six localized READMEs line by line against `README.md`, classifying findings as mechanical (spacing, untranslated link text), terminological (calques, inconsistent renderings of the same term), or meaning-changing (dropped negations, reversed comparisons, corrupted numbers, translated identifiers). The meaning-changing set drove the decision to rewrite rather than patch. Each correction was verified against the English source; the `From your library` decision was verified against `library_index.py` rather than assumed.

### Security

N/A. Documentation only — no input handling, command execution, path handling, auth, secrets, or dependencies are touched.

## Notes

**This needs a native-speaker pass before merge.** Every correction here is verifiable against the English source, but the author is not a native speaker of these five languages. The marketing-voice sections are where fluent-but-slightly-off phrasing hides best, and the incorrect version is already public on `main`, so there is a real argument for taking the time rather than landing fast on top of it.

Worth considering as follow-up: the errors that mattered most were identifiers being translated (`last30days`, the repo path, ride names, `r/ClaudeCode`). A do-not-translate list committed alongside the READMEs would make the next re-translation regress less, and a test asserting those tokens appear literally in every localized file would catch the whole class mechanically.

- [x] None
- [ ] Yes — disclosure:

## Related issues

Related: #915

https://claude.ai/code/session_016T5XBP3y6AMoFUbkcAuEh9

### DIFF
--- a/README.de.md
+++ b/README.de.md
@@ -16,151 +16,151 @@
   </a>
 </p>
 
-**Eine von KI-Agenten geführte Suchmaschine, bewertet durch Upvotes, Likes und echtes Geld – nicht durch Redakteure.**
+**Eine von einem KI-Agenten gesteuerte Suchmaschine, die nach Upvotes, Likes und echtem Geld gewichtet – nicht nach Redaktionen.**
 
-Dieses README verfolgt die aktuelle v3-Pipeline. Die Runtime-Skill-Spezifikation befindet sich in [skills/last30days/SKILL.md](skills/last30days/SKILL.md), was die Quelle der Wahrheit für das neueste Befehls- und Setup-Verhalten ist.
+Dieses README beschreibt die aktuelle v3-Pipeline. Die Laufzeitspezifikation der Skill liegt in [skills/last30days/SKILL.md](skills/last30days/SKILL.md) und ist maßgeblich für das aktuelle Verhalten von Befehlen und Setup.
 
-**Claude Code (empfohlen — automatische Aktualisierungen über den Marktplatz):**
+**Claude Code (empfohlen – automatische Updates über den Marketplace):**
 ```
 /plugin marketplace add mvanhorn/last30days-skill
 /plugin install last30days
 ```
 
-**Codex, Cursor, Copilot, Gemini CLI, oder einer von 50+ [Agent Skills](https://agentskills.io) Hosts:**
+**Codex, Cursor, Copilot, Gemini CLI oder einer von 50+ [Agent Skills](https://agentskills.io)-Hosts:**
 ```
 npx skills add mvanhorn/last30days-skill -g
 ```
-(`-g` installiert sich global für deinen Nutzer und ist in allen Projekten verfügbar. Setze es in den Umfang pro Projekt.)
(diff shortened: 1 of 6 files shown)


Lackmello2345-creator:
🫷🏾👑🫸🏾 Check there trash can throw wires

Lackmello2345-creator:
🫷🏾👑🫸🏾  Get someone who speaks both languages, test them first to make sure they can only speak those 2 languages. But don’t make it noticeable. 👂🏾

PR #925, opened 2026-08-01
fix(setup): surface ScrapeCreators /profile 5xx instead of already-linked

### LINKED ISSUE #882: ScrapeCreators device flow: GET /v1/github/device/profile returns HTTP 500 after successful GitHub auth; fetch_api_key() masks it as the already-linked case
## Summary

The ScrapeCreators GitHub device flow completes GitHub authorization successfully, then fails at the final step: `GET https://api.scrapecreators.com/v1/github/device/profile` returned **HTTP 500 Internal Server Error** for a freshly-issued, valid access token. No API key is minted or persisted.

Separately, `fetch_api_key()` catches every `HTTPError` in a single `except` (`setup_wizard.py:842`) and returns `None`, collapsing a server-side 5xx into the same path as a 200 that merely lacks an `api_key` field. SKILL.md then has the agent tell the user their "GitHub is probably already linked to a ScrapeCreators account" — which, for a 5xx, is a misdiagnosis that dead-ends the user.

## Steps to Reproduce

Observed on 2026-07-25 by running the bundled `scripts/test_device_auth.py` logic with the `webbrowser.open()` call removed (the machine's default browser did not hold the GitHub session). That harness prints the raw `/profile` result instead of discarding it, which is the only reason the status code was visible:

1. `POST /v1/github/device/code` → **200**, valid `device_code` / `user_code` / `verification_uri`, `expires_in: 899`
2. User enters the code at `github.com/login/device` and authorizes → completes without error
3. `POST /v1/github/device/token` → **200**, returns an `access_token`
4. `GET /v1/github/device/profile` with `Authorization: Bearer <token>` → **HTTP 500 Internal Server Error**

The normal wizard path (`setup --github-start` → `setup --github-poll`) had been run previously on this machine during first-run setup and also produced no persisted key, but its HTTP status is unknown — `fetch_api_key()` masks it. I can only confirm the 500 for the harness run above.

## Expected Behavior

**Upstream:** `/profile` should return an `api_key` for an authorized token, or a clean `4xx` with a machine-readable reason (e.g. `404 no account for this GitHub identity`) — not a 500.

**In this skill:** `fetch_api_key()` should discriminate HTTP status classes rather than flattening them to `None`. A `5xx` is an upstream fault; a `200`-without-`api_key` is the already-linked case the current guidance describes. `poll_device_auth()` already does this — `setup_wizard.py:806` special-cases 400/403/428. `fetch_api_key()` (`:829`, sole call site `:1052`) does not.

## Error

```
AUTHORIZED. Fetching /profile ...
PROFILE FETCH FAILED: HTTP Error 500: Internal Server Error
```

The response body was **not** captured — the harness lets `urlopen` raise and does not read `HTTPError.read()`. I can re-run to capture it if useful.

## Additional context — the endpoint is not globally down

Unauthenticated read-only probes against the same host, run just now:

```
GET /v1/github/device/profile                    -> 401
GET /v1/reddit/search                            -> 401
GET /v1/github/device/profile  (invalid bearer)  -> 401
   body: {"message":"Bad credentials",
          "documentation_url":"https://docs.github.com/rest",
          "status":"401"}
```

Two things follow:

1. `/profile` is healthy for the bad-token case — it returns a well-formed 401. So the 500 is localized to whatever runs **after** GitHub auth succeeds: presumably the ScrapeCreators account lookup/creation for a valid GitHub identity. That is a much narrower target than "the endpoint is broken."
2. The 401 body is GitHub's own REST error format (`documentation_url: https://docs.github.com/rest`), which indicates `/profile` forwards the bearer token to GitHub's API and returns GitHub's raw response. This is an independent, reproducible confirmation of the concern raised in #663 Finding B.

Corroborating but not conclusive: signing in at `scrapecreators.com` with the same GitHub account reports **"user not found"**, consistent with no account record existing on their side and the handler faulting on the missing row rather than returning a clean error.

## Impact on the user-facing path

SKILL.md's handling for this error instructs the agent to say: *"GitHub authorized, but I couldn't auto-grab your ScrapeCreators key - your GitHub is probably already linked to an account. Get your key at scrapecreators.com and paste it here."* When the cause is a 5xx with no account, that instruction cannot succeed — the user signs in, gets "user not found", and has no path forward.

There is also no alternative route to a key. Every `setup --github*` entry point funnels into the same device flow and terminates at the single `fetch_api_key()` call site (`:1052`); `run_github_auth()` (`:1082`) is a back-compat wrapper delegating to `run_full_device_auth()`, and its docstring states *"This path must not read or forward local GitHub CLI tokens"* — so the `gh auth token` PAT path described in #401 is deliberately absent, and I am **not** suggesting it be restored given #663. Meanwhile `doctor` reports tiktok, instagram, threads, linkedin, and pinterest as `unconfigured` with the prescription `setup --github`, which cannot currently succeed on this machine.

## Suggested fixes

1. Split the `except` in `fetch_api_key()` (`setup_wizard.py:842`) by status class: report `5xx` as an upstream fault distinct from the already-linked case, retry it a bounded number of times, and include `HTTPError.read()` in the logged/returned detail so the failure is diagnosable without a custom harness.
2. When `/profile` returns `5xx`, have both the SKILL.md guidance and the `doctor` prescription route the user to web signup rather than "paste your key from scrapecreators.com" — the latter presumes an account that may not exist.
3. Consider raising the 500 with ScrapeCreators directly, since the localization above suggests it is their account-lookup path rather than anything this skill controls.

## Environment

last30days v3.18.0 (Claude Code plugin, versioned cache install) · macOS Darwin 25.5.0 · Python 3.14.6 · `gh` 2.87.2 authenticated · config source `global:~/.config/last30days/.env`, `SETUP_COMPLETE=true`

**Impact note (corrected).** Per @mvanhorn's announcement, the GitHub device-flow path grants last30days users a partner allocation of **10,000 free credits**, materially larger than the public web signup. That makes this 500 costlier than a routine setup hiccup: the blocked path is the only route to the partner allocation, and the fallback the current guidance offers ("get your key at scrapecreators.com") lands the user on the smaller public tier even when it succeeds. Worth weighing when triaging severity.

Minor related data point: that announcement describes the public signup as 1,000 credits, while scrapecreators.com currently advertises 100 free credits and #367 documented 100. Not central to this bug, but the three figures in circulation may be worth reconciling in the docs.

arimu1:
## Summary

`fetch_api_key()` previously caught every `HTTPError` and returned the same failure shape as a successful `/profile` response with no `api_key`. That made a ScrapeCreators **HTTP 500** look like the already-linked case, so SKILL.md told users their GitHub was "probably already linked" and to paste a key that may not exist.

This PR:

- Returns a structured result from `fetch_api_key` (`ok` / `no_api_key` / `upstream_error` / `http_error` / `request_failed`)
- Retries 5xx a bounded number of times and includes a truncated response body in `detail`
- Emits a distinct poll error message for upstream failures (`Authorized but ScrapeCreators profile failed (HTTP …)`) with `reason: upstream_error`
- Keeps the historical `Authorized but failed to fetch API key` message only for the 2xx-without-key path
- Updates SKILL.md modal + prose guidance so agents do **not** misdiagnose server errors as already-linked

Fixes #882

## Testing

- [x] `uv run pytest` (focused): `TestFetchApiKey`, `TestRunFullDeviceAuth`, `test_github_device_auth`, `test_onboarding_contract` — all green
  - New: 5xx → `upstream_error` with retries; poll path distinct message; onboarding contract for SKILL.md branch

## Changelog

- [x] Added `changelog.d/882.fixed.md`

### AI review

Self-review of the device-auth path: call sites of `fetch_api_key` (only `run_github_poll`), mock return shapes in tests, SKILL.md contract tests, and that the historical already-linked message string remains exact for the no-key case. Confirmed no open PR already targeting #882; #126/#317 are unrelated historical merges.

### Security

`/profile` error bodies are truncated and logged as diagnostic detail only (no bearer token logging). Field-name-only logging on the no-key path is unchanged. No new network destinations, secrets, or shell execution.

## Notes

Does not fix the upstream ScrapeCreators 500 itself — only stops masking it as already-linked and improves agent recovery guidance.

- [x] None
- [ ] Yes — disclosure:

## Related issues

Fixes #882


mvanhorn:
Speaking as Matt's firstmate:

PR #929, opened 2026-08-02
feat: add Dia to macOS Chromium-family cookie extraction

arkadevbanerjee:
## Summary

Adds **Dia** (The Browser Company) to the macOS Chromium-family cookie extraction, alongside Chrome/Brave/Edge/Vivaldi/Opera/Arc/Chromium. Dia uses the same `<name>/User Data` profile layout as Arc and the standard `Dia Safe Storage` Keychain entry, so it needs no new decryption logic — it routes through the existing shared v10 AES-128-CBC core.

Motivated by a real switch: with Chrome uninstalled and Dia as the default browser, X/Twitter auth had no working cookie source, and the failure surfaced as `X/Twitter errored - log into x.com in your browser` even though the user *was* logged in.

Registering a browser touches more places than the registry alone, so this covers all of them:

- `lib/chrome_cookies.py` — `CHROMIUM_BROWSER_PROFILES` entry (decryption)
- `lib/cookie_extract.py` — `extract_dia_cookies()`, the `extractors` dispatch dict, and the `auto` order list
- `lib/env.py` / `last30days.py` — the `chromium_browsers` allow-lists that gate `FROM_BROWSER`
- Docstrings and `CONFIGURATION.md` browser lists

Missing just the dispatch dict is quietly bad: `FROM_BROWSER=dia` logs `Unknown browser: dia` at warning level and then reports the generic "log into x.com" message, which blames the user for a registry gap. The new routing test covers exactly that.

Usage: `FROM_BROWSER=dia`, or `FROM_BROWSER=auto` picks it up.

## Testing

- [x] `uv run pytest` — 3583 passed, 6 skipped, 48 subtests passed

Test changes: `dia` added to `NEW_CHROMIUM_BROWSERS`, both `TestCookieExtractRouting` parametrizations, and the exact-set assertion in `test_registry_has_expected_browsers`. Also added an `extract_dia_cookies` mock to the two `auto`-order tests — without it those tests stop being hermetic on any machine with Dia installed and logged in, since `auto` reaches Dia before Firefox.

Verified the new tests actually fail for the right reason: removing only the `"dia": extract_dia_cookies` dispatch entry (leaving the registry intact) fails `test_routes_to_registry[dia]` with `KeyError: 'dia'` and `test_auto_macos_order_includes_chromium_family`.

Verified end-to-end against a real Dia install on macOS 15: `.x.com` `auth_token`/`ct0` decrypted from a logged-in profile via the `Dia Safe Storage` Keychain entry, and a live `--search x` run returned real items.

Note the Keychain entry requires a one-time user approval on first read, same as every other browser here.

## Changelog

Added `changelog.d/+dia-browser-cookies.added.md`.


arkadevbanerjee:
Good catch — valid, and it was the one place I missed. Fixed in f110570.

`run_auto_setup()` keeps its own hardcoded chromium-first list rather than deriving one from `cookie_extraction_browsers()`, and that list is what runs when `FROM_BROWSER` is unset — i.e. exactly the first-run case this PR is motivated by. Runtime support with the wizard blind to Dia would have reported "no browser cookies found" to the only user who needs it.

Added `dia` to that list, plus `test_default_scan_includes_dia`. Verified it fails for the right reason before trusting it — with `dia` removed from the list again:

```
tests/test_setup_wizard.py:178: in test_default_scan_includes_dia
    assert "dia" in tried
E   AssertionError: assert 'dia' in ['chrome', 'brave', 'edge', 'vivaldi', 'arc', 'chromium', ...]
```

Full suite now `3584 passed, 6 skipped, 48 subtests passed`.

One thing I noticed but deliberately left alone, since it predates this PR: that same wizard list also omits `opera`, which *is* in `cookie_extraction_browsers()`. Same drift, same symptom for Opera users on first run. Happy to fix it here or leave it as a separate issue — your call. The durable fix is probably deriving the wizard's scan order from the registry instead of hardcoding it twice, but that's a bigger change than this PR should carry.

mvanhorn:
Speaking as Matt's firstmate:
Classified **opt-in** (Dia in macOS Chromium-family cookie extraction + auto/setup routing). Greptile **5/5**, mergeable CLEAN, CI green on head. Proposing merge for Firstmate/captain OK — does not change behavior unless Dia is present / `FROM_BROWSER=dia|auto`. STOP before merge per standing order.

PR #939, opened 2026-08-02
LinkedIn source: honor the date window, follow the cursor, and signal partial results

jasonericsiegel:
## What I hit

Running the LinkedIn source over a 90-day window returned 10 posts spanning 26 days. Digging in, `--days` never reaches the API at all.

`search_linkedin()` accepts `from_date`/`to_date` and never reads either one. The window comes entirely from `DEPTH_CONFIG`, where `deep` is hardcoded to `last-month`. The docstring says `from_date` is "used for depth mapping", but nothing maps it.

## Three defects, all confirmed against the live endpoint

**1. The date window is ignored.** Any `--days` value produces the same bucket.

**2. The 10-post cap is ours, not the API's.** The response carries a `cursor` field the code never reads, so it takes page 1 and stops. `deep` asks for 30 and can never get more than ~10. Walking the cursor on the same query returned 28 unique posts.

**3. Queries over ~80 characters return HTTP 404.** Verified: 108 chars fails, 79 passes. The `except` block catches it and the engine silently falls back to a generic search, so a targeted query can be replaced by a broad one with no visible error.

I also probed the accepted `date_posted` values, since they aren't documented in the module — the API's own 400 message lists them: `last-hour, last-day, last-week, last-month, last-year`. There is no 90-day bucket, which matters for the fix below.

## The measurement that shaped the fix

Results inside a `date_posted` bucket are **relevance-ranked, not recency-ranked**. Querying `last-year` for a 90-day window returned 30 posts spread across a full year, only **4** of them inside the window.

So simply widening the bucket makes things worse, not better — a wide bucket does not contain the narrow one's yield. This PR queries the narrow bucket *and* the covering bucket and unions them. `normalize.filter_by_date_range()` already hard-filters to the caller's exact window downstream, so the buckets only need to cover it, not match it.

One more consequence: the bucket has to be sized by how **old** the window is (`today - from_date`), not by its **span**. Sizing by span sends a 31-day window from two years ago to `last-month`, which cannot contain a single matching post.

## Also in this PR

These came out of reviewing the fix above, and are all consequences of turning one call into a loop:

- `max_results` bounds the returned collection again, via a per-bucket quota plus a global cap. Without it a 2-bucket run returned 2x the configured limit.
- A bucket that fails after another succeeded no longer reads as complete — the return carries `partial: True` plus the first error. Previously a silently-dropped window looked like a real finding about LinkedIn having little discussion.
- `401/403/429` abort the whole search rather than immediately retrying the next bucket with the same credential. Retrying a 429 only deepens the rate limit.
- A whole-call wall-clock budget. `MAX_PAGES` bounds the request *count* but not *time*; a slow-but-alive endpoint could stretch a single call to roughly 10 minutes with no error to attribute the stall to.
- Dedupe reads `postUrl`/`post_url`/`postId` as well, and falls back to a content fingerprint. Keyless posts were never deduped, which also meant a repeating cursor never tripped the all-duplicates guard.
- An all-duplicate page no longer ends a bucket immediately. The covering bucket's first page routinely repeats the narrow bucket's top hits, so breaking on the first zero-add page skipped exactly the older tail that second bucket exists to fetch.
- Per-bucket logging records post count, page count, and which exit condition fired. There are six distinct ways a bucket can stop and they were previously indistinguishable in the logs.

## Effect

One query, 90-day window: **9 usable posts before, 20 in-window after**, with the window actually spanning 90 days rather than 26.

## Testing

Verified `_buckets_for_window` by hand across spans of -5, 0, 1, 2, 7, 8, 31, 32, 366 and 400 days, plus unparseable, empty, inverted, and full-ISO-timestamp inputs. Verified the loop behavior (duplicate first page, global cap, partial failure, fatal-status abort) against a stubbed API, and ran a live smoke test against the real endpoint.

Happy to adjust anything here — in particular, I picked 80 chars for `MAX_QUERY_CHARS` and 120s for the budget empirically, and you may have better values.


jasonesiegel:
Two corrections pushed, one of them to my own diagnosis.

**The 404 was never a length limit.** I originally read the 404 on a long query as an upstream length ceiling and added an 80-character trim. That was wrong, and I should have probed it before writing the fix. The endpoint returns `404` with `{"message": "No posts found"}` for any query that matched nothing — a nonsense query 404s at 49 characters, and a real query succeeds at 120. I verified lengths 80 through 120 (all 200) and nonsense queries at 49 and 72 (both 404).

So the trim solved nothing and would have silently narrowed any topic over 80 characters. It's removed. A 404 now ends that window as an empty result instead of being recorded as an error, which also stops a legitimate null from being reported as a partial failure — arguably the more valuable half of the correction, since the old behavior turned "nobody posted about this" into what looked like an outage.

**The wall-clock budget was advertised but not enforced** (thanks @greptile-apps). The deadline was only checked between requests, so a page starting just before it could still consume its full timeout-times-retries budget and overrun by ~90 seconds. Each request is now clamped to the remaining budget, and the retry is dropped when there isn't room for one.

The window, pagination, cap, partial-signalling, dedupe, and fatal-status changes from the first commit are unchanged.

grahamgourlay-prog:
Cross-referencing #1086, which corrects defect 3 in this issue.

The 404 is not a query-length limit — it tracks **result count**. Same key, same `date_posted=last-month`:

| chars | query | result |
|---|---|---|
| 31 | `Twenty7tec Mortgage Brain Iress` | **404** |
| 39 | `zxqwvbn qqjjkl nonexistent phrase 12345` | **404** |
| 90 | `the mortgage broker industry and mortgage brokers working across the mortgage market today` | **200**, 20 posts |
| 92 | `mortgage broker mortgage brokers mortgage industry mortgage market mortgage lending mortgage` | **200**, 20 posts |

A 31-char query 404s; a 92-char query returns 20 posts. The controlled case is the nonsense query — 39 chars, syntactically valid, guaranteed no matches, 404. Narrowing confirms it tracks specificity rather than length: `Twenty7tec` alone returns 20 posts, `Twenty7tec Mortgage Brain Iress` returns 404.

The 108-fails/79-passes pair here is consistent with those two queries differing in result count rather than length.

Flagging it because it changes the fix: a length-based guard (truncating or splitting long queries) would not address the cause, and would degrade long queries that currently work. #1086 proposes treating 404 as an empty result in `search_linkedin()` instead, and documents a `MAX_SOURCE_FETCHES['linkedin']=1` interaction that makes this silently zero the source on normal multi-term topics.

The date-window and cursor defects reported here are unaffected — those still stand.

grahamgourlay-prog:
Follow-up to my comment above — I should have read this diff before commenting, and my framing was unfair to it.

This PR **already treats 404 as an empty result** (`EMPTY_RESULT_STATUS = 404`, with the `{"message": "No posts found"}` note), which is exactly the right handling. My earlier comment implied the length theory drove the fix; the code plainly does not depend on it. Apologies for the noise.

The measurement still stands and bears on one part of the change: since the 404 tracks result count rather than query length (31 chars 404s; 92 chars returns 20 posts), the query-shortening step may be addressing a symptom that does not exist as described — and could truncate long queries that currently work. Non-blocking, and it may be worth keeping as defensive anyway.

The date-window, cursor-walking, and `partial: True` work here is the substantive fix and I have no quibble with any of it.

mvanhorn:
Speaking as Matt's firstmate:

PR #947, opened 2026-08-05
fix(arxiv): preserve adapter-valid papers during normalization

### LINKED ISSUE #946: arXiv source reports '0 papers' despite 'found 10 entries' — normalize.py 30-day window drops all papers (different layer from #908)
## Summary

The arXiv source returns `0 papers` in every report even when the adapter logs `[arXiv] found 10 entries`. Papers are parsed and survive the adapter's own 365-day recency gate, then get silently dropped by a second, stricter date filter in `normalize_source_items`.

This is a **different layer** from #908 (query construction). #908 is about the exact-phrase query returning zero entries; this bug drops entries that **were** returned and parsed.

## Environment
- Version: 3.18.4
- Host: Hermes agent on Linux (Ubuntu 22.04)
- CLI: `arxiv-pp-cli` 2026.7.2 (`doctor`: Config ok / Auth not required / API reachable)

## Steps to Reproduce
Run a normal (non-`--quick`) research pass:

```console
$ last30days.py --emit context "video generation"
[arXiv] query video generation (relevance, max=10)
[arXiv] found 10 entries
✓ Research complete ... arXiv: 0 papers
```

`last-report.json` shows `arxiv: state=no-results items_returned=0`, and `errors_by_source` has no entry for arxiv — no error, just zero items.

Direct adapter call returns items:

```python
>>> arxiv.parse_arxiv_response(arxiv.search_arxiv("video generation", "2026-07-06", "2026-08-05"))
3 items  # e.g. "Human Motion Video Generation: A Survey" (2025-09-04)
```

But `pipeline.run(..., requested_sources=["arxiv"])` reports `items_by_source["arxiv"] == []`.

## Root Cause

Two layers disagree on the arXiv recency window:

1. `scripts/lib/arxiv.py` — `RECENCY_DAYS = 365`; the adapter's parse layer intentionally allows papers up to one year old (research does not trend on a 30-day clock).
2. `scripts/lib/normalize.py` — `normalize_source_items()` calls `filter_by_date_range(normalized, from_date, to_date)` with the **30-day community window** for every source. Every arXiv paper older than `from_date` is dropped here.

The `evergreen_ok` escape hatch only covers `youtube`:

```python
if filtered:
    return filtered
if freshness_mode == "evergreen_ok" and source == "youtube":
    ...
return filtered   # arxiv falls through here -> always empty
```

So the adapter's 365-day contract is nullified by the 30-day filter one layer up. arXiv effectively never contributes, while the logs misleadingly show entries were found.

## Suggested Fix

Give `arxiv` the same exemption as `jobs` in `normalize_source_items` — the adapter's own `RECENCY_DAYS` gate is the authoritative recency contract for that source:

```python
if source == "arxiv":
    # arXiv papers do not trend on a 30-day clock; the adapter parse layer
    # already enforces RECENCY_DAYS=365. Re-applying the 30-day community
    # window here silently drops every paper published before from_date.
    return normalized
```

With this change the same run reports `arXiv: 3 papers`.

## Impact
- arXiv source is effectively dead in default (30-day) runs for any topic whose top relevance-ranked papers are older than 30 days — which is most research topics, since arXiv's relevance sort favors established work.
- `--quick` runs are unaffected only because they skip arXiv entirely (not in the quick source set).

dpersek:
## Summary

Preserve arXiv papers that already passed the adapter's 365-day recency gate instead of applying the report's narrower date window a second time. Add normalization-seam regression coverage and a release-note fragment for the fix.

## Testing

- [x] `uv run pytest`

Focused proof passed: `uv run pytest tests/test_arxiv_source.py tests/test_normalize_v3.py` reported 37 passed, and the new arXiv regression first failed before the production change and passed afterward. The final full-suite run reported 3581 passed, 5 skipped, 48 subtests passed, and one unrelated pre-existing failure in `tests/test_footer_nudge_suppression.py::FooterNudgeSuppressionTests::test_bare_run_emits_web_promo`; the same exact test also fails on an untouched detached `upstream/main` worktree at `52f5331`.

## Changelog

If this change should appear in the next release notes, add a fragment under `changelog.d/` (see `changelog.d/README.md` and [CONTRIBUTING.md](../CONTRIBUTING.md)). Do **not** edit `CHANGELOG.md` or bump version/manifest files in this PR.

### AI review

Independent review checked the source-specific date-policy boundary, neighboring source behavior, regression coverage, changelog compliance, and diff hygiene. A final source-aware review found no actionable defects; no review-driven code changes were needed.

### Security

Reviewed date and metadata handling plus the changed control-flow boundary. The patch adds no command execution, path handling, authentication, secret access, dependency, or external-I/O behavior; no security follow-up is needed.

## Notes

Deterministic proof covers the reproduced adapter-to-normalization drop. A disposable live `arxiv-pp-cli` 2026.7.2 run also found 10 entries for `video generation`; 3 passed the adapter's 365-day policy and all 3 reached the final arXiv report source with state `ok` and no source error. The old shared 30-day filter would have retained none of those 3.

Disclose employment, contracting, equity, or other paid ties to a company/product/service this PR adds or meaningfully promotes (example: you work at the API vendor being integrated).

- [x] None
- [ ] Yes — disclosure: 

## Related issues

Fixes #946

### DIFF
--- a/changelog.d/946.fixed.md
+++ b/changelog.d/946.fixed.md
@@ -1,0 +1,1 @@
+Adapter-valid arXiv papers are no longer dropped by the report date window during normalization.
--- a/skills/last30days/scripts/lib/normalize.py
+++ b/skills/last30days/scripts/lib/normalize.py
@@ -79,6 +79,10 @@ def normalize_source_items(
         normalizer(source, item, index, from_date, to_date)
         for index, item in enumerate(items)
     ]
+    if source == "arxiv":
+        # The adapter owns arXiv's 365-day recency contract. Applying the
+        # report window again here drops relevant papers the adapter accepted.
+        return normalized
     if source == "jobs":
         # A careers board is a snapshot of CURRENTLY OPEN roles. An open posting
         # is current evidence regardless of when it was posted, so date-windowing
--- a/tests/test_normalize_v3.py
+++ b/tests/test_normalize_v3.py
@@ -45,6 +45,32 @@ class NormalizeV3Tests(unittest.TestCase):
         )
         self.assertEqual([], normalized)
 
+    def test_arxiv_keeps_adapter_valid_items_older_than_report_window(self):
+        items = [
+            {
+                "id": "http://arxiv.org/abs/2509.00001v1",
+                "title": "Reliable Agent Memory",
+                "url": "https://arxiv.org/abs/2509.00001v1",
+                "summary": "A study of memory systems for coding agents.",
+                "author": "Ada Lovelace",
+                "authors": ["Ada Lovelace"],
+                "date": "2025-09-04",
+                "relevance": 0.9,
+            }
+        ]
+
+        normalized = normalize.normalize_source_items(
+            "arxiv",
+            items,
+            "2026-07-06",
+            "2026-08-05",
+        )
+
+        self.assertEqual(1, len(normalized))
+        self.assertEqual("Reliable Agent Memory", normalized[0].title)
+        self.assertEqual("https://arxiv.org/abs/2509.00001v1", normalized[0].url)
+        self.assertEqual("2025-09-04", normalized[0].published_at)
+
     def test_youtube_top_comments_passthrough_with_field_mapping(self):
         """YT comments from enrich_with_comments use likes/text; normalize must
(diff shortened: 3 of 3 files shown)


mvanhorn:
Speaking as Matt's firstmate:

PR #958, opened 2026-08-08
feat(hooks): add LAST30DAYS_QUIET to quiet the SessionStart status hook

dnorth123:
## Summary

The SessionStart status hook has no off switch, so an install that deliberately settles on the free sources is advised to run the setup wizard at every session start, in every project, indefinitely. This adds `LAST30DAYS_QUIET`: when truthy it drops the first-run welcome and the ScrapeCreators tip while keeping the one-line status and the last-run summary. Unset or falsy leaves output byte-for-byte as it is today.

## Testing

- [x] `uv run pytest` (3606 passed, 6 skipped)

`tests/test_check_config_quiet.py` adds 9 cases. Each asserts both directions, comparing a quiet run against an otherwise-identical non-quiet run, so no test can pass merely because the hook printed nothing.

The tests were calibrated against the unmodified hook before the change was kept: the expected failure set was predicted in advance as 4 of the 9, and the run produced exactly those 4 (`suppresses_new_user_welcome`, `keeps_status_and_drops_scrapecreators_tip`, `honored_from_config_file`, `preserves_last_run_line_when_unconfigured`), with the 5 falsy-value cases passing in both states as intended.

## Changelog

- [x] Added `changelog.d/+quiet-session-start-hook.added.md` (type: `added`)
- [ ] Skip changelog

No version or manifest file is touched.

## Agent disclosure

This PR was written by Claude Code (Opus 5) acting on my behalf. I reviewed the diff and the verification below before opening it.

### AI review

Risks checked, and what came of each:

- **Does the flag actually work, or do the tests only look green?** Ran the new tests against the unmodified hook and predicted the failure set first. Result matched the prediction exactly, so the tests discriminate rather than passing vacuously.
- **Can an untrusted project config silence the hook?** No. `is_quiet` reads `ENV_LAST30DAYS_QUIET`, which `load_env_vars` only populates after the existing `project_config_trusted` gate, so an untrusted repo's `.claude/last30days.env` cannot reach it. This is the same property the trust gate already provides for every other setting.
- **Do the early exits skip anything that matters?** Both guards sit after config loading, so `check_perms` and the `LAST30DAYS_MEMORY_DIR` `mkdir` still run. Verified by running the hook with `LAST30DAYS_QUIET=1` against a `0644` config file: the permissions warning and the auto-`chmod 600` still fire, followed by the status line and no upsell.
- **Regressions elsewhere?** Full suite green. The three existing `check_config` test modules pass unchanged.

### Security

No new input parsing, command execution, path handling, auth, or secrets. `LAST30DAYS_QUIET` flows through the existing `trim_ws` / `strip_outer_quotes` / `is_truthy` path and, when it comes from a file, through the existing `^[A-Za-z_][A-Za-z0-9_]*$` key validation in `load_env_vars` that guards against the array-subscript injection noted in that function's comment. No new dependencies. No follow-up needed.

## Notes

The contract is deliberately narrow: quiet trims onboarding and upsell copy, it does not silence the hook. A configured install still prints its status line, and the last-run summary survives on both paths, including the unconfigured one, so a returning user keeps the only line there that carries state.

Nothing is host-specific. Behavior is identical wherever the hook already runs.

- [x] None
- [ ] Yes

No ties to ScrapeCreators or any other provider referenced by the copy this flag suppresses.

## Related issues

N/A. I searched open and closed issues for an existing report of this before writing the change and did not find one.


mvanhorn:
Speaking as Matt's firstmate:

Captain OK via Firstmate: merging #958 as opt-in. CI green, Greptile 5/5 on current head.

Issue #899, opened 2026-07-28
Reddit 429/403 transport failures can be reported as no-results/success


Frosti7:
## Bug Description

`last30days` can report Reddit as healthy/no-results even when live Reddit public endpoints return HTTP 429/403 from the host. This makes `doctor` / `postmortem` overstate Reddit reliability and hides a real transport failure as an empty result set.

## Observed Environment

- Skill/engine: `last30days v3.18.3`
- Host: Linux VPS / datacenter egress
- Config: no Reddit paid/keyed backfill (`SCRAPECREATORS_API_KEY` absent), browser cookies off
- Source path: public keyless Reddit composite

## Repro

Run:

```bash
python3.12 scripts/last30days.py doctor
python3.12 scripts/last30days.py "Claude Code reddit user feedback" \
  --max-results 25 --max-per-source 15 --emit json --debug \
  --save-dir /tmp/last30days-reddit-report
python3.12 scripts/last30days.py doctor --postmortem
```

Direct endpoint probes from the same host observed:

| Endpoint | Result |
|---|---|
| `reddit.com/search.rss` | 429 |
| `reddit.com/search.json` | 403 |
| `old.reddit.com/search` | 403 |
| `reddit.com/r/all/hot.json` | 403 |
| known thread RSS | 429 |
| Shreddit comments | 403 |
| Arctic Shift comments API | 200 |

## Actual Behavior

Debug logs show Reddit transport failures:

```text
[RedditKeyless] Tier 1 (RSS) 0 posts
HTTP Error 429: Too Many Requests
HTTP Error 403: Blocked
```

But exported/postmortem status can say:

```text
"source_status": { "reddit": "no-results" }
Succeeded: reddit (0)
No failures on the last run.
```

## Expected Behavior

If Reddit discovery/enrichment hits 429/403, preserve it as degraded/failure state (`partial`, `rate-limited`, or `blocked`), not `no-results`. `doctor --postmortem` should not report success/no failures when the source hit transport failures.

## Suggested Fix

- Preserve Reddit adapter HTTP 429/403 failures in `SourceOutcome.detail` and state.
- Treat discovery transport failures as `partial`/`rate-limited`/`blocked` even if fallback lanes return zero items.
- Make `postmortem` classify `no-results` as success only when the source completed without transport failures.
- Consider adding a live Reddit discovery/enrichment probe to doctor.

eugenewong22:
hi i'd like to work on this

PR #960, opened 2026-08-08
feat(reddit): arctic-shift fallback for shreddit listing lanes

technicianofthesacred:
## Summary

On datacenter egress, Reddit 403s the keyless `/svc/shreddit` listing partials and soft-blocks `search.rss` (HTTP 200 + empty body), so the keyless Reddit pipeline returned zero items and reported `auth-failed` on those hosts. This PR adds a fallback to the **arctic-shift archive** (https://arctic-shift.photon-reddit.com — public Reddit archive, keyless, IP-agnostic, already used for score backfill in `reddit_arctic.fetch_scores`) as the scored listing lane:

- `reddit_arctic.fetch_listings()` — new scored listing lane mirroring `reddit_listing`'s post schema (real `score`/`num_comments`, `YYYY-MM-DD` dates, dedupe, best-effort never-raises)
- `reddit_keyless._scored_listings()` — shreddit-first, arctic-shift fallback for the dedicated / listing / score-only lanes; fallback failure is caught so it can never break the pipeline
- `reddit_listing.fetch_discovery_listings()` — falls back to arctic when every shreddit partial fails, and clears the error bucket when the fallback recovers the source (keeps `--discover` mode honest)

Verified live on a VPS: `"matcha tea trends" --days=30 --search=reddit` went from 0 items / `auth-failed` to **5 threads, 205 upvotes, 266 comments** (r/tea, r/MatchaEverything), and `"ai agents"` across reddit+digg+arxiv+techmeme+hn+github returned all six sources clean.

## Testing

- [x] `uv run pytest` (full suite; coverage 88.41% vs the 84% gate — the one failure, `test_setup_wizard.py::TestWriteApiKey::test_unwritable_target_returns_false`, also fails on clean `main` here because the suite runs as root, which bypasses the test's read-only permission setup; unrelated to this change)
- [x] Added tests:
  - `test_reddit_arctic.py::TestFetchListings` — normalization, real scores, `t3_` strip, depth volume, dedupe, `r/`-prefix strip, rate-limit/network degradation, empty input
  - `test_reddit_keyless.py::TestScoredListingsFallback` — fallback fires on empty shreddit results, skipped when shreddit delivers, never raises, sorts pass-through
  - `test_reddit_listing.py::TestFetchDiscoveryListingsFallback` — recovery clears errors, errors preserved when arctic is also empty, no arctic call when shreddit succeeds
  - Updated `test_keyless_path_runs_rss_and_listings` to mock the fallback lane

## Changelog

- [x] Added `changelog.d/+reddit-arctic-listing-fallback.added.md`

### AI review

Reviewed against the shreddit-lane semantics in #899/#900 (transport failures must surface when nothing is recovered): `fetch_discovery_listings` clears `errors` **only** when the arctic fallback actually returns items, so an unrecovered block still reaches the doctor. The fallback is best-effort and never raises in either module; `_scored_listings` additionally catches fallback exceptions so a broken arctic-shift can only degrade discovery, never crash the run. Arctic-shift rows are normalized to the exact `reddit_listing.parse_cards` schema so downstream dedupe/ranking/relevance paths are unchanged.

### Security

No new command execution or auth paths — one new HTTPS GET to a public read-only API the skill already calls (score backfill). No secrets involved. The arctic-shift endpoint is third-party infra; degradation behavior is covered by tests (`{"error": ...}` and network-failure cases return empty, never raise).

## Notes

Reddit's own endpoints remain blocked on such hosts (nothing this PR can change); arctic-shift listings are recency-sorted with real scores rather than top/hot/new lanes, so the caller's engagement ranking does the ordering (the fallback lane documents this). Comment-text enrichment via shreddit still degrades silently on blocked hosts — follow-up candidate for an arctic-shift comments lane.

- [x] None

## Related issues

Companion to `fix/source-outcome-swallowed-lane-failures` (outcome honesty when lanes 403 while items are delivered).

### DIFF
--- a/changelog.d/+reddit-arctic-listing-fallback.added.md
+++ b/changelog.d/+reddit-arctic-listing-fallback.added.md
@@ -1,0 +1,1 @@
+Reddit keyless discovery now falls back to the arctic-shift archive when the shreddit listing partials return nothing — hosts on datacenter egress (where Reddit 403s `/svc/shreddit`) keep scored Reddit discovery, score backfill, and discover-mode listings instead of reporting `auth-failed`.
--- a/skills/last30days/scripts/lib/reddit_arctic.py
+++ b/skills/last30days/scripts/lib/reddit_arctic.py
@@ -16,16 +16,22 @@ count rather than failing the Reddit source.
 
 import sys
 import time
-from typing import Dict, List
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
 
 from . import http
 
 API = "https://arctic-shift.photon-reddit.com/api/posts/ids"
+SEARCH_API = "https://arctic-shift.photon-reddit.com/api/posts/search"
 BATCH = 50          # ids per request
 TIMEOUT = 15
 MAX_BATCHES = 3     # cap total requests per run (bounds latency + rate-limit risk)
 PACE_SECONDS = 0.4  # gap between batches; arctic-shift answers 422 "slow down"
 CACHE_MAX = 4096    # hard size bound so the in-run memo can never grow unbounded
+# Listing-lane knobs (mirror reddit_listing's DEPTH_LIMITS so callers get the
+# same per-depth volume from either backend).
+_LISTING_DEPTH_LIMITS = {"quick": 10, "default": 25, "deep": 50}
+_MAX_LISTING_SUBS = 8  # cap concurrent subreddits per listing call
 # In-run memo: base36 id -> {score, num_comments}. Module-level so repeated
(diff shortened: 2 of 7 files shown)


mvanhorn:
Thanks for this PR! We cherry-picked your Arctic listing fallback into #992, plus follow-ups so partial Shreddit success still falls back for failed subs/sorts and so recovered errors are attributed by exact lane, not prefix. Really appreciate the contribution.

Closing this in favor of #992. That replacement is CI-green and Greptile 5/5; it is not on main yet.

PR #961, opened 2026-08-08
feat(reddit): arctic-shift fallback for shreddit comment enrichment

technicianofthesacred:
## Summary

On datacenter egress, Reddit 403s the keyless `/svc/shreddit/comments` partials, so comment enrichment silently degraded to counts only — briefs carried upvote/comment *counts* but no comment *text*. This PR adds a fallback to the **arctic-shift archive** (`/api/comments/search`, keyless, IP-agnostic — same archive as the existing score backfill):

- `reddit_arctic.fetch_post_comments()` — a post's comments normalized to `reddit_shreddit.parse_comments` shape (score/author/body/excerpt/permalink/date/url), ranked by score client-side (the archive's `sort_type` only supports `created_utc`), best-effort never-raises
- `reddit_shreddit.fetch_comments()` — falls back to arctic when the shreddit partial returns nothing; the fallback call is guarded so enrichment can never break. `num_comments` stays `None` on the fallback (archive rows carry no post total — discovery already holds the count)

Verified live from a datacenter host: `fetch_comments` on a blocked post now returns 10 top comments (scores 133/56/46…) with comment insights.

## Testing

- [x] `uv run pytest` (full suite; coverage 88.40% vs the 84% gate — the one failure, `test_setup_wizard.py::TestWriteApiKey::test_unwritable_target_returns_false`, also fails on clean `main` here because the suite runs as root, which bypasses the test's read-only permission setup; unrelated to this change)
- [x] Added tests:
  - `test_reddit_arctic.py::TestFetchPostComments` — normalization, score-desc ranking, `t3_` prefix handling, limit cap, deleted/removed filtering, rate-limit/network degradation, empty input
  - `test_reddit_shreddit.py::TestFetchCommentsFallback` — fallback fires on empty shreddit result, skipped when shreddit delivers, both-empty returns empty, never raises
  - Updated `test_fetch_failure_returns_empty` to mock the fallback lane

## Changelog

- [x] Added `changelog.d/+reddit-arctic-comments-fallback.added.md`

### AI review

Mirrors the review done for the listing-lane fallback (PR #960): the fallback only engages when the shreddit partial yields nothing; `comment_insights` extraction runs on the normalized rows so the renderer contract (`top_comments`/`comment_insights` shape) is unchanged; `num_comments` degradation to `None` matches the existing "never raises, caller falls through" contract (`_enrich_one` in `reddit_keyless` only sets a comment count when non-None). Both new paths are covered by mocked-network tests and one live verification run.

### Security

No new command execution or auth paths — one new HTTPS GET to the same public read-only archive the skill already calls (score backfill). No secrets involved. Degradation on `{"error": ...}` / network failure returns empty, never raises.

## Notes

Comment rows are snapshots from the archive (slightly stale vs live, same caveat as the score backfill). Sorting is client-side by score over a 100-row recency slice. If the shreddit comment partials come back for a host, behavior is unchanged (shreddit first, arctic only as fallback).

- [x] None

## Related issues

Companion to `feat/reddit-arctic-listing-fallback` (#960) — together they restore full keyless Reddit (discovery, scores, and comments) on datacenter-egress hosts.

### DIFF
--- a/changelog.d/+reddit-arctic-comments-fallback.added.md
+++ b/changelog.d/+reddit-arctic-comments-fallback.added.md
@@ -1,0 +1,1 @@
+Reddit comment enrichment now falls back to the arctic-shift archive when the shreddit comment partials return nothing — hosts on datacenter egress (where Reddit 403s `/svc/shreddit`) keep top-comment text (ranked by score) instead of silently degrading to counts only.
--- a/skills/last30days/scripts/lib/reddit_arctic.py
+++ b/skills/last30days/scripts/lib/reddit_arctic.py
@@ -16,16 +16,21 @@ count rather than failing the Reddit source.
 
 import sys
 import time
-from typing import Dict, List
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
 
 from . import http
 
 API = "https://arctic-shift.photon-reddit.com/api/posts/ids"
+COMMENTS_API = "https://arctic-shift.photon-reddit.com/api/comments/search"
 BATCH = 50          # ids per request
 TIMEOUT = 15
 MAX_BATCHES = 3     # cap total requests per run (bounds latency + rate-limit risk)
 PACE_SECONDS = 0.4  # gap between batches; arctic-shift answers 422 "slow down"
 CACHE_MAX = 4096    # hard size bound so the in-run memo can never grow unbounded
+# Comment fallback: fetch a wide recency slice, then rank by score client-side
+# (the archive's sort_type only supports created_utc).
+MAX_COMMENT_ROWS = 100
 # In-run memo: base36 id -> {score, num_comments}. Module-level so repeated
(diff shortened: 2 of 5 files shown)


mvanhorn:
Thank you @technicianofthesacred. #1083 landed the Reddit lane fixes for the rate-limit and fusion problems; this arctic-shift comment fallback addresses datacenter egress 403s and is still wanted. It is conflicting against main now, so a rebase after #1083 would be welcome.

mvanhorn:
Speaking as Matt's firstmate: classified **opt-in / corrective** (arctic-shift comment fallback when shreddit returns empty; same public archive already used for score backfill). Greptile **4/5** (fallback can drop a known `num_comments` total). Merge state **DIRTY**; captain already asked for a rebase after #1083 (2026-08-31 PT). Please rebase, fix the Greptile comment-total finding, and re-green. Author last commit 2026-08-08 (~44d) with no author follow-up after the rebase ask — assessing for stale-close (will not close without captain OK).

PR #967, opened 2026-08-09
fix(github): query is:issue and is:pull-request when authenticated

### LINKED ISSUE #966: GitHub source silently returns 0 results when a token is present (HTTP 422)
With any GitHub credential available — `GITHUB_TOKEN`, or just `gh auth login` picked up by the `gh auth token` fallback in `_resolve_token` — the GitHub source returns no results on v3.18.4.

```
[GitHub] 422 unprocessable: https://api.github.com/search/issues?q=rust+async+created%3A%3E2026-08-02&sort=reactions&order=desc&per_page=15
✓ Research complete (0.6s) - Github: 0 results
```

GitHub now requires `is:issue` or `is:pull-request` on **authenticated** `/search/issues` requests. Anonymous requests are still grandfathered, so the source works fine until a user adds a token — at which point it fails silently.

Confirmed directly against the API from the same host:

```
WITH token:    http=422  "Query must include 'is:issue' or 'is:pull-request'"
WITHOUT token: http=200  total_count: 8115
```

Two things make this hard to spot:

1. Adding a credential is normally an improvement, so the failure arrives at the moment a user expects things to get better.
2. `doctor` reports `github [CLI: gh ✓]` as WORKING regardless, because it checks that the `gh` binary exists rather than that the source returns anything.

**Note for whoever fixes it:** appending a single qualifier resolves the 422 but is lossy. For that same query GitHub reports 1,072 issues and 7,044 pull requests against 8,115 combined, so `is:issue` alone silently discards ~87% of matches.

PR opening shortly that queries both and merges, keeping coverage and the authenticated rate limit.

Environment: v3.18.4 (`1004324`), Ubuntu 24.04, Python 3.12.3.

Mariomarquezt:
Fixes #966.

## Problem

GitHub rejects authenticated `/search/issues` requests that carry neither `is:issue` nor `is:pull-request` with HTTP 422. Anonymous requests are still grandfathered, so the GitHub source works until a user supplies a credential — including implicitly, through the `gh auth token` fallback in `_resolve_token`.

The result is a silent `Github: 0 results` for anyone who has run `gh auth login`, while `doctor` still reports the source as healthy because it only checks that the `gh` binary exists.

## Why not just append `is:issue`

That fixes the 422 but halves the corpus. Measured against the live API:

| Query | `total_count` |
|---|---|
| no qualifier (anonymous) | 8,115 |
| `is:issue` | 1,072 |
| `is:pull-request` | 7,044 |

So a single qualifier would silently drop ~87% of matches — a quieter bug than the one being fixed.

## Approach

When authenticated, run both qualifier-scoped queries and merge:

- dedupe by item id, since an id can surface in both result sets
- **re-sort by reaction count before truncating** — each sub-query is reaction-sorted independently, so a plain concatenation would let the second query's tail outrank the first query's head
- unauthenticated path unchanged, still a single qualifier-free query

Cost is one extra request per search on the authenticated path, which is what the authenticated rate limit is for. The alternative — dropping to anonymous search — keeps coverage but gives up that headroom.

## Verification

Live, on the affected host, with a token present:

```
[GitHub] Found 15 issues/PRs
[GitHub] Enriching top 3 items with comments
✓ Research complete - Github: 6 results
```

Envelope inspection confirms the merge keeps both categories and the ordering holds:

```
total items: 15
  pull requests: 14
  issues       : 1
  reactions sorted desc: True
```

The 14:1 split tracks the 7,044:1,072 corpus ratio, so nothing is being preferentially dropped.

## Tests

- `test_planner_qualifiers_stripped_before_query_build` now asserts the stripped-subject invariant across **both** sub-queries rather than against one exact string. The helper also records every request — it previously kept only the last URL, which would have masked the second query.
- New `test_authenticated_search_merges_issues_and_pull_requests` covers dedupe and reaction ordering.
- New `test_unauthenticated_search_omits_qualifier` pins the anonymous path staying qualifier-free.

Full suite green (`pytest tests/` exit 0) on Python 3.12.13.


mvanhorn:
Thanks for this PR! We cherry-picked your authenticated GitHub search fix into #994, plus a small follow-up so a failed issues/PRs partition reports as partial instead of silent success. Really appreciate the contribution.

Closing this in favor of #994. That replacement is CI-green and Greptile 5/5; it is not on main yet.

Mariomarquezt:
Cool, glad this was helpful.

PR #971, opened 2026-08-09
fix(doctor): retry rate-limited probes instead of reporting a false outage

obbax:
## Summary

`doctor` reported Reddit as `✕ NOT WORKING` while Reddit was serving research runs fine in the same session. The probe had been refused with HTTP 429, and `_PROBE_BLOCKED_STATUSES` treats a blocked status as permanent. This retries a rate-limited probe once and, if it is still refused, reports it as **unverified** instead of a false outage. 403 stays a hard failure — that is the standing keyless block, not a burst.

The asymmetry is the bug: the research lane already retries with backoff across several endpoints, so it rides out the very 429 that a single unretried probe was treating as death. The health check was stricter than the thing it was checking, and it pointed users at a source that had nothing wrong with it.

## Testing

- [x] `uv run pytest` — 3602 passed, 6 skipped, 48 subtests passed

Five new tests in `tests/test_doctor.py`: the retry fires exactly once and sleeps the declared backoff; a probe that lands on the second attempt reports plain success with no transient residue; 403 is not retried; the carve-out is scoped per source; and `audit_state` maps a transient probe failure to unverified rather than not-working. All five fail against `main` and pass with the fix.

## Changelog

- [x] Added `changelog.d/+doctor-transient-probe.fixed.md` (`fixed`)
- [ ] Skip changelog

### AI review

Written and reviewed by a coding agent (Claude). Risks checked:

- **Could the retry mask a genuine outage?** It is bounded to one retry, and a source that refuses twice is surfaced as *unverified*, not as working. The fix trades a false negative for an honest "unknown", never for a false positive.
- **Could 403 get swept into the same bucket?** It is excluded explicitly and covered by a test asserting no retry and no transient flag.
- **Could the carve-out leak to other sources?** It is keyed per source, mirroring the existing `_PROBE_BLOCKED_STATUSES` structure, with a test pinning that scope.
- **Probe budget.** One extra 2s sleep, only on the failing path of a source that is already rate-limiting us.

The review also flagged that the new `time.sleep` would slow the existing `test_reddit_probe_429_is_not_reachable` by the full backoff. The shared test helper now patches `time.sleep`, so suite runtime is unchanged.

### Security

N/A. No input handling, command execution, path handling, auth, secrets, or dependency changes. The diff adds a stdlib `time` import, two module-level constants, and a retry branch inside an existing probe.

## Notes

Found while debugging a real run: `doctor` called Reddit down in a session where Reddit had just returned results. Reported as unverified, the same state reads honestly.

- [x] None

## Related issues

N/A

### DIFF
--- a/changelog.d/+doctor-transient-probe.fixed.md
+++ b/changelog.d/+doctor-transient-probe.fixed.md
@@ -1,0 +1,1 @@
+`doctor` no longer reports a rate-limited source as an outage. A probe refused with HTTP 429 is retried once and, if still refused, shown as unverified instead of `NOT WORKING`. Reddit was the visible case: a burst of keyless probes draws a 429 while the research lane, which retries with backoff, serves the same query fine. HTTP 403 still counts as a hard failure.
--- a/skills/last30days/scripts/lib/doctor.py
+++ b/skills/last30days/scripts/lib/doctor.py
@@ -48,6 +48,7 @@ import json
 import os
 import shutil
 import sys
+import time
 import urllib.error
 import urllib.request
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


mvanhorn:
Speaking as Matt's firstmate:

PR #972, opened 2026-08-09
fix(x): only run auto-extracted handle lanes on person-anchored runs

obbax:
## Summary

Auto-extracted X handle lanes now run only when the caller anchored the run to a person (`--x-handle`, `--x-related`, or `--github-user`). Explicitly named handles are untouched; person runs behave exactly as before.

`entity_extract.extract_entities(..., max_handles=3)` returns whichever `@handles` appear most in the phase-1 corpus, and the pipeline then runs `from:` and `@mentions` lanes on each. On a **person** topic that is exactly right — those handles are the subject and their circle. On a **thematic** topic it is just whoever happened to get mentioned, and their timelines swamp the topic search that found the relevant posts.

Measured on one thematic run before the change: **39 of 43 X items** came from three auto-extracted handles, none of them about the topic, while the topic-search lane that had found on-topic posts contributed about 5. No flag could turn it off.

Same topic, before and after:

| | X items | from handle lanes | on-topic |
|---|---|---|---|
| Before | 15 | majority | ~10% |
| After | 18 | 0 | 100% |

Person regression check (`--x-handle=steipete --github-user=steipete`): handle lanes run as before and the gate logs no skip.

## Testing

- [x] `uv run pytest` — 3599 passed, 6 skipped, 48 subtests passed

Two new tests cover the gate in both directions: auto-extracted handles are skipped with no anchor, and the same handles run once `github_user` is set. Both fail against `main` and pass with the change.

**One thing worth reviewing closely:** eight existing tests in `TestSupplementalSearches` now pass `x_handle="analyst1"`. Those tests cover backend routing and dedupe, not gating — they reached the handle-lane code incidentally, via auto-extraction. Anchoring them keeps them testing what they were written to test. I chose that over weakening the gate to keep them green, but it is a judgement call and I would rather flag it than bury it.

## Changelog

- [x] Added `changelog.d/+x-handle-lane-person-anchor.fixed.md` (`fixed`)
- [ ] Skip changelog

### AI review

Written and reviewed by a coding agent (Claude). Risks checked:

- **Does this regress person topics?** That is the main risk, since person runs are where the lane earns its keep. The skill's person-topic pre-flight always sets one of the three anchors, so the gate opens exactly when it used to matter. Verified with a live person run plus a dedicated test.
- **Is an explicit anchor the right signal, versus inferring intent from the topic string?** Inference would be a second guess layered on the guess that already misfired. The three flags are the caller stating what the run is about, and they are already set on the path that needs the lane.
- **Silent behaviour change?** No — a skipped lane logs at source level with the handle count and the flags that would enable it (`tty_only=False`, per the repo rule).
- **Scope.** Explicit `--x-handle` / `--x-related` handles bypass the gate entirely; only auto-extracted ones are affected.

### Security

N/A. No input handling, command execution, path handling, auth, secrets, or dependency changes. The diff threads an existing parameter through one call and adds a boolean gate plus a log line.

## Notes

Found while tuning thematic runs across five measured runs on the same engine version. The related finding — that X search is literal rather than semantic, so long natural-language `search_query` values collapse the lane (1 post for a full phrase vs 29-30 for 2-3 keywords) — is not addressed here; it is query-plan authoring, not pipeline behaviour.

- [x] None

## Related issues

N/A


mvanhorn:
Speaking as Matt's firstmate: classified **default-behavior** (thematic runs no longer fan out auto-extracted X handle lanes; person-anchored runs unchanged) with a **corrective** quality intent. Greptile **5/5**, but merge state is **DIRTY/CONFLICTING** — please rebase onto current `main` and re-green CI. When conflict-free, this still needs a captain product call before merge (default thematic X behavior change). Not merging; not resolving conflicts from triage.

Issue #906, opened 2026-07-29
Comment-enrichment slots follow relevance-dominant result order, so the highest-comment thread can miss enrichment


cassneilson:
## Summary

Comment-enrichment slots are allocated in result-ranking order. That order is deliberately relevance-dominant with a heavily damped engagement term, which is the right call for ranking results but the wrong call for choosing which threads to spend scarce comment fetches on. The practical effect is that the thread with the most comments in a corpus can be excluded from comment enrichment while near-empty threads consume slots.

Not a crash. A slot-allocation observation, filed because `## Top Community Comments` is the headline value of the tool and this is what silently starves it.

## Mechanism

Two things compose here.

**1. Slots are scarce and taken off the top of the list.** `lib/reddit_shreddit.py`:

```python
ENRICH_LIMITS = {"quick": 3, "default": 5, "deep": 8}
```

`lib/reddit_keyless.py` `_enrich` (L181-185) takes a straight prefix:

```python
limit = ENRICH_LIMITS.get(depth, ENRICH_LIMITS["default"])
to_enrich = posts[:limit]
rest = posts[limit:]
```

**2. The order it slices is relevance-dominant.** `lib/reddit_keyless.py` `_relevance_rank_key` (L45-54):

```python
total = (eng.get("score", 0) or 0) + (eng.get("num_comments", 0) or 0)
return (post.get("relevance") or 0.0) + min(0.25, math.log10(total + 1) / 20.0)
```

Engagement is present but damped to near-irrelevance for realistic values:

| thread | score | comments | total | engagement bonus |
|---|---|---|---|---|
| richest in corpus | 39 | 45 | 84 | 0.09647 |
| near-empty | 1 | 4 | 5 | 0.03891 |

Spread: **0.058**. The `0.25` cap requires a combined total of 99,999 to bind, so in practice the `/20` divisor is the whole story.

`_slot_priority` (L214-252) then reorders entity-matches ahead of misses but preserves the incoming order within each tier, so this ranking is what reaches `_enrich`.

The damping is correct and intentional *for ranking* — the docstring says it is sized so it cannot "lift an off-topic post (relevance ~0) above an on-topic one." The issue is only that enrichment reuses it.

## Observed

Topic: `side hustles people are actually starting and making money with right now`, default depth, v3.18.4, keyless Reddit path, two runs with different subreddit splits.

[...]

PR #976, opened 2026-08-09
feat(reddit): optional Scrapling browser fallback for blocked shreddit comments

OSideMedia:
## Summary

Reddit comment enrichment gains an **optional, CLI-gated browser fallback** for the case where the keyless shreddit HTTP path is blocked (HTTP 403 from many contexts). When the [`scrapling`](https://github.com/D4Vinci/Scrapling) CLI is installed on PATH, `reddit_shreddit.fetch_comments` falls through to a `scrapling extract stealthy-fetch` that renders the same `/svc/shreddit/comments` page in a real browser and feeds the **existing** parser.

While wiring this up I found a real parser gap: a browser lowercases HTML attribute *names* during DOM serialization, so the server partial's `thingId` arrives as `thingid`. The parser's attribute lookup was case-sensitive and silently dropped every comment (found the page, extracted zero). Fixed by making the attribute-name match case-insensitive — backward-compatible with the server partial.

## Design

- **Strictly additive.** The fallback only runs *after* the keyless HTTP path already returned nothing, and no-ops to `None` when `scrapling` is not on PATH. Behavior is unchanged everywhere the CLI is absent (CI, Cowork, default installs).
- **New helper** `lib/scrapling_fetch.py` wraps `scrapling extract` through the existing `subproc.run_with_timeout`, mirroring the other CLI-gated optional sources (`arxiv-pp-cli`, `yt-dlp`, `digg-pp-cli`). Returns `None` on any failure — never raises.
- **No new required dependency.** `scrapling` is optional and manually installed (`uv tool install 'scrapling[all]' && scrapling install`), documented as a new row in `CONFIGURATION.md`.
- Scope was kept honest: Bluesky's block is IP-level (a same-IP browser can't clear it) and Trustpilot's AWS WAF was **not** cleared by a stealthy-fetch in live testing, so neither was wired and the Trustpilot WAF harvest is untouched.

## Testing

- `tests/test_scrapling_fetch.py` — 13 new tests (helper at ~90% line coverage): availability gate, argv construction, success path, and every failure mode (absent CLI, unknown mode/format, non-zero exit, timeout, empty output).
- `tests/test_reddit_shreddit.py` — extended: browser-lowercased-attribute parse, fallback-fires-when-blocked, no-regression when the keyless path succeeds, and graceful no-op when `scrapling` is absent.
- **Proven live:** `stealthy-fetch` of a real `/svc/shreddit/comments` endpoint returned HTTP 200 with 161 `shreddit-comment` markers, and the existing parser (with the case fix) extracted real scored comments.
- Plugin/version-lockstep contract test unaffected; no version files touched. A `changelog.d/` fragment is included.
- Full suite is green except one pre-existing, environment-dependent test (`test_bare_run_emits_web_promo` expects `BRAVE_API_KEY` in a footer; it fails identically on a clean tree when a ScrapeCreators key is present in the environment — unrelated to this change).

## Agent disclosure

This change was authored with Claude Code. It has been reviewed and is being submitted by a user of the skill; happy to adjust to fit your contribution conventions (fragment wording, naming, the optional-dependency framing, or the SKILL.md surface).


OSideMedia:
Both review findings addressed in 02eb1c8:

1. **Timeout vs. enrichment budget**: enrichment workers now carry the 45s `ENRICH_BUDGET` as a monotonic deadline down to the Scrapling fallback. The browser subprocess timeout is capped to the time remaining (`subproc.run_with_timeout` kills the process group at the cap), and the fallback is skipped entirely when <10s remain — so a slow fetch can no longer outlive a cancelled future or leave browser processes running past the budget.

2. **Temp-file creation**: `NamedTemporaryFile` is now inside its own `OSError` guard in `scrapling_fetch.fetch`, honoring the documented never-raise contract on a full/unwritable temp dir.

New tests cover the deadline cap, the skip-when-nearly-spent path, the no-deadline default, the budget handoff in `reddit_keyless._enrich`, and the temp-file failure; each was verified to fail against the pre-fix code.

OSideMedia:
Rebased onto current `main` (3.23.0) and both of Greptile's open points are fixed; all checks green, no new findings on the rebased head.

- **403-only trigger** (`93a1713`): `fetch_comments` now wraps the keyless GET in `http.tee_failures()` and only falls back when a 403 was actually recorded — a 5xx, a 429, a silent `None`, or an empty 200 body never launches a browser. Three negative tests, each red on the pre-fix code.
- **Cleanup inside the budget** (`135c1c4`): the cap reserves `run_with_timeout`'s SIGTERM/SIGKILL window (10s) before comparing against the minimum, so timeout + cleanup fits inside the enrichment deadline instead of overrunning it by up to ten seconds.
- Two of the #906 slot-priority test mocks were widened to accept the `deadline` kwarg the budget cap passes.

Per the note on #1050, happy for this to go through the beta channel first — but `last30days-skill-private` is not visible to me. If you'd like it there, an invite and I'll open the same branch against it; otherwise it's ready here whenever you want it.

mvanhorn:
Speaking as Matt's firstmate: classified **opt-in** (Scrapling CLI on PATH only; no-op when absent) plus a small **corrective** shreddit attribute-name parse fix. CI green, mergeable CLEAN, Greptile **5/5**. No VISION.md on default branch. Ready to merge pending captain OK — stopping before merge per standing order.

PR #995, opened 2026-08-14
feat(setup): zero-click Amazon onboarding via GitHub

mvanhorn:
## Summary

Makes Amazon buyer signals work out of the box. First-run setup offers the capability to every user; accepting installs the Bright Data CLI and registers a free account through their existing GitHub login — no browser, no signup form, no credit card. Declining installs nothing.

Follow-up to #988, which shipped the source but reached only users who had already installed and authenticated the CLI themselves.

## Blocked on an upstream fix — do not merge expecting it to work end to end

`brightdata login --github` **is broken in CLI v0.3.3**. It fails server-side at Bright Data with `HTTP 403 verification_failed — "Gist content does not match expected nonce"`. Six client-side causes were ruled out (content mangling, propagation race, secret-gist readability, decorated-vs-bare nonce, public gist, missing scope), and their `/init` endpoint returns a `gist_content` containing a literal `undefined` where the email belongs.

Their PM has promised a fix. **Every unit here is built and unit-tested against mocks; live registration verification is deliberately skipped and will stay skipped until that fix ships.** Reporting a false pass would be worse than reporting none.

A second, independent defect surfaced while building this: **the CLI exits 0 when auth fails.** That one survives the nonce fix, because any future auth failure will also exit clean. Both are reported upstream.

One question remains that we cannot answer from outside, and the feature's whole premise rests on it: **does this flow register a new account, or only log in an existing one?** Every test ran against an already-registered account. If registration still needs a web signup, the zero-click promise does not hold for new users.

## What ships

| Unit | |
|---|---|
| U1 | `npm i -g @brightdata/cli` from setup, with off-PATH detection asking the same question the engine gate asks |
| U2 | `setup --brightdata` — ordered prerequisites, then `brightdata login --github` |
| U3 | A seventh onboarding stage in both branches |
| U4 | `LAST30DAYS_AMAZON_ENABLED` satisfies the request half of the dual gate |
| U5 | doctor and setup branch on the `gh` prerequisite; docs and changelog |

**Availability changes; firing does not.** The activation key makes `amazon` available without editing `INCLUDE_SOURCES`, but whether it runs on a given topic stays the model's per-run judgment. Without that split, a default-on lane would start spending credits on topics with no product dimension. A test pins it.

**The onboarding contract was extended, not weakened.** `tests/test_onboarding_contract.py` asserts a strict anchor ordering that AGENTS.md treats as locked. The new stage appends after the ScrapeCreators opt-in, so every prior assertion holds unmodified — net diff on that file is +64/−1, and the single deletion is a docstring line describing the new order.

## Testing

- [x] `uv run pytest` — 0 failures

Coverage 88.58% against the 84% floor. Live registration smoke skipped for the reason above.

## Changelog

- [x] Added `changelog.d/+brightdata-zero-click-onboarding.added.md`

### AI review

An independent reviewer returned eleven findings; all were fixed.

**The most important one changed a design.** My "did registration succeed?" check inferred success from *not matching* a failure string. The reviewer's point: any wording the pattern misses reads as success, and that is precisely the dangerous direction — the user is told Amazon is on while the lane is dark. Success is now a positive post-condition (the credential must actually exist); the pattern survives only to pick a useful message once failure is already established.

Also caught: re-running setup would have **overwritten an existing paid Bright Data login**, silently billing later pulls to the wrong account; a **TypeError after the point of no return** in no-config mode, crashing once the package was installed and the account created; and an **upsert bug** where a user who set the key to `0` and later re-registered got `persisted: true` while the source stayed off.

Two consent gaps: the copy enumerated everything sent to Bright Data and everything done on GitHub but never said a **global npm package** lands on the machine — disclosed in CONFIGURATION.md and the changelog, but not at the point of acceptance, which is the only place it counts. And `persisted: false` had no branch, though the neighbouring ScrapeCreators flow already had one.

### Security

The GitHub token never enters this process: `gh` holds it, the Bright Data CLI reads it, and `gh auth status` is checked for exit status only — never `--show-token`, never `gh auth token`. Our code writes exactly one thing, `LAST30DAYS_AMAZON_ENABLED=1`; the Bright Data API key is owned by the CLI's own credentials file. A test asserts no token-shaped value appears in any returned dict.

Hardening from review: vendor error strings are relayed verbatim to users, and an auth-failure line is exactly where a CLI echoes back the credential it was handed, so token-shaped values are scrubbed first. `gh auth status` is pinned to `github.com` — an Enterprise-only login passed the check and failed later, reporting the wrong first-failing prerequisite. Neither subprocess inherits stdin, since the CLI prompts interactively even in a non-TTY.

**Consent is the only control on this flow**, since it creates a third-party account from a user's identity. Its contents are asserted in both branches rather than left as prose: Bright Data named explicitly, the GitHub identity disclosed, the transient gist disclosed, the global npm install disclosed, no-credit-card stated, free tier unconflicted.

## Notes

Deliberately first-run only. Existing users stay dark until a follow-up offers the capability to them — its own design problem.

The `$100 community promo code` remains deferred and orthogonal to the free tier this depends on.

- [x] None
- [ ] Yes — disclosure:

## Related issues

Follow-up to #988.


mvanhorn:
Speaking as Matt's firstmate:

Closing as superseded by #1030 (conflict-resolved copy of this onboarding). We are not merging Amazon zero-click until Bright Data's `brightdata login --github` is actually fixed upstream — that hold stays on #1030. Thank you for the original design notes on this PR.

PR #1010, opened 2026-08-16
fix: read host JSON from stdin

miketuckman:
## Summary

Allow all host-authored JSON inputs (`--plan`, `--competitors-plan`, `--judgments`, and `--angles`) to use `-` for stdin. Update the agent skill templates to use quoted heredocs instead of temporary files with `rm -f` cleanup traps, which Codex can reject before the research command executes.

Inline JSON and existing JSON file paths remain supported.

## Testing

- [x] `uv run pytest` — 4,007 passed, 6 skipped, 48 subtests passed
- [x] Added or updated tests that would catch a regression, including end-to-end stdin coverage for all four JSON inputs and a host-contract check that rejects reintroducing cleanup traps

## Changelog

- [x] Added `changelog.d/+stdin-json-inputs.fixed.md` (types: `added`, `changed`, `fixed`, `removed`, `deprecated`, `security`)

### AI review

Codex reviewed backward compatibility for inline and file-backed JSON, error handling for invalid stdin JSON, shell quoting in every documented heredoc, and the full three-leg discovery protocol. The review identified indentation drift in one new fixture, which was corrected before the full test run. `git diff --check` is clean.

### Security

Stdin content follows the same `json.loads` parsing and schema validation as existing inline/file inputs. The quoted heredocs suppress shell interpolation, and removing temporary files also removes the associated cleanup command and filesystem race surface. No auth, secrets, or dependencies changed.

## Notes

The generic Codex skill validator was also attempted, but this repository's existing `SKILL.md` frontmatter contains supported project metadata fields that the generic validator does not accept. The repository's own complete test suite and skill contract tests pass.

- [x] None
- [ ] Yes — disclosure: 

## Related issues

N/A — observed directly in Codex Desktop when the local command safety layer rejected the mandatory `rm -f` cleanup trap before execution.


mvanhorn:
Speaking as Matt's firstmate:

**Triage (FM-CLEAR-L30-2):** DRAFT fix read host JSON from stdin — mark ready or close

PR #1018, opened 2026-08-18
feat(env): add Secret Service (libsecret) credential source

ferdousbhai:
## Summary

Adds a Secret Service (libsecret) credential source to `lib/env.py` — the Linux/BSD desktop analog of the existing macOS Keychain and `pass`(1) sources. Keys stored in any `org.freedesktop.secrets` provider (GNOME Keyring, KWallet, KeePassXC) under the attribute `service=last30days-<KEY>` are now picked up automatically, mirroring Keychain's service-name convention. New `scripts/setup-keyring.sh` mirrors `setup-keychain.sh` for storing them.

The gap this closes: on a Linux desktop the only non-`.env` option today is `pass`, which needs GPG. Most desktop users already have a running keyring their browser and other apps use.

## Testing

- [x] `uv run pytest`

`4021 passed, 5 skipped` locally. New `tests/test_env_libsecret.py` (17 cases) mirrors `test_env_pass.py` one-for-one: binary-missing, Darwin short-circuit, present/missing keys, first-line + whitespace handling, timeout/OSError swallowing, stop-on-timeout, prefix honoring, the five `get_config` precedence cases, and the `ALL_KEYS`/`KEYCHAIN_KEYS` drift guard.

One pre-existing failure is unrelated to this branch and also fails on a clean `HEAD` on my machine: `TestGrokExpiryStates::test_no_grok_cli_is_missing` asserts `"not found on PATH" in grok.detail`, but the detail string is the richer "grok is installed at … but that directory is not on this process's PATH" when the binary exists off-PATH. Host-specific; left alone.

## Changelog

- [x] Added `changelog.d/+libsecret-secret-service.added.md`

### AI review

Written by Claude Opus 5 via Claude Code, reviewed against `AGENTS.md` and `CONTRIBUTING.md` before opening.

Risks checked and what changed as a result:

1. **Test pollution from real host credentials.** The new loader initially broke `test_env_pass.py::test_get_config_reports_pass_source` and `test_env_keychain.py::test_get_config_reports_env_only_when_keychain_empty` — both stub `_load_keychain`/`_load_pass` but nothing stubbed the new source, so `get_config` read my actual keyring and reported `_CONFIG_SOURCE == 'libsecret'`. Both now stub `_load_libsecret` too. Worth flagging that this is latent for the Keychain source on macOS as well; CI is unaffected because no Secret Service daemon runs there.
2. **Hermes scanner.** `tests/hermes/test_no_critical_scan_findings.py` flags any `os.environ.get()` on a literal containing `KEY`/`TOKEN`/`SECRET`/`PASSWORD`/`CREDENTIAL`. My first two names for the override (`LAST30DAYS_LIBSECRET_PREFIX`, then `LAST30DAYS_KEYRING_PREFIX`) both matched. Resolved by routing the lookup through the existing `read_secret_env()` helper, which is exactly what its docstring describes; the name kept is `LAST30DAYS_KEYRING_PREFIX` since it matches `setup-keyring.sh`.
3. **Cost when the source is unused.** Copied both of the `pass` loader's guards — prefix resolved at call time, and only keys still unset after higher-priority sources are probed — plus the stop-after-first-timeout behaviour, so a locked collection cannot stall a config load by 5s per key.
4. **Coverage gate.** `fail_under` untouched; the suite passes with coverage enabled.

### Security

- No secrets committed; all tests use obvious dummy values (`xai-abc`, `sk-from-keyring`).
- Secrets are read transiently from a subprocess and never written to disk or logged, matching the `pass` loader.
- Lookups are exact-attribute; no shell interpolation. The prefix is caller-supplied config, not attacker-controlled.
- **Deliberate honesty in the docs**: unlike Keychain and `pass`, a Secret Service item is only as protected as its collection. A password-protected keyring encrypts; a passwordless one leaves items readable by anything running as the user — equivalent to a `0600` file, not encryption. I verified this empirically (a stored probe value appears verbatim in a passwordless keyring file). `CONFIGURATION.md`, the section heading, and `setup-keyring.sh`'s header all say so plainly rather than implying encryption, and `--collection` is offered for users who want a password-protected keyring.

## Notes

- Section heading changed from "Encrypted credential sources (Keychain / pass)" to "OS credential sources (Keychain / keyring / pass)" because the new source is not unconditionally encrypted and the old title would have been misleading once it was listed there.
- `_CONFIG_SOURCE` gains a `libsecret` value; anything matching on that field exhaustively should learn it.
- Priority is Keychain > Secret Service > `pass`, so macOS behaviour is unchanged and the loader short-circuits on Darwin.

- [x] None

## Related issues

N/A


mvanhorn:
Speaking as Matt's firstmate: classified as opt-in (Secret Service / libsecret credential source, Linux/BSD analog of Keychain/pass). Greptile 3/5 (first-run detection + setup-keyring prefix mismatch — not merge-ready). mergeable=CONFLICTING / DIRTY. Please rebase onto current main and address Greptile. Conflict resolution is hard work → factory ship path only. No merge ask (Greptile <5/5 + conflicts). Not a security vulnerability flag; credential-source addition stays opt-in.

PR #1024, opened 2026-08-19
fix: argv substitution corrupts runtime preflight; drop contradicting RECOMMENDATIONS template

rollingaccess-droid:
Two independent fixes against `main` (3.21.1). The commits are separable if you only want one.

---

## 1. Argv substitution corrupts the runtime preflight

Slash-command argument substitution rewrites bare shell positional parameters inside `SKILL.md` with the user's own argument words **before the model reads the file**. The Runtime Preflight block used positional parameters in its helper bodies, so it arrives corrupted on essentially every real invocation.

Invoking `/last30days how to improve the skill` delivers this to the model:

```bash
try_last30days_python() { candidate="to"   # was the first positional param
windows_path_to_unix()  { path="to"        # same
awk '/^version:/{gsub(/"/,"",improve); print improve; exit}'
```

`try_last30days_python` then ignores its argument and probes for a binary literally named `to`. Every candidate returns 1, the Windows discovery loop and the uv fallback both fail, `LAST30DAYS_PYTHON` never resolves, and the skill exits with:

```
ERROR: last30days v3 requires Python 3.12+.
```

on a machine that has 3.12+. Reproduced on Windows 11 with Python 3.13.13 installed and on PATH.

It only misfires when the topic is two or more words, which is nearly all of them, and `SKILL.md` is correct on disk. That combination is why it survives review: you cannot see it by reading the file.

**Fix:** move the preflight to `skills/last30days/scripts/preflight.sh` and source it. Replace the `awk` version lookup with a `sed` expression that uses no positional parameters. Also reworded the surrounding prose, since positional tokens in prose get rewritten too.

`preflight.sh` sets `LAST30DAYS_PYTHON` and `LAST30DAYS_MEMORY_DIR` in the calling shell, so it must be sourced rather than executed. Its header comment records why the logic cannot live inline, to keep it from being folded back in later.

**Verified:** sourcing resolves Python 3.13.13 through the Windows discovery path, and `last30days.py --diagnose` runs clean under it, reporting 16 available sources.

## 2. A RECOMMENDATIONS template that contradicts its own rule

`SKILL.md` carries two conflicting instructions for RECOMMENDATIONS output, 141 lines apart.

The `### If QUERY_TYPE = RECOMMENDATIONS` section says to rank by signal quality rather than mention count, and names a mention-count leaderboard as the 2026-04-18 failure mode:

> **BAD RECOMMENDATIONS synthesis (counting):**
> "🏆 Most mentioned: Python (15 mentions), TypeScript (10x), Go (7x), Rust (5x)."

The later display section then hands the model that exact shape as the template to fill in, `{n}x mentions` slots included.

A model resolving the conflict picks arbitrarily, and the mention-count version is the more concrete of the two, so it tends to win. That reproduces the documented failure the earlier section exists to prevent.

**Fix:** delete the mention-count template, point at the signal-weighted one, and rename its `Sources:` field to `Voices:` to match the field name that template actually specifies.

---

Happy to split into two PRs or adjust the wording of the added `SKILL.md` prose if you'd prefer it shorter.

### DIFF
--- a/skills/last30days/SKILL.md
+++ b/skills/last30days/SKILL.md
@@ -1860,23 +1860,14 @@ Identify from the ACTUAL RESEARCH OUTPUT:
 
 **FIRST - What I learned (based on QUERY_TYPE):**
 
-**If RECOMMENDATIONS** - Show specific things mentioned with sources:
-```
-🏆 Most mentioned:
-
-[Tool Name] - {n}x mentions
-Use Case: [what it does]
-Sources: @handle1, @handle2, r/sub, blog.com
-
-[Tool Name] - {n}x mentions
-Use Case: [what it does]
-Sources: @handle3, r/sub2, Complex
-
-Notable mentions: [other specific things with 1-2 mentions]
-```
+**If RECOMMENDATIONS** - use the signal-weighted template from the
+`### If QUERY_TYPE = RECOMMENDATIONS` section above. Rank by signal quality, NOT
+by mention count. Do NOT emit a `🏆 Most mentioned:` list with `{n}x mentions`
+counts - that exact shape is the named 2026-04-18 failure mode documented in that
+section, where counting buried the real headline.
 
 **CRITICAL for RECOMMENDATIONS:**
-- Each item MUST have a "Sources:" line with actual @handles from X posts (e.g., @LONGLIVE47, @ByDobson)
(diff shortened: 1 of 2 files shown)


mvanhorn:
Speaking as Matt's firstmate:

**Triage (FM-CLEAR-L30-2):** BLOCKED+CI fail argv preflight fix — needs green CI

PR #1027, opened 2026-08-20
fix(xurl): recognize the ~/.xurl/auth.yml directory layout

### LINKED ISSUE #978: Bug: xurl diagnostic treats ~/.xurl directory as a missing token store
## Summary

Last30Days v3.18.4 can successfully use xurl for X research, but its diagnostic path reports xurl as unavailable when xurl uses its current directory-based credential layout.

## Environment

- macOS
- xurl v1.3.1 installed via Homebrew Cask
- Last30Days plugin v3.18.4

## Observed behavior

- `xurl auth status` succeeds and reports OAuth2 plus bearer configured.
- `xurl whoami` succeeds.
- xurl stores credentials at `~/.xurl/auth.yml`; `~/.xurl` is necessarily a directory.
- Direct backend calls return `stored_auth_status=('missing', 'no token store at ~/.xurl')` while `is_available=True`.
- `last30days.py --diagnose` omits X and returns `x_backend: null`, but actual research records `x_search_backend=xurl` and returns X data.

## Root cause

In `scripts/lib/xurl_x.py`, `token_store_path()` returns `Path.home() / '.xurl'`. `stored_auth_status()` then requires that path to satisfy `path.is_file()`, so it returns missing for the valid `.xurl` directory.

A filesystem symlink cannot make both expectations true: a directory symlink still fails `is_file()`, while a file symlink at `~/.xurl` breaks xurl's required child path `~/.xurl/auth.yml`.

## Suggested fix

Have `token_store_path()` return `~/.xurl/auth.yml` for current xurl layouts, with a legacy-file fallback if needed. Please add one regression test for the directory-based layout.

SomSamantray:
## Summary

`--diagnose` and doctor no longer report xurl as unauthenticated when credentials live in the current `~/.xurl/auth.yml` directory layout. `stored_auth_status()` now resolves the actual credential file across both the directory layout and the legacy flat `~/.xurl` file, and a permission-denied store surfaces as the typed `AUTH_ERROR` instead of a misleading "no token store" (pathlib's `is_file()`/`is_dir()` swallow `OSError` — the fix probes via `stat`).

## Testing

- [x] `uv run pytest`

New tests cover: directory `auth.yml` with markers → OK; empty/missing `auth.yml` → missing; legacy flat file (direct and parent walk-up) → OK; permission-denied read and permission-denied parent stat → `AUTH_ERROR`; grandparent-permission-denied (a distinct, previously-uncovered `except OSError` branch guarding the directory-vs-file candidate resolution) → `AUTH_ERROR`; no subprocess spawned in the diagnostic path; and — new this round — `--diagnose`'s `x_backend`/`available_sources` fields reflect the fix end-to-end through the real CLI entrypoint (not just at the internal `has_stored_auth()` unit level), for both the credentialed and no-store cases. Full suite (excluding network-dependent tests) passes, aside from one pre-existing, unrelated flaky test in `test_amazon.py` (confirmed via repeated runs and a fresh `main` clone under the same conditions — not touched by this PR).

## Changelog

- [x] Added `changelog.d/978.fixed.md`

### AI review

Reviewed across three rounds: correctness, testing, reliability, and security lenses (rounds 1–2, pre-merge), then a third round after independent reviewer confirmation added an end-to-end regression test for `--diagnose` — correctness, project-standards, testing, security, adversarial, and previous-comments lenses. Round 1 flagged that pathlib's OSError-swallowing probes misreported a permission-denied store as `AUTH_MISSING` — fixed with stat-based probes that propagate `EACCES` while keeping `FileNotFoundError` as absence. Round 2 confirmed the fix holds (no P0–P2 findings). Round 3 (this round) found and fixed one real P2 test-coverage gap (the grandparent-permission-denied branch, above) and one P2 documentation-accuracy issue (a test docstring overclaiming end-to-end coverage of the `AUTH_ERROR` path, which `--diagnose` cannot actually distinguish from `AUTH_MISSING` — narrowed to state that scope honestly); security, project-standards, and previous-comments lenses returned clean. All prior residual P3 items are recorded below and in the residual-review file; nothing regressed.

### Security

`stored_auth_status()` reads a local credential file to detect xurl OAuth auth; it scans only for fixed marker substrings and never echoes credential values. The error detail strings echo the store path and errno only. No new input handling, command execution, or secrets exposure. `N/A` for injection/SSRF risk — local-only filesystem read, no network, no subprocess.

## Notes

**Reviewer caveat, investigated and resolved.** `@pollymallen`'s comment raised a concern that hosts building their source list from `--diagnose` (per `SKILL.md`) might still see X/xurl as unavailable even with a fix that works internally. Reproduced with a clean environment (no leaked credentials) and confirmed `--diagnose`'s `x_backend`/`available_sources` already resolve correctly on this branch — matching the "with this branch" output in that comment exactly. This is now covered by a permanent regression test rather than a one-off manual check. Scope note: this confirms the fix for the two store layouts `stored_auth_status()` recognizes today (current `~/.xurl/auth.yml` and the legacy flat file); a future xurl release writing a third, unrecognized layout would reproduce the same class of issue and is out of this PR's scope. The comment's timeline: it landed after round 1/round 2's fixes were already on this branch, so the caveat was checked against the same code state, not an earlier one.

The walk-up (`path.name == "auth.yml"` → parent) is keyed to the current canonical file name; a future upstream rename of `auth.yml` would need the same change. `AUTH_MISSING` detail re-invokes `token_store_path()` (cosmetic — identical value in scope). Both recorded in the residual-review findings file.

**One deferred finding (not fixed in this PR):** round 3's adversarial lens found a pre-existing TOCTOU race in `stored_auth_status()`'s read path — if the credential store is deleted between the existence check and the read, the typed `AUTH_ERROR` is reported instead of the more accurate `AUTH_MISSING`. This predates this PR (the old code had the same race with a single `is_file()` check) and is narrow (requires a concurrent rewrite of xurl's own store during a `--diagnose`/`doctor` run). Flagging for a follow-up rather than expanding this PR's scope.

- [x] None

## Related issues

Fixes #978

### DIFF
--- a/changelog.d/978.fixed.md
+++ b/changelog.d/978.fixed.md
@@ -1,0 +1,1 @@
+`--diagnose` and doctor no longer report xurl as unauthenticated when credentials live in the current `~/.xurl/auth.yml` directory layout; the legacy flat `~/.xurl` file is still recognized.
--- a/docs/residual-review-findings/fix-xurl-x978.md
+++ b/docs/residual-review-findings/fix-xurl-x978.md
@@ -1,0 +1,28 @@
+## Residual Review Findings
+
+Source run: `ce-code-review` on branch `fix/xurl-x978` (issue #978), plan `docs/plans/2026-08-20-fix-xurl-x978-plan.md`. Run dirs: `20260820-220120-767b63a3` (round 1), `20260820-221941-c2d05f7f` (round 2).
+
+### Round 1 findings (P1 fixed, P3s applied)
+
+- **P1 — `stored_auth_status()` masked permission-denied store as AUTH_MISSING** (`skills/last30days/scripts/lib/xurl_x.py`). `pathlib`'s `is_file()`/`is_dir()` swallow `OSError` into `False`, so a chmod-000 `~/.xurl/auth.yml` reported "no token store" instead of the typed `AUTH_ERROR`. Fixed with stat-based probes (`_is_file`/`_is_dir`) that propagate `EACCES`; `FileNotFoundError` stays absent → MISSING. Commits `d1ef34e`, `98795ae`.
+- **P3 — legacy flat file precedence** (`xurl_x.py`): candidate order now prefers the live `auth.yml` over a stale flat file when both are physically possible. Covered by `test_canonical_path_preferred_when_both_layouts_exist`.
+- **P3 — AUTH_MISSING detail naming** (`xurl_x.py`): `token_store_path()` is re-invoked in the detail string; cosmetic (identical value in scope). Not changed — behavior identical, one redundant `Path.home()` call in a rare error path.
+- **P3 — test stub fallback** (`xurl_x.py`): the `except (AttributeError, TypeError)` fallback for path stubs was made explicit and now routes correctly; covered by existing `_UnreadableStore`/`_BrokenFile` tests.
+
+### Round 2 findings (P3, advisory)
+
+- **P3 — dangling-symlink docstring overclaim** (`xurl_x.py:118-138`): `_is_file` returns `False` (MISSING) for a dangling symlink since `stat()` raises `FileNotFoundError` on a missing target. Docstring updated to scope the broken-store claim to permission failures; a dangling target is a missing target and stays MISSING.
+- **P3 — stat-PermissionError on candidate scan untested** (testing reviewer): added `test_permission_denied_parent_stat_reports_error_not_missing` (chmod-000 parent dir → AUTH_ERROR) and hardened `test_permission_denied_read_reports_error_not_missing` with a root-euid guard.
+- **P3 — "both layouts exist" test naming** (testing reviewer): the two layouts cannot physically coexist (`~/.xurl` is either a file or a directory), so the test pins canonical-path selection rather than a file-vs-dir race. Left as-is with accurate comment.
+- **P3 — redundant `token_store_path()` in MISSING detail** (reliability reviewer, advisory): cosmetic; no behavior change.
+
+### Not filed as tracker tickets
+
+All remaining items are P3 advisory/cosmetic with no user-visible impact; the two with real regression value were applied as tests. No GitHub issues filed.
+
+### Verification
+
+- `uv run pytest tests/test_xurl_x.py` → 54 passed
+- `uv run pytest tests/test_env_v3.py tests/test_backend_descriptors.py tests/test_diagnose_compat.py` → 132 passed
+- `uv run pytest -q --ignore=tests/test_network.py` → full suite passed
+- `git diff --check` clean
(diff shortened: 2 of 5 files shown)


pollymallen:
Independent confirmation — still reproduces on **v3.23.1** (issue reported against 3.18.4), and this PR fixes it.

**Environment:** macOS 26.3.1, xurl **1.3.1 installed via npm** (`@xdevplatform/xurl`), Python 3.14.6, app-only bearer auth.

Worth noting the original report used Homebrew Cask — npm produces the same `~/.xurl/` directory layout, so this isn't install-method specific.

**Before (v3.23.1, unpatched):**
```
stored_auth_status: ('missing', 'no token store at /Users/…/.xurl')
is_available:       True
--diagnose:         x_backend: null, "x" absent from available_sources
```

**With this branch:**
```
token_store_path:   /Users/…/.xurl/auth.yml
stored_auth_status: ('ok', 'stored OAuth credentials found in …/auth.yml')
--diagnose:         x_backend: "xurl", "x" present in available_sources
```

`uv run pytest tests/test_xurl_x.py` → 54 passed.

One note for anyone hitting this via an agent host: the engine's own path uses the live check and works fine, but hosts that build their source list from `--diagnose` (as SKILL.md instructs) will see X as unavailable and can omit it from the query plan. Cosmetic in headless use, load-bearing in agent-driven use.

SomSamantray:
@pollymallen Thanks for the thorough repro and for flagging the `--diagnose` caveat.

Investigated it directly: reproduced the same scenario with a clean environment (no leaked env-var credentials) and confirmed `--diagnose`'s `x_backend`/`available_sources` already resolve correctly on this branch — matching your "with this branch" output exactly (`x_backend: "xurl"`, `"x"` present in `available_sources`). That's now locked in by a permanent regression test (`tests/test_diagnose_compat.py::DiagnoseXurlAuthWiring`) that drives the real `--diagnose` CLI entrypoint end-to-end with a fake `~/.xurl/auth.yml`, rather than a one-off manual check — verified non-vacuous by confirming it fails against the pre-fix code and passes against current code.

One scope note worth being explicit about: this confirms the fix for the two store layouts `stored_auth_status()` recognizes today (the current `~/.xurl/auth.yml` directory layout and the legacy flat file). A future xurl release writing a third, unrecognized layout would reproduce the same class of issue — that's out of this PR's scope but worth keeping in mind.

Also ran a fresh audit pass over the rest of the PR's changed files and a full code review round on top of the new test — found and fixed one real test-coverage gap (a previously-uncovered error branch in `stored_auth_status()`) and tightened a docstring that slightly overclaimed what the new test proves. Full details in the updated PR description.

mvanhorn:
Speaking as Matt's firstmate:

PR #1028, opened 2026-08-20
fix(reddit): add anti-bot resilience via host cooldown, cache, and old.reddit fallback

nish3451:
Lands the dirty-worktree work that had been sitting idle >24h.

- Serialize requests per host (www/old) and honor bounded Retry-After cooldowns so parallel fan-out cannot amplify a 403/429 block.
- Fall back to old.reddit.com for subreddit feeds after a www 403 or malformed anti-bot page, canonicalizing post links back to www.
- Cache successful feeds (fresh 60s / stale 300s) to bridge a bounded outage and dedupe concurrent callers.
- Detect YouTube's anti-bot/sign-in interstitial in yt-dlp stderr and the watch page so it is not misreported as zero demand or no captions.
- Update xurl docs to the official @xdevplatform/xurl package.
- Remove stale test-run.log dev artifact.

Tests: tests/test_reddit_rss.py, tests/test_youtube_yt.py


mvanhorn:
Thank you @nish3451. Your feed cache was the prior art for the per-command request memo that landed in #1083, which is credited in that change. The host cooldown and old.reddit fallback target datacenter 403s, a different problem from the rate-limit loss #1083 fixed, so this stays open for a follow-up; a rebase onto main after #1083 would need to sit alongside the memo and the 1 req/s bucket rather than its own per-host serialization.

mvanhorn:
Speaking as Matt's firstmate: classified as corrective (Reddit anti-bot resilience / host cooldown / old.reddit fallback). Greptile 4/5 (host-wide cooldown on ordinary feed failures — not merge-ready). mergeable=CONFLICTING / DIRTY. Captain already noted (2026-08-31) this stays open for follow-up rebase alongside #1083's memo / 1 req/s bucket. Please rebase and address Greptile. Conflict resolution is hard work → factory ship path only. Author clock last push 2026-08-20 (>14d); treating captain "stays open for a follow-up" as soft hold — **not** proposing stale-close without Firstmate/captain OK.

PR #1036, opened 2026-08-22
fix(transcribe): send the skill User-Agent in _post_audio so Groq's Cloudflare edge stops returning 403 (#983)

### LINKED ISSUE #983: Groq transcription always fails: _post_audio sends no User-Agent, Cloudflare returns 403 error code: 1010
# Groq transcription always fails: `_post_audio` sends no `User-Agent`, Cloudflare returns 403 `error code: 1010`

**Version:** 3.11.0
**Impact:** Caption-free YouTube transcription via Groq is broken for every user, and fails silently.

## Summary

`lib/transcribe.py:_post_audio` builds its request with only `Authorization` and `Content-Type` headers. Python's urllib therefore sends `User-Agent: Python-urllib/3.x`, and Groq's Cloudflare edge rejects that client with **HTTP 403** and a bare body of `error code: 1010` (banned client fingerprint).

Because `transcribe_media` is documented "Never raises", the failure surfaces as an empty transcript with no error — the user sees missing transcripts and no reason.

## Reproduction

With a valid `GROQ_API_KEY` and ffmpeg installed:

```python
from lib import env as E, transcribe as T
cfg = E.get_config()
name, key = E.transcription_providers(cfg)[0]   # ('groq', ...)
T._post_audio(name, "clip.mp3", key, 60)
# urllib.error.HTTPError: HTTP Error 403: Forbidden
# body: error code: 1010
```

The same key, file, and endpoint succeed via curl:

```
$ curl -X POST https://api.groq.com/openai/v1/audio/transcriptions \
    -H "Authorization: Bearer $GROQ_API_KEY" \
    -F "model=whisper-large-v3" -F "file=@clip.mp3"
{"text":" The quick brown fox jumps over the lazy dog. Testing transcription."}
HTTP 200
```

## Cause, isolated

One variable changed, everything else identical:

| `User-Agent` | Result |
|---|---|
| *(omitted — current code)* | 403 `error code: 1010` |
| `Python-urllib/3.12` (explicit) | 403 `error code: 1010` |
| `last30days/3.11.0` | **200 OK**, correct transcript |
| `curl/8.7.1` | 200 OK, correct transcript |

The ban keys on the `Python-urllib` agent string, not on the header being absent. Any honest product agent clears it.

## Suggested fix

```diff
+_USER_AGENT = "last30days/3.11.0"   # or derive from plugin.json version
+
 req = urllib.request.Request(
     endpoint,
     data=body,
     headers={
         "Authorization": f"Bearer {api_key}",
         "Content-Type": f"multipart/form-data; boundary={boundary}",
+        "User-Agent": _USER_AGENT,
     },
     method="POST",
 )
```

Verified locally: `transcribe_media()` then returns `ok=True, provider=groq, chunks=1, health=ok` with correct text.

## Related: `doctor` prescribes the unreachable fix

`doctor` reports `youtube` as `ok` and prescribes "add a free Groq key from console.groq.com so caption-free videos still get transcripts". In 3.11.0 that prescription cannot work — and because the failure is a 403, it reads as an auth error, so users will re-check a key that was never the problem.

A health probe that distinguishes "no key configured" from "key rejected at the edge" would turn a silent dead end into an actionable line.

terry-li-hm:
Fixes #983.

## Summary

`_post_audio` posted multipart audio with `Authorization` and `Content-Type` but no `User-Agent`, so urllib sent `Python-urllib/3.x`. Groq's Cloudflare edge 403s that fingerprint with `error code: 1010`, and the chunk loop swallowed the exception, so Groq silently degraded to the next provider or an empty transcript.

The request now sends `http.USER_AGENT` (the same identity `lib/http.py` already uses for get/post). Provider failures are logged at debug under `LAST30DAYS_DEBUG` instead of being dropped, so the next person sees which provider failed and why.

## Testing

- [x] `uv run pytest`

New class `TestPostAudioUserAgent` in `tests/test_transcribe.py`: `test_user_agent_is_canonical_skill_ua`, `test_user_agent_not_urllib_default`, `test_user_agent_sent_for_openai_too`. Full suite green on this branch (one environment-dependent test, `TestGrokExpiryStates::test_no_grok_cli_is_missing`, deselected locally because this host has the grok CLI installed).

## Changelog

- [x] Added `changelog.d/+groq-transcribe-user-agent.fixed.md`
- [ ] Skip changelog

## Agent disclosure

### AI review

Reviewed `_post_audio` header construction and the chunk-loop except path: the new `User-Agent` is the static skill identity already used by `lib/http.py`, `Authorization` is unchanged, and the debug line records `{exc!r}` (the urllib/HTTP error), never the bearer token.

### Security

No new input handling. The User-Agent is a compile-time string; API keys stay in the `Authorization` header only. Tests use dummy keys and a mocked `urlopen`.

## Notes

The same UA is sent for OpenAI so a later edge fingerprint there does not regress the helper. Debug lines appear only with `LAST30DAYS_DEBUG`.

- [x] None
- [ ] Yes — disclosure:

## Related issues

Fixes #983.

### DIFF
--- a/changelog.d/+groq-transcribe-user-agent.fixed.md
+++ b/changelog.d/+groq-transcribe-user-agent.fixed.md
@@ -1,0 +1,1 @@
+Audio transcription posts the skill's canonical User-Agent instead of Python-urllib, so Groq's Cloudflare edge no longer 403s the request with error 1010.
--- a/skills/last30days/scripts/lib/transcribe.py
+++ b/skills/last30days/scripts/lib/transcribe.py
@@ -24,7 +24,7 @@ import tempfile
 from dataclasses import dataclass, field
 from typing import Optional
 
-from . import env, health, subproc
+from . import env, health, http, log, subproc
 
 # Whisper's documented upload ceiling. We compress to stay under it and chunk
 # when a single clip still exceeds it.
@@ -177,7 +177,11 @@ def _transcribe_chunk(
     for name, key in providers:
         try:
             text = _post_audio(name, path, key, timeout=timeout)
-        except Exception:  # noqa: BLE001 - any provider failure -> try the next
+        except Exception as exc:  # noqa: BLE001 - any provider failure -> try the next
+            # Surface the failing provider + error (e.g. Groq's Cloudflare 403,
+            # "error code: 1010") under LAST30DAYS_DEBUG instead of silently
+            # degrading to an empty transcript (#983).
+            log.debug(f"transcribe: provider {name} failed on a chunk: {exc!r}")
             text = None
         if text is not None:
             return text, name
@@ -214,6 +218,10 @@ def _post_audio(provider: str, path: str, api_key: str, timeout: float) -> Optio
         headers={
             "Authorization": f"Bearer {api_key}",
             "Content-Type": f"multipart/form-data; boundary={boundary}",
+            # A plain-urllib UA is fingerprinted and 403'd by Cloudflare-style
+            # edges (Groq returns "error code: 1010", #983). Send the skill's
+            # canonical identity instead, same as lib/http.py's get/post.
+            "User-Agent": http.USER_AGENT,
         },
         method="POST",
     )
--- a/tests/test_transcribe.py
+++ b/tests/test_transcribe.py
@@ -2,7 +2,7 @@
 
 from unittest import mock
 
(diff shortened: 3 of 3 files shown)


mvanhorn:
Speaking as Matt's firstmate:

PR #1037, opened 2026-08-22
fix(reddit): pace the keyless bucket at 1 req/s with an env knob and retry 429'd sub-requests once (#985)

### LINKED ISSUE #985: Reddit: keyless token bucket (5 req/sec) outpaces the unauthenticated surface — every run loses sub-requests to 429 with no in-lane retry
## Summary

The shared keyless-Reddit token bucket (`REDDIT_KEYLESS_LIMITER = RateLimiter(rate_per_sec=5.0, burst=5)`) paces faster than reddit.com's unauthenticated surface tolerates once the engine's own fan-out is applied on top of it. The result is that a **routine** run loses a portion of its Reddit sub-requests to HTTP 429, every time. Individual lanes swallow those failures with no in-lane retry, so the loss is diffuse and unrecoverable within the run.

This is adjacent to but distinct from #867 (which covered the ScrapeCreators backfill being unreachable behind a non-empty free result, now addressable via `LAST30DAYS_REDDIT_SC_MIN_ITEMS`). That knob compensates for the symptom; this issue is about the pacing that produces it.

Observed on v3.18.4, macOS, residential IP, no VPN.

## Measurements

All three tests exercise `reddit_rss.search_rss` + `reddit_listing.fetch_listings` for a single subquery (3 subreddits), under `http.capture_failures()`:

| Test | Sub-requests 429'd | Posts returned |
|---|---|---|
| Discovery pass immediately after a full `/last30days` run | 10 of 10 | 0 |
| Discovery pass on a fresh token bucket (120s idle) | 9 | 116 |
| Two lone `reddit_keyless_get_text` calls after 90s idle, 5s apart | 0 | both 200 OK |

The third row is the important one: **the IP is not blocked and the endpoints are not gone.** Requests spaced by seconds succeed reliably. The bucket's 5 req/sec sustained rate is simply above what these endpoints accept, and a real run multiplies it further — one full keyless composite per subquery (RSS feeds + listing partials across sorts + shreddit enrichment at `MAX_ENRICH_WORKERS=4`), with `reddit` carrying no entry in `MAX_SOURCE_FETCHES`, all inside the outer `ThreadPoolExecutor`.

Across 23 saved runs on this machine, every run since v3.18.4 was installed reports Reddit `partial ... HTTP 429`, ending at 7, 29, 30, 31 and 45 items.

## Why this looks new but isn't

Worth recording, because it cost a full investigation to rule out: the 429s are almost certainly not new. v3.18.4 rewired the RSS and listing workers from `executor.submit` to `http.submit_with_context` (#899), which is what finally propagates the `capture_failures()` sink into those threads. `REDDIT_KEYLESS_LIMITER` is byte-identical in v3.8.3 and v3.18.4. Before the rewire the same 429s were discarded and the source reported clean — the upgrade changed observability, not execution. #899 was a correct fix; it just revealed a pre-existing condition that reads as a regression.

## Suggested directions

1. **Lower the sustained rate.** 5 req/sec against an unauthenticated host is aggressive. Empirically ~1 req/sec with a small burst is comfortable. Making it configurable (`LAST30DAYS_REDDIT_KEYLESS_RATE`) would let users trade wall-clock for coverage.
2. **Retry a 429'd sub-request within its lane.** `_fetch_feed` and `_fetch_one_with_status` currently record the failure and return nothing. A single backoff-and-retry per sub-request would recover most of what's lost, since the limit is transient rather than a block.
3. **Back off adaptively.** The bucket has no feedback path from observed 429s — the rate is fixed regardless of how many refusals the run is collecting.

## Minor: `partial after N items` wording

`render.py::_format_outcome` composes `partial after {items_returned} items`, where `items_returned` is the run's **final** post-filter count. It reads as "collection stopped at item N," which is not what happened — collection completed and lost some sub-requests along the way. Three runs here happened to land at 29/30/31 and the phrasing made that look like a hard cutoff at ~30, which sent the investigation off in the wrong direction for a while. Something like `partial — N items returned, some requests rate-limited` would be less inviting of that reading.

## Repro

```python
import sys, collections
sys.path.insert(0, "<skill>/scripts")
from lib import http, reddit_rss, reddit_listing

SUBS = ["ClaudeAI", "ExperiencedDevs", "ChatGPTCoding"]
with http.capture_failures() as failures:
    rss = reddit_rss.search_rss("ai coding agents", depth="default", subreddits=SUBS)
    listings = reddit_listing.fetch_listings(SUBS, depth="default", query="ai coding agents")

print(len(rss), len(listings), collections.Counter(f.status_code for f in failures))
```

terry-li-hm:
Fixes #985.

## Summary

Unauthenticated reddit.com traffic (RSS, listing, shreddit) shared one token bucket at 5 req/sec with burst 5. Engine fan-out on a typical home IP tripped HTTP 429 and those sub-requests were dropped; the partial outcome then read as `partial after N items`, which sounds like a cutoff rather than a rate limit.

The shared limiter now defaults to 1 req/sec with burst 2; `LAST30DAYS_REDDIT_KEYLESS_RATE` overrides the rate (invalid or non-positive values fall back to 1). RSS and listing fetch through `reddit_keyless_get_text_retry_429`, which retries a 429 once after a short jittered pause and re-acquires the bucket; a recovered 429 is not left in the pipeline sink. Partial outcomes that mention 429 or rate limiting now read `partial — N items returned, some requests rate-limited`. ScrapeCreators routing is unchanged.

## Testing

- [x] `uv run pytest`

New coverage in `tests/test_reddit_keyless_backoff.py`: `TestRedditKeylessRateKnob` (default 1/sec burst 2, env override, invalid fallback, process-env sync, `.env` export), `TestRedditKeyless429Retry` (429-then-200 recovers with two acquires; double 429 records and stops; non-429 is not retried; RSS and listing recover after one 429; listing records double 429), `TestPartialOutcomeWording`. `tests/test_reddit_rss.py` autouses a limiter-acquire patch so RSS unit tests do not sleep. Full suite green on this branch (one environment-dependent test, `TestGrokExpiryStates::test_no_grok_cli_is_missing`, deselected locally because this host has the grok CLI installed).

## Changelog

- [x] Added `changelog.d/985.fixed.md`
- [ ] Skip changelog

## Agent disclosure

### AI review

Reviewed the limiter default change, env parse/export, the in-lane 429 retry (local `capture_failures` so a recovered 429 is not recorded, then `tee_failures` on the retry), and the RSS/listing call sites. No credential paths touched.

### Security

Env parsing rejects non-finite and non-positive rates and falls back to 1. No new credential handling. The retry sleep is bounded (1s plus up to 0.5s jitter). Tests use dummy HTML/RSS fixtures, not live reddit.com.

## Notes

`CONFIGURATION.md` documents `LAST30DAYS_REDDIT_KEYLESS_RATE`. `LAST30DAYS_REDDIT_BACKEND` and `LAST30DAYS_REDDIT_SC_MIN_ITEMS` are unchanged. The 1 req/sec default trades wall-clock for fewer 429s; raise the knob on IPs with more headroom.

- [x] None
- [ ] Yes — disclosure:

## Related issues

Fixes #985.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -135,7 +135,7 @@ python3 skills/last30days/scripts/last30days.py "MCP servers" \
 | Source | Key(s) | Required for | Free tier |
 |---|---|---|---|
 | Local corpus | `--corpus <dir>` or `LAST30DAYS_CORPUS_DIRS` | private `.md`/`.txt`; `.pdf` when `pdftotext` is on PATH | yes (offline) |
-| Reddit (public) | none (default free keyless path). With `SCRAPECREATORS_API_KEY`: empty-only search backup by default; `LAST30DAYS_REDDIT_SC_MIN_ITEMS=<N>` backfills thin free runs; `LAST30DAYS_REDDIT_BACKEND=scrapecreators` pins SC primary with free fallback | always on; SC knobs require `SCRAPECREATORS_API_KEY` | yes |
+| Reddit (public) | none (default free keyless path). With `SCRAPECREATORS_API_KEY`: empty-only search backup by default; `LAST30DAYS_REDDIT_SC_MIN_ITEMS=<N>` backfills thin free runs; `LAST30DAYS_REDDIT_BACKEND=scrapecreators` pins SC primary with free fallback. `LAST30DAYS_REDDIT_KEYLESS_RATE` paces unauthenticated reddit.com requests (default `1` req/sec) | always on; SC knobs require `SCRAPECREATORS_API_KEY` | yes |
 | Hacker News | none | always on | yes |
 | Polymarket | none | always on | yes |
 | StockTwits | none | auto-on for ticker/crypto topics only (gated by symbol detection); never registered for non-financial topics | yes (public API, ~200 req/hr per IP) |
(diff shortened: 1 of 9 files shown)


mvanhorn:
Thank you @terry-li-hm. Your pacing change, the LAST30DAYS_REDDIT_KEYLESS_RATE knob, and the single in-lane 429 retry landed on main verbatim through #1083 (squash-merged with a Co-authored-by credit), and your #985 measurements were the basis for the whole plan. One thing worth knowing from the live runs: the 1 req/s bucket exposed that the fixed 20-second future timeouts expired while fetches were still queued, so #1083 also sizes those timeouts from the bucket's queue depth and memoizes repeat requests per command. Closing this as superseded by #1083; #985 closes with it.

mvanhorn:
Landed via #1083.

PR #1038, opened 2026-08-22
fix(grounding): keep undated keyed-web results instead of dropping the whole response (#928)

### LINKED ISSUE #928: Keyed web backends (serper/brave/exa/parallel) return 0 results: client-side date filter drops all undated organic results
# Keyed web backends (serper/brave/exa/parallel) return 0 results: client-side date filter drops all undated organic results

## Summary

`grounding.serper_search()` almost always returns **0 items**, even with a valid
Serper API key that returns results. The Serper API responds `200` with organic
results, but the client-side date filter discards every result whose `date`
field is missing or unparseable — and Serper/Google rarely populates `date` on
organic results. Net effect: the Serper web backend is effectively a no-op.

## Environment

- last30days `v3.18.4`
- macOS, Python 3.13
- Valid `SERPER_API_KEY` (confirmed working via raw API call, see below)
- `--web-backend serper` (also selected automatically via `native_web_backend`)

## Steps to reproduce

```python
from datetime import date, timedelta
from scripts.lib import grounding as g

today = date.today()
dr = ((today - timedelta(days=30)).isoformat(), today.isoformat())
items, meta = g.serper_search("OpenAI", dr, "<valid_key>", count=5)
print(len(items), meta)   # -> 0  {'label': 'serper', ..., 'resultCount': 0}
```

## Actual vs expected

- **Actual:** `serper_search()` returns `[]` (`resultCount: 0`) for common,
  result-rich queries.
- **Expected:** returns the organic results Serper actually returned.

## Root cause

`serper_search()` sends `tbs=cdr:1,cd_min:...,cd_max:...`, which already
constrains Google to the date window **server-side**. It then applies a
**second, client-side** filter on each result's `date` field:

```python
raw_date = r.get("date") or ""
pub_date = _parse_serper_date(raw_date)
if not _in_date_range(pub_date, date_range):   # <-- drops undated results
    continue
```

`_in_date_range(None, date_range)` returns `False`, so any result **without** a
parseable `date` is dropped. In practice Serper returns `date` on very few
organic results (often 0–1 of 5), so nearly everything is filtered out.

### Evidence

Raw POST with the exact `tbs` the function builds:

```
tbs = cdr:1,cd_min:07/03/2026,cd_max:08/02/2026
organic returned by Serper: 5
  ...of those, with a `date` field: 1   (and it was "Apr 28, 2025", out of window)
_in_date_range(None, dr) => False        # the 4 undated results are all dropped
=> serper_search() returns 0
```

## Suggested fix

The server-side `cdr` filter already bounds results to the window, so an
unknown/unparseable date should **not** drop a result. Only drop results whose
**known** date is out of range:

```diff
         raw_date = r.get("date") or ""
         pub_date = _parse_serper_date(raw_date)
-        if not _in_date_range(pub_date, date_range):
+        # `tbs=cdr:...` already constrains Google to the window server-side, and
+        # Serper rarely populates `date` on organic results. Dropping undated
+        # results filtered out ~everything (Serper returned 0 items). Only drop
+        # results whose KNOWN date is out of range.
+        if pub_date is not None and not _in_date_range(pub_date, date_range):
             continue
```

After this change, `serper_search("OpenAI", dr, key)` returns 4 results;
`"query x"` returns 2 relevant results, etc.

## Scope: all four keyed web backends are affected

This is not Serper-specific. `brave_search()`, `exa_search()`, and
`parallel_search()` in `grounding.py` use the **same** `if not
_in_date_range(pub_date, date_range): continue` guard on a `date` field that is
frequently missing. So every keyed web backend (`brave`, `exa`, `serper`,
`parallel`) drops undated results and can silently collapse to 0 items, leaving
only the keyless floor. The fix above should be applied to all four call sites
(or factored into a shared helper).

terry-li-hm:
## Summary

Keyed web backends (Brave, Exa, Serper, Parallel) no longer drop organic results whose date is missing or unparseable. The client-side window filter now drops only results whose **known** date is out of range, so a valid API response no longer collapses to zero items.

Fixes #928

## Testing

- [x] `uv run pytest`

Focused (`tests/test_keyed_web_date_filter.py` + `tests/test_grounding_v3.py`): 38 passed in 0.09s.

Full suite: `uv run pytest -q --deselect tests/test_backend_descriptors.py::TestGrokExpiryStates::test_no_grok_cli_is_missing` → 4085 passed, 5 skipped, 1 deselected, 53 subtests passed in 175.95s.

New `tests/test_keyed_web_date_filter.py` failed on all 9 cases before the fix (undated results emptied the backend) and passes after. Known out-of-range dates still drop. Existing grounding tests were updated to match the keep-undated contract.

## Changelog

- [x] Added `changelog.d/928.fixed.md` (types: `added`, `changed`, `fixed`, `removed`, `deprecated`, `security`)

## Agent disclosure

### AI review

Checked that undated and unparseable dates are kept, and that a known date outside the window still drops. Confirmed Brave freshness, Exa published-date bounds, and Serper `tbs=cdr` still go out on the request. Parallel still has no server-side date param; it gets the same keep-undated rule the issue asked for. Dummy keys only; HTTP mocked; no real network.

### Security

N/A. No new input handling, command execution, path handling, auth, secrets, or dependencies.

## Notes

Parallel does not send a date constraint on the API request. Known out-of-range Parallel dates still drop. Unknown dates are kept, matching the other three backends.

- [x] None
- [ ] Yes — disclosure: 

## Related issues

Fixes #928

### DIFF
--- a/changelog.d/928.fixed.md
+++ b/changelog.d/928.fixed.md
@@ -1,0 +1,1 @@
+Keyed web backends (Brave, Exa, Serper, Parallel) no longer drop organic results that omit a parseable date, so a valid API response no longer collapses to zero items.
--- a/skills/last30days/scripts/lib/grounding.py
+++ b/skills/last30days/scripts/lib/grounding.py
@@ -64,7 +64,7 @@ def brave_search(
     for i, r in enumerate((data.get("web", {}).get("results", []))[:count]):
         raw_date = r.get("page_age") or ""
         pub_date = _normalize_date(raw_date[:10]) if raw_date else None
-        if not _in_date_range(pub_date, date_range):
+        if _known_date_out_of_range(pub_date, date_range):
             continue
         items.append({
             "id": f"WB{i + 1}",
@@ -109,7 +109,7 @@ def exa_search(
             continue
         raw_date = r.get("publishedDate") or ""
         pub_date = _normalize_date(raw_date.split("T")[0] if "T" in raw_date else raw_date[:10]) if raw_date else None
-        if not _in_date_range(pub_date, date_range):
+        if _known_date_out_of_range(pub_date, date_range):
             continue
         items.append({
             "id": f"WE{i + 1}",
@@ -146,7 +146,7 @@ def serper_search(
     for i, r in enumerate((data.get("organic", []))[:count]):
         raw_date = r.get("date") or ""
         pub_date = _parse_serper_date(raw_date)
-        if not _in_date_range(pub_date, date_range):
+        if _known_date_out_of_range(pub_date, date_range):
             continue
         items.append({
             "id": f"WS{i + 1}",
@@ -187,7 +187,7 @@ def parallel_search(
             continue
         raw_date = r.get("publish_date") or ""
         pub_date = _normalize_date(raw_date[:10]) if raw_date else None
-        if not _in_date_range(pub_date, date_range):
+        if _known_date_out_of_range(pub_date, date_range):
             continue
         items.append({
             "id": f"WP{i + 1}",
@@ -363,6 +363,11 @@ def _in_date_range(pub_date: str | None, date_range: tuple[str, str]) -> bool:
     if not pub_date:
         return False
     return date_range[0] <= pub_date <= date_range[1]
+
+
+def _known_date_out_of_range(pub_date: str | None, date_range: tuple[str, str]) -> bool:
+    """True when a parseable date falls outside the window. Unknown dates stay."""
+    return pub_date is not None and not _in_date_range(pub_date, date_range)
 
 
 def _domain(url: str) -> str:
--- a/tests/test_grounding_v3.py
+++ b/tests/test_grounding_v3.py
@@ -24,17 +24,19 @@ class BraveSearchTests(unittest.TestCase):
                     {
                         "title": "Undated Article",
                         "url": "https://example.com/undated",
-                        "description": "Should also be filtered",
+                        "description": "Kept because date is unknown",
                     }
                 ]
             }
         }
         with patch("lib.grounding.http.request", return_value=mock_response) as mock_req:
             items, artifact = grounding.brave_search("test", ("2026-02-25", "2026-03-27"), "fake-key")
-            self.assertEqual(1, len(items))
+            self.assertEqual(2, len(items))
             self.assertEqual("Test Article", items[0]["title"])
             self.assertEqual("https://example.com/article", items[0]["url"])
             self.assertEqual("2026-03-10", items[0]["date"])
+            self.assertEqual("Undated Article", items[1]["title"])
+            self.assertIsNone(items[1]["date"])
             self.assertEqual("brave", artifact["label"])
             call_url = mock_req.call_args.args[1]
             self.assertIn("freshness=2026-02-25to2026-03-27", call_url)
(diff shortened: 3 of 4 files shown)


mvanhorn:
Speaking as Matt's firstmate: classified as corrective (keyed-web undated organic results retained at adapters). Fixes open #928. CI green, MERGEABLE/CLEAN, but Greptile 4/5 with open P1: downstream grounding still drops undated results after adapters retain them — so the PR does not yet deliver the claimed fix end-to-end. Greptile must be 5/5 before any merge ask. Please address the open thread. Not proposing merge.

maxitoon:
Rechecked head `67ea585` against current main `349ca44`. One additional constraint for addressing the [existing downstream-filter thread](https://github.com/mvanhorn/last30days-skill/pull/1038#discussion_r3834749719): **keyed does not necessarily mean server-windowed**.

In [`parallel_search()`](https://github.com/mvanhorn/last30days-skill/blob/67ea585ed3d5a8a40df6f4072f5323969e4c9789/skills/last30days/scripts/lib/grounding.py#L169-L201), the outgoing payload contains only `search_queries` and `advanced_settings.max_results`; neither date bound is sent. Brave sends `freshness`, Exa sends both published-date bounds, and Serper sends `tbs=cdr`. A blanket keyed-provider exemption downstream would therefore admit unverified-date Parallel results, including for historical `--as-of` windows. The PR body acknowledges the missing Parallel constraint; this matters when fixing normalization, not just at the adapter.

Offline verification on this head: the existing two focused test files pass (**38 passed**). A separate mocked-HTTP probe through the real `web_search()` → `normalize_source_items('grounding', ...)` returns **1 adapter item → 0 normalized items for each of Brave, Exa, Serper and Parallel**. The existing P1 is thus still reproducible; current main also retains [`require_date = source == "grounding"`](https://github.com/mvanhorn/last30days-skill/blob/349ca444b4fda466e74d471dffa2aff36bb997f1/skills/last30days/scripts/lib/normalize.py#L98-L101), with `_normalize_grounding()` mapping `date` to `published_at` and `filter_by_date_range()` rejecting missing dates. Local checks used Python 3.11.15 (below the project's 3.12+ requirement), so these are bounded diagnostic evidence, not supported-runtime/full-suite validation. No live provider calls.

Suggested minimal acceptance checklist on this PR:
- [ ] Preserve missing/unparseable-date Brave/Exa/Serper results through normalization using explicit per-item server-window provenance, with request-bound assertions.
- [ ] Keep unknown-date Parallel/keyless results strict unless a complete requested-window contract is established; do not infer trust from “keyed” or a display string.
- [ ] Add adapter → normalization fixtures for all-undated responses, mixed trusted/untrusted rows, known dates outside either bound, and historical windows.
- [ ] Refresh CI on the revised head and address the existing review thread before requesting maintainer review, consistent with the owner's latest triage.

mvanhorn:
Speaking as Matt's firstmate: re-checked after later activity (maxitoon note that keyed ≠ server-windowed, esp. Parallel). Still **corrective**; Fixes open #928. CI green, MERGEABLE/CLEAN, but Greptile 4/5 with open P1: downstream grounding still drops undated results after adapters retain them. Please address the end-to-end path (and Parallel's lack of server date bounds) and get Greptile to 5/5 before any merge ask. Not proposing merge. (CLEAN/MERGEABLE — not a stale-close candidate.)

PR #1042, opened 2026-08-23
fix(instagram): unwrap user reels media envelopes

### LINKED ISSUE #1017: Instagram --ig-creators silently returns 0 reels: _user_reels doesn't unwrap the media envelope
**Version:** 3.21.0 · **Platform:** macOS 24.6.0 (Darwin), Python 3.13 · **Host:** Claude Code

## Summary

`--ig-creators` fetches reels successfully and then discards every one of them. The user sees `Instagram: 0 items (error: HTTPError: HTTP 404: Not Found)`, which points at the network layer — but the network call succeeded. The reels are dropped locally by a shape mismatch between two functions.

## Repro

```
python3 scripts/last30days.py "AI news and updates" \
  --emit=compact --ig-creators=openai,anthropic,googledeepmind,nvidia
```

stderr shows the fetches working:

```
[Instagram] User reels: @openai
[Instagram]   -> 12 reels from @openai
[Instagram] User reels: @googledeepmind
[Instagram]   -> 12 reels from @googledeepmind
[Instagram] User reels error for @anthropic: HTTP 404: Not Found
```

Final report:

```
- Instagram: 0 items (error: HTTPError: HTTP 404: Not Found (run doctor for fixes))
```

24 reels in, 0 out.

## Root cause

The two ScrapeCreators Instagram endpoints return different shapes:

| Endpoint | Item shape |
|---|---|
| `/v2/instagram/reels/search` | flat reel object |
| `/v1/instagram/user/reels` | `{"media": { ...reel... }}` |

`_user_reels()` (`scripts/lib/instagram.py:272`) returns the `user/reels` items **still wrapped**:

```python
raw_items = data.get("items") or data.get("reels") or data.get("data") or []
_log(f"  -> {len(raw_items)} reels from @{handle}")
return raw_items
```

`_parse_items()` then reads fields off the outer dict — `raw.get("id")`, `raw.get("caption")`, `raw.get("like_count")`, and `_parse_date(raw)` → `item.get("taken_at")`. Against a `{"media": {...}}` envelope every one of those is absent, so `_parse_items` returns 12 objects with all fields `None`. The null `date` then fails the date-range filter downstream and the batch vanishes.

Confirmed by inspection — the envelope holds everything the parser wants, one level down:

```python
>>> raw = instagram._user_reels('openai', tok)
>>> list(raw[0].keys())
['media']
>>> {k: type(raw[0]['media'][k]).__name__ for k in
...  ['id','pk','code','caption','taken_at','play_count','like_count','comment_count','user','owner']}
{'id':'str','pk':'str','code':'str','caption':'dict','taken_at':'int',
 'play_count':'int','like_count':'int','comment_count':'int','user':'dict','owner':'dict'}

>>> instagram._parse_items(raw, 'ai')[0]['date']          # current behaviour
None
>>> instagram._parse_items([x['media'] for x in raw], 'ai')[0]['date']   # unwrapped
'2026-08-14'
```

The `# Author info -- 'owner' in reels/search, 'user' in user/reels` comment at `_parse_items` shows the intent was for one parser to serve both endpoints; only the outer envelope was missed.

Why the message says 404: with `items` empty after both the creator lane and the keyword lane, `search_and_enrich` returns `{"items": [], "error": last_error}` — and `last_error` is whatever the *keyword* search left behind. So the creator-lane bug is reported using the keyword lane's unrelated error.

## Suggested fix

Unwrap at the source so `_parse_items` keeps seeing a single shape (`scripts/lib/instagram.py`, after line 272):

```python
raw_items = data.get("items") or data.get("reels") or data.get("data") or []
# /v1/instagram/user/reels wraps each reel in a {"media": {...}} envelope, while
# /v2/instagram/reels/search returns them flat. Unwrap so _parse_items sees one
# shape; without this every field (taken_at included) reads as None and the date
# filter silently drops the whole batch.
raw_items = [it.get("media", it) if isinstance(it, dict) else it for it in raw_items]
_log(f"  -> {len(raw_items)} reels from @{handle}")
return raw_items
```

`it.get("media", it)` keeps it a no-op for any already-flat payload. Verified locally against v3.21.0: dates parse to `2026-08-14`, `2026-08-08`, `2026-08-04`, `2026-08-03` — all inside the window.

Happy to send this as a PR if useful.

---

## Two adjacent things found while diagnosing

Separate from the bug above; both cost real time to get to the bottom of, so noting them in case they're worth their own issues.

### 1. ScrapeCreators uses HTTP 404 to mean "no results", and the engine reports it as a failure

A keyword reel search that simply matched nothing returns:

```json
{"success": false, "credits_remaining": 9964, "credits_charged": 0, "error": "not_found",
 "errorStatus": 404, "message": "No scrapeable reels found for that query. You were not charged for this request."}
```

Note `credits_charged: 0` — SC is explicitly saying "empty result", not "failure". `http.get` raises on any non-200, so this surfaces in the report as `Instagram error: HTTP 404: Not Found` and in `doctor` as `✕ NOT WORKING`. On a broad topic like "AI news" the bare-token query `ai` legitimately matches nothing on SC's v2 reels endpoint (it wraps Google Search), so a healthy, correctly-configured install gets told its Instagram source is broken.

Treating a 404 whose body carries `"error": "not_found"` / `credits_charged: 0` as an empty result rather than an exception would make `doctor` and the `## Partial Coverage` block a lot more truthful. Same distinction the `no-results` vs `error` states in `source_status` already draw.

### 2. bird-search requires Node >= 22 but nothing checks before invoking bare `node`

`scripts/lib/vendor/bird-search/package.json` declares `"node": ">=22"`, and `lib/bird_x.py` gates only on `shutil.which("node") is not None` (`:169`, `:239`) before spawning `"node", str(_BIRD_SEARCH_MJS)` (`:271`, `:508`, `:580`).

On a machine whose `node` is older, X fails like this:

```
X: All X backends failed — bird: .../vendor/bird-search/lib/twitter-client-constants.js:2
import queryIds from './query-ids.json' with { type: 'json' };
                                        ^^^^
SyntaxError: Unexpected token 'with'
Node.js v18.15.0
```

Import attributes need Node >= 20.10 (`assert` before that), so any Node 18 fails here. This was genuinely hard to attribute: cookie auth had just been configured, `--diagnose` reported `x_backend: bird`, and `doctor` listed X as configured — every signal said X was working, while a raw Node stack trace was buried inside an "All X backends failed" string.

A version check next to the existing `which("node")` gate would turn that into something actionable, e.g. `X unavailable: bird-search needs Node >= 22, found v18.15.0 at /usr/local/bin/node`.

Worth noting the failure isn't limited to genuinely-old machines. In this case Node 22, 25 and 26 were all installed via Homebrew and `~/.zshrc` already prepended `node@22` — but a root-owned 2023 nodejs.org `.pkg` binary at `/usr/local/bin/node` sat earlier in `PATH` and won. `shutil.which("node")` faithfully returned the wrong one. Checking the resolved interpreter's version, rather than its existence, catches that class of environment cleanly.

Tethys0:
## Summary

Fix Instagram creator discovery dropping every reel returned by ScrapeCreators' `/v1/instagram/user/reels` endpoint. That endpoint wraps each reel in a `media` object, while the shared parser expects a flat reel mapping; the source now unwraps that envelope while preserving already-flat payloads.

## Testing

- [x] `uv run pytest tests/test_instagram.py -q` (12 passed)
- [x] `uv run pytest` (4078 passed, 5 skipped, 53 subtests passed)
- [x] Added a regression test covering parsed reel id, date, caption, and engagement from the wrapped response
- [x] `git diff --check` passed

## Changelog

- [x] Added `changelog.d/1017.fixed.md`
- [ ] Skip changelog - chore/internal only

### AI review

The change was developed with AI assistance and independently reviewed for spec compliance, endpoint-shape handling, regression risk, scope, and security. The reviewers found no blocking logic or security concerns. The focused Instagram tests and full test suite pass.

### Security

N/A. This is a local response-shape normalization change. No command execution, path handling, authentication behavior, dependency, or secret-handling code was changed.

## Notes

The source-boundary normalization keeps `_parse_items()` on one flat payload shape and leaves already-flat Instagram search results unchanged.

- [x] None
- [ ] Yes - disclosure: 

## Related issues

Fixes #1017

### DIFF
--- a/changelog.d/1017.fixed.md
+++ b/changelog.d/1017.fixed.md
@@ -1,0 +1,1 @@
+Instagram creator reels now unwrap ScrapeCreators `media` envelopes before parsing, preserving their metadata and date filtering.
--- a/skills/last30days/scripts/lib/instagram.py
+++ b/skills/last30days/scripts/lib/instagram.py
@@ -271,7 +271,15 @@ def _user_reels(
 
     raw_items = data.get("items") or data.get("reels") or data.get("data") or []
     _log(f"  -> {len(raw_items)} reels from @{handle}")
-    return raw_items
+
+    # User-reels responses wrap each reel in a ``media`` envelope, while
+    # other Instagram endpoints already return flat reel dictionaries.
+    return [
+        item["media"]
+        if isinstance(item, dict) and isinstance(item.get("media"), dict)
+        else item
+        for item in raw_items
+    ]
 
 
 def search_instagram(
--- a/tests/test_instagram.py
+++ b/tests/test_instagram.py
@@ -40,6 +40,37 @@ class TestInstagramOwnerTypeSafety(unittest.TestCase):
         self.assertEqual("fallbackuser", items[0]["author_name"])
 
 
+class TestInstagramUserReelsEnvelope(unittest.TestCase):
+    def test_wrapped_user_reel_is_parsed_with_its_fields(self):
+        from unittest.mock import patch
+        from lib import instagram
+
+        reel = {
+            "id": "reel-1017",
+            "code": "ABC1017",
+            "taken_at": "2026-08-01T12:34:56.000Z",
+            "caption": {"text": "Wrapped reel caption"},
+            "video_play_count": 123,
+            "like_count": 45,
+            "comment_count": 6,
+        }
+        response = {"items": [{"media": reel}]}
+
+        with patch.object(instagram.http, "get", return_value=response):
+            raw_items = instagram._user_reels("creator", token="test-token")
+
+        items = _parse_items(raw_items, "wrapped reel")
+
+        self.assertEqual(1, len(items))
+        self.assertEqual("reel-1017", items[0]["video_id"])
+        self.assertEqual("2026-08-01", items[0]["date"])
+        self.assertEqual("Wrapped reel caption", items[0]["text"])
+        self.assertEqual(
+            {"views": 123, "likes": 45, "comments": 6},
+            items[0]["engagement"],
+        )
+
+
 class TestInstagramComments(unittest.TestCase):
     """U1: Instagram comment fetching via ScrapeCreators."""
 


mvanhorn:
Speaking as Matt's firstmate:

PR #1048, opened 2026-08-23
fix(reddit): honour x-ratelimit-reset on 429, not just Retry-After

immy1991:
## Summary

Reddit's anonymous search and RSS endpoints answer a 429 with `x-ratelimit-reset` and **no** `Retry-After`, so the 429 branch in `http.fetch_url` and `reddit_public` always fell through to exponential backoff (3s, 5s, 9s) — every step shorter than the ~42s Reddit is asking for. Each retry re-429s, the retry budget drains, and the run reports Reddit as rate-limited with 0 items when the source was only being polled too early.

Adds `http.retry_delay_from_headers(headers, fallback)`, which reads `Retry-After` first and `x-ratelimit-reset` second, ignores unparseable and non-positive values, and falls back to the caller's existing backoff otherwise. Both retry sites now use it.

Reproduced against the live endpoint on 2026-08-23 from a residential macOS host:

```
HTTP/2 429
x-ratelimit-used: 1
x-ratelimit-remaining: 0.0
x-ratelimit-reset: 42
```

No `Retry-After` in that response. The captured headers are asserted verbatim in the new test so a future refactor can't silently reintroduce the gap.

## Testing

- [x] `uv run pytest`

Full suite before and after: **4084 passed, 6 skipped, 53 subtests passed**. New file `tests/test_reddit_ratelimit_reset.py` covers header precedence, the real captured Reddit response, unparseable values (`Retry-After` as an HTTP-date), non-positive values, and fallback behaviour.

## Changelog

- [x] Added `changelog.d/+reddit-ratelimit-reset.fixed.md`
- [ ] Skip changelog — chore/internal only

No `CHANGELOG.md` edit and no version/manifest bumps, per CONTRIBUTING.

## Agent disclosure

This PR was authored by Claude (Anthropic) working in Claude Code, at the direction of the repo owner of a downstream tool, and reviewed by them before submission.

### AI review

Main risks checked:

- **Scope creep into the rate limiter.** The repo already has a token-bucket `RateLimiter` and `Retry-After` handling; my first read wrongly assumed backoff was absent entirely. Re-read narrowed the change to the one real gap — the header name Reddit actually sends — rather than touching the limiter.
- **Behaviour change for non-Reddit hosts.** The helper returns the caller's existing fallback when neither header is present, so every other host keeps its current backoff exactly. Verified by the full suite passing unchanged.
- **Header-parsing edge cases.** `Retry-After` is legally an HTTP-date, not only delta-seconds; a date must not become a float. Handled and tested. A `0` or negative value is also rejected, because sleeping 0 just re-429s immediately.
- **Import cycle.** `reddit_public` now imports `lib.http`; checked that `http` does not import `reddit_public`.

### Security

`N/A` for secrets and auth — no credential, command execution, or path handling is touched. The only external input newly parsed is two response headers, both coerced through `float()` inside a `try/except` with a bounded fallback, so a hostile or malformed header cannot produce an unbounded sleep or an exception. No dependency changes.

## Notes

Two things worth flagging:

1. This makes the retry *correct*, not unlimited — anonymous Reddit is still ~1 request per 42s, so the practical effect is "a slow result instead of a guaranteed zero", not parity with a keyed path.
2. Separately, and outside this PR's scope: this repo doesn't appear to use Reddit's official OAuth API anywhere (`grep` finds no `REDDIT_CLIENT_ID` / `oauth.reddit.com`). A registered script app is free and allows 100 req/min with real scores, versus the RSS path which carries no score at all. Happy to open that as a separate PR or an issue if it's of interest — keeping it out of this one since it's a feature, not a fix.

- [x] None
- [ ] Yes — disclosure:

## Related issues

N/A — found while building a downstream tool against this engine.


mvanhorn:
Thank you @immy1991. Your commit landed on main verbatim through #1083 (squash-merged with a Co-authored-by credit), and the captured Reddit 429 headers in your test are what kept the follow-up honest: a later review pass added a 60-second cap and epoch-to-delta conversion on top of your parser, because GitHub sends x-ratelimit-reset as a timestamp. Closing this as superseded by #1083.

mvanhorn:
Landed via #1083.

PR #1050, opened 2026-08-24
fix(env): add LAST30DAYS_SKIP_KEYCHAIN so tests can seal the Keychain source

OSideMedia:
## Problem

`tests/test_footer_nudge_suppression.py::test_bare_run_emits_web_promo` fails on macOS for any contributor who has a `last30days-BRAVE_API_KEY` item in their Keychain. It passes in CI, so the failure is invisible to the project and looks like promo-selection drift to whoever hits it.

The test seals two credential sources:

```python
"LAST30DAYS_CONFIG_DIR": "",          # skip ~/.config/last30days/.env
...
for key in ("BRAVE_API_KEY", "EXA_API_KEY", ...):
    env.pop(key, None)                # strip paid web keys from os.environ
```

macOS Keychain is a **third, independent source**. `BRAVE_API_KEY` is resolved anyway, so:

`native_web_backend = "brave"` → `_missing_sources_for_promo()` returns `None` → the `BRAVE_API_KEY` nudge is never printed → `assertIn("BRAVE_API_KEY", combined)` fails.

The assertion had no subject. Measured on an affected machine:

```
reddit available?   True
x available?        True
native_web_backend: brave      <- from Keychain
native_search:      False
PROMO RESULT:       None
```

Confirmed as the cause both ways: the test is red normally and **green** when `security` is made unreachable via `PATH=/nonexistent`, with nothing else changed.

## Fix

`_load_keychain()` gains a `LAST30DAYS_SKIP_KEYCHAIN` opt-out, and the test sets it.

The switch is read from the **process environment only**, never from a config file — it gates a credential source consulted *while* the config is being assembled, so a file-sourced value would be read too late to take effect. That asymmetry with `LAST30DAYS_KEYCHAIN_ALIASES` is documented in `CONFIGURATION.md` and in the docstring.

Engine behaviour is unchanged when the switch is unset, and Linux CI is unaffected either way — the loader already no-ops off Darwin.

## Verification

Rather than assert the new tests pass, I checked they can fail:

- The guard test goes **red with the guard removed and the constant kept**, so it fails on behaviour rather than on a missing attribute (removing both gives an `AttributeError`, which is a pass for the wrong reason).
- A **counterexample** test asserts the key *is* read when the switch is absent, so `== {}` cannot pass just because something else went empty.
- `run.assert_not_called()` pins that the opt-out short-circuits *before* any lookup.
- A third test pins that falsy values (`"0"`, `""`, `"false"`, `"no"`) do **not** disable the source.
- `test_bare_run_emits_web_promo` now passes with the Keychain item still in place; full suite green on macOS (first clean run on this machine).

## Scope

Additive only — 85 insertions, no deletions. Independent of #976; happy to rebase if that lands first.


mvanhorn:
Closing as split.

The Keychain skip switch landed on main as #1090 (squash `41fcb4d`), credit @OSideMedia. Scrapling stays off public `main` — that hitchhiker belongs on the beta channel, as noted on this PR.

Merging this branch as-is would also have conflicted with #1087's `env.py` and brought Scrapling onto the public skill.

PR #1056, opened 2026-08-25
fix(render): collapse and defang scraped titles so they cannot forge engine sentinels

### LINKED ISSUE #1053: Scraped titles bypass untrusted-content escaping and can forge the PASS-THROUGH FOOTER sentinel

olliedigital:
## Summary

Scraped titles were interpolated raw at nine render sites while snippets went through `_format_untrusted_evidence`. Since `normalize.py` sets `title=text[:140]` for X, TikTok, Instagram and LinkedIn and only strips the ends, a short post kept its internal newlines and could emit lines closing the `EVIDENCE FOR SYNTHESIS` envelope and opening a `PASS-THROUGH FOOTER` block  -  text LAW 5 tells the host model to relay to the user verbatim. Titles are now collapsed to one line and the sentinel markers are defanged, in titles and in snippet/comment evidence.

## Testing

- [x] `uv run pytest`

`tests/test_render_title_defang.py`  -  7 cases. The main one drives a real `render_compact()` with a forged title and asserts each envelope opens and closes exactly once and the payload never lands inside the real footer. I confirmed the test fails on unmodified `render.py` (6 of 7 fail, including the end-to-end case) and passes with the fix, so it is genuine regression coverage rather than a test that passes either way.

Full suite is green except `tests/test_setup_wizard.py::TestWriteApiKey::test_unwritable_target_returns_false`, which also fails on unmodified `main` in my environment  -  it asserts a `chmod`-denied write fails, and I ran as root, where it succeeds. Unrelated to this change.

## Changelog

- [x] Added `changelog.d/1053.security.md`
- [ ] Skip changelog

### AI review

Found during an AI-assisted source review of v3.21.1 before installing the skill, verified by hand against the source. The review checked for credential exfiltration, telemetry, autostart persistence, and prompt-injection surface; it found no exfiltration and no telemetry, and this was one of two injection findings (the other is #1054).

Worth saying plainly: the injection posture here is better than most tools in this category. `UNTRUSTED_CONTENT_NOTICE`, `_AI_SAFETY_NOTE` on every render path, `_format_untrusted_evidence`, and the URL sanitization are real, test-covered defenses. This is a field that missed an existing control, not a missing control  -  titles simply never got the treatment snippets got.

Two risks checked on the fix itself:

- **Losing what the post said.** Defanging breaks the marker syntax but preserves the visible text, so the reader still sees the content (`test_title_text_is_preserved_for_the_reader`).
- **Mangling ordinary titles.** A normal title is unchanged (`test_ordinary_titles_are_unchanged`). A post that genuinely discusses HTML comment syntax will render with the delimiters broken; that seemed the right trade against a forged footer, and it matches the trade `_defang_corpus_sentinels` already makes.

### Security

Input handling and output rendering only. No command execution, path handling, auth, secrets, or dependency changes. No new dependencies.

Note on why escaping alone was not enough for snippets: `_format_untrusted_evidence` indents continuation lines, which stops CommonMark from parsing a forged heading, but does nothing to an HTML comment a model is reading as text. That is why `_defang_engine_sentinels` is applied there as well, not only to titles.

## Notes

`_safe_title` collapses whitespace the same way `_one_line` already does in `discovery_handoff.py:911` before fencing  -  that path had it right, the compact/brief render paths just did not use it. I kept the helper local to `render.py` rather than moving `_one_line` into a shared module, to keep the diff small; say the word if you would prefer it centralized.

Breaking the comment delimiters is what actually neutralizes the sentinels. The phrase substitutions (`PASS-THROUGH FOOTER`, `EVIDENCE FOR SYNTHESIS`) are defense in depth for a model that pattern-matches on the wording rather than the comment syntax.

- [x] None

## Related issues

Fixes #1053. Relates to #1054 (same review, separate PR).


olliedigital:
The private-corpus finding is correct and I'd missed it — fixed in c47fb78.

`_render_corpus_section` applied only `_defang_corpus_sentinels`, which handles the `LAST30DAYS_PRIVATE_CORPUS` marker and nothing else, while corpus titles, filenames and snippets all render inside the EVIDENCE FOR SYNTHESIS envelope. Corpus titles also weren't newline-collapsed. I confirmed it before fixing: an unpatched render of a forged corpus title emits **two** `<!-- END EVIDENCE FOR SYNTHESIS -->` markers, so the envelope really was closable from a local file.

Corpus evidence now goes through `_safe_title` (title, filename) and `_format_untrusted_evidence` (snippet) on top of the corpus-marker defanging — the same treatment every other path inside the envelope gets.

Worth stating the threat model explicitly, since "private corpus" sounds trusted by definition: a file on disk isn't automatically trustworthy input. It may have been downloaded, shared by someone else, or machine-generated. Treating it as untrusted-by-default costs nothing here.

Added `CorpusDefangTest` with three cases (forged sentinels in title/snippet, in the filename, and a guard that corpus-marker defanging still works). I verified the two meaningful ones fail without the fix — the filename case initially passed either way, so I tightened it to assert the exact forged marker is absent rather than counting the engine's own.

Also corrected `changelog.d/1053.security.md`, which overstated coverage exactly as you noted.

mvanhorn:
Speaking as Matt's firstmate:

Captain OK via Firstmate: merging #1056 as security-ok. CI green, Greptile 5/5 on current head.

PR #1060, opened 2026-08-27
fix(security): refuse cleartext provider endpoint overrides

Comradery64:
## Summary

`OPENAI_BASE_URL`, `XAI_BASE_URL`, and `OPENROUTER_BASE_URL` redirect the request that carries the provider's `Authorization: Bearer` header, so they are a credential boundary rather than a plain URL setting. They were honored verbatim with no scheme check, so an `http://` override to a remote host put the API key on the wire in cleartext. All three now resolve through `providers.base_url_override()`, which accepts `https://` anywhere and `http://` only on loopback, and otherwise warns to stderr and falls back to the vendor endpoint.

Separately, `OPENROUTER_BASE_URL` was propagated into `os.environ` by `_propagate_config_to_environ` but was missing from `permission_preflight.ENDPOINT_OVERRIDE_KEYS`, so an OpenRouter redirect was invisible in the `--preflight` summary that exists to surface exactly that. Added.

## Testing

- [x] `uv run pytest`

`tests/test_provider_base_url_guard.py`, 18 cases: unset falls back to the vendor URL; `https` honored; refused for `http://attacker.example`, `http://10.0.0.5:8080`, `ftp://`, `//host`, and a bare `host/v1`; loopback allowed for `localhost`, `127.0.0.1`, `127.0.0.2`, `[::1]`; and a parametrized check that every propagated override is present in `ENDPOINT_OVERRIDE_KEYS`, so the OpenRouter gap cannot silently return.

Full suite: 4087 passed, 5 skipped, 53 subtests passed.

## Changelog

`changelog.d/+provider-endpoint-override-guard.security.md`. No `CHANGELOG.md` edit, no version/manifest bump.

### AI review

Found while auditing this skill's credential boundaries with Claude Code, prompted by the public Gen Agent Trust Hub and Snyk audit results (both dated 2026-05-18, both HIGH). Most of what those reports flagged did not hold up against the current tree, and I want to be straightforward about that since it is the context for this PR:

- No credential exfiltration found. `auth_token`/`ct0` reach `x.com` and nowhere else. `hosted.py` sends only the query and has no default endpoint. `--publish-html` is explicit opt-in. The watchlist webhook is https-only and sends topic plus counts.
- No command-execution risk found. Zero `shell=True`, `os.system`, `eval`, or `exec` in the tree; everything goes through argv lists in `subproc.py`. The yt-dlp SSH route is regex-validated with `shlex.quote` and a `--` terminator.
- The cookie path is consent-gated, default-off (`FROM_BROWSER` unset yields an empty browser list), gated per-command by `ConfigLoadPolicy`, and temp copies are locked to 0600.
- I also read `scripts/lib/vendor/bird-search/` directly, since `.skillignore` excludes it from install-time scanning: only `x.com`/`twitter.com`, zero npm dependencies, no `child_process`, no `eval`. Clean.

What did hold up is this PR's subject, plus a separate `setup-keychain.sh` issue sent as its own PR.

### Security

Endpoint overrides, credential handling in request headers, and the `--preflight` permission report. The guard sits at read time rather than at `.env` load, so it covers both ways a value arrives: the process environment and `_propagate_config_to_environ`. Loopback `http://` is deliberately still allowed so local LiteLLM/Ollama gateways and SSH tunnels keep working; nothing leaves the machine in that case. The warning names the variable and the rejected value but never the credential.

Follow-up filed separately as an issue, not fixed here: `lib/http.py` uses plain `urllib.request.urlopen` with no redirect policy, and CPython's `HTTPRedirectHandler` forwards `Authorization`/`x-api-key` across an origin change. That fix belongs in the shared choke point for every source, so it seemed better as your call than as an uninvited patch.

## Notes

`CONFIGURATION.md` gains a "Provider endpoint overrides" section. These three variables were previously undocumented despite being able to redirect an API key. `test_env_doc_contract.py` is one-directional (documented is a subset of registered) and all three are registered in `get_config`, so it stays green.

- [x] None

## Related issues

N/A


mvanhorn:
Speaking as Matt's firstmate: classified as security (refuse cleartext `*_BASE_URL` provider endpoint overrides; credential-boundary guard). Greptile 2/5 (URL-embedded secrets in diagnostics; OpenRouter preflight vs runtime). mergeable=CONFLICTING / DIRTY. Flagging Firstmate/captain — do not auto-merge; do not stale-close while this security flag is owed. Author should address Greptile + rebase, but captain security review is required before any merge ask.

PR #1061, opened 2026-08-27
fix(security): keep keychain secrets out of argv and presence checks

Comradery64:
## Summary

Two secret-hygiene issues in `scripts/setup-keychain.sh`:

1. `security add-generic-password ... -w "$value"` put the credential on the command line, where `ps` exposes it to every user on the machine for the lifetime of the call. The value is now piped to `-w` on stdin instead. With no argument, `security` prompts for the value and then a retype, which is the form `man security` itself recommends ("Put at end of command to be prompted (recommended)"), so the value goes down the pipe twice and the prompts are discarded.
2. `find-generic-password ... -w` was used for presence checks in both `--list` and the prompt loop, asking Keychain to decrypt a secret only to discard it. In the prompt loop it was worse: the plaintext was assigned to a shell variable purely to test emptiness, where a stray `set -x` or error trace could surface it. Exit status already answers the question (0 found, 44 missing), so `-w` is dropped and both sites branch on that.

Also `unset value` after each store, so the plaintext does not outlive the iteration that needed it.

## Testing

- [x] `uv run pytest`

Two static guards appended to `tests/test_env_keychain.py`, next to the existing `ALL_KEYS` drift guard: one rejects a literal argument after `add-generic-password ... -w`, one rejects `find-generic-password ... -w`. Both parse the script with comments stripped, so prose cannot satisfy them, and both were verified to fail against the pre-patch script via `git show HEAD:...` rather than only passing after the change.

These are static assertions because there is no runtime path to exercise without writing to the developer's real Keychain. Behaviour was instead verified manually against an isolated keychain prefix in a temp copy of the script, never the real `last30days-` namespace: add, skip-when-set, `--replace` overwrite, empty-input skip, `--list`, and `--delete` all behave as before, and the stored value round-trips byte-for-byte.

Full suite: 4069 passed, 5 skipped, 53 subtests passed.

## Changelog

`changelog.d/+keychain-secret-hygiene.security.md`. No `CHANGELOG.md` edit, no version/manifest bump.

### AI review

Found with Claude Code while auditing this skill's credential boundaries. I originally flagged only the `add-generic-password` line; the `find-generic-password -w` presence checks turned up on a second pass looking for the same pattern elsewhere in the file, which is why this PR fixes two sites rather than one.

`ALL_KEYS` is untouched, so `test_keychain_keys_match_setup_script` is unaffected.

### Security

Credential handling and process-listing exposure. No behaviour change to what gets stored or where. The retype-prompt requirement was confirmed empirically before relying on it: a single-line pipe fails with "passwords don't match", two lines store correctly, and the stored value was verified by byte count without printing it.

Worth noting for context, since the public third-party audits of this skill flag Keychain use as credential harvesting: this script is the mitigation, not the risk. It is what lets a user keep keys out of a plaintext `~/.config/last30days/.env`, and the loader is scoped to the `last30days-` prefix. This PR just closes the two places where the helper itself was careless with the secret it was given.

## Notes

The `--list` header string still reads "Stored last30days-* keychain items:" with the prefix hardcoded in the echo rather than interpolated from `$PREFIX`. Cosmetic and left alone to keep this PR to the security change.

- [x] None

## Related issues

N/A


mvanhorn:
Thanks @Comradery64. Your safe presence-check and exit-status/counting hunks from `6ef5a3e5697a345a349123859f73e8737a568752` are now selectively ported and merged in #1111, with co-author credit. The port adds multiline guards and stubbed add/skip/replace tests; full CI passed before merge.

The piped-password writer was deliberately omitted. `security add-generic-password -w` without a value uses a prompt backed by `getpass`, which prefers `/dev/tty`; piping two copies is not evidence that an interactive run will consume stdin rather than reprompt. We kept the existing writer rather than introducing that uncertainty.

The original command-line exposure remains unresolved and is explicitly documented in #1111. A secure writer needs a separate design and isolated interactive/noninteractive macOS verification. Closing this original after retaining the useful hunks, not claiming that the full secret-hygiene concern is fixed.

Comradery64:
You were right to drop the writer, and it's worse than "may reprompt" — I went back and tested it under a real pty and it **hangs and stores nothing**:

```bash
timeout 20 script -q /dev/null bash -c \
  "printf 'DUMMYVAL\nDUMMYVAL\n' | security add-generic-password -U -a \"\$USER\" -s l30d-pty-test -w"
```

```
exit=124
^D^H^Hpassword data for new item: [hangs until timeout]
NO ITEM STORED
```

With a controlling terminal present, getpass opens `/dev/tty` and never looks at the pipe. My original verification ran with no tty anywhere in the chain, so getpass fell back to stdin and the two-line pipe appeared to work — a test that couldn't fail in the way that mattered, on the script's primary code path. Thanks for not taking it on faith; I'd have shipped a hang.

For whatever it's worth toward the open item in #1111, one direction does avoid argv without new machinery: drop the script's own `read -rs` and let `security` own the prompt, so the secret goes tty → `security` and never through a shell variable or an argument. Verified interactively via `pty.fork()`, answering each prompt as a human would:

```
prompts answered: 2
transcript: password data for new item:  | retype password for new item:  |
stored bytes: 9   # DUMMYVAL + newline, byte-exact
```

Real costs, so you can weigh it rather than take my word twice in one thread:

1. The user types each value twice — `security` always asks for a retype and there's no flag to suppress it.
2. The script loses the empty-input-means-skip behaviour, since it no longer reads the value itself. That needs a y/n gate per key before handing off, which is a UX change, not a drop-in.
3. `security`'s own prompt text replaces the aligned `%-28s` column, so the output shape changes.

Given (2), it's a redesign of the prompt loop rather than a patch, which matches your read that this needs its own design and isolated interactive/noninteractive verification. Happy to send it as a PR if the tradeoff looks acceptable, or to leave it alone — no expectation either way. Thanks for porting the presence-check hunks and for the co-author credit on #1111.

PR #1064, opened 2026-08-27
fix(bluesky): return actionable error after refresh retry fails

huyua9:
## Summary

- replace the internal `refresh` sentinel with an actionable error after a second Bluesky `401 Unauthorized`
- add regression coverage for old token / 401 / refreshed token / 401
- add a changelog fragment

## Validation

- `uv run pytest tests/test_bluesky.py` — 47 passed
- `uv run pytest --cov --cov-report=term-missing` — 4092 passed, 6 skipped, 88.92% coverage
- `git diff --check`

Generated with Codex.

### DIFF
--- a/changelog.d/6037.fixed.md
+++ b/changelog.d/6037.fixed.md
@@ -1,0 +1,1 @@
+- Return an actionable error when Bluesky search still receives `401 Unauthorized` after the one-time session refresh.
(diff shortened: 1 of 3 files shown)


mvanhorn:
Thanks @huyua9. The useful repeated-401 error behavior is now covered in the corrected and merged #1065: both successful refresh followed by another 401, and fresh-login recovery followed by another 401, return actionable credential guidance without a literal refresh sentinel or an unbounded loop. The broader fix also corrects Bearer refresh transport and missing-token recovery. Closing this overlapping implementation now that its regression is preserved and the replacement has landed.

PR #1065, opened 2026-08-27
fix(bluesky): use refresh tokens for expired sessions

huyua9:
## Summary

- use the cached Bluesky `refreshJwt` to recover an expired access token after a search `401`
- fall back to `createSession` only when the refresh token is rejected with `400` or `401`
- keep refresh and re-authentication bounded, with actionable errors instead of an internal sentinel

## Testing

- [x] `uv run pytest tests/test_bluesky.py -q`
- [x] `uv run pytest --cov --cov-report=term-missing` — 4093 passed, 6 skipped, 88.92% coverage
- [x] `git diff --check`
- [x] Added regression coverage for refresh success, invalid-refresh fallback, and bounded unauthorized retry

## Changelog

- [x] Added `changelog.d/6034.fixed.md`

## Agent disclosure

The change was reviewed against the current `main` implementation and the merged Bluesky session history in PRs #182 and #202. The implementation keeps the existing one-retry boundary, never logs token values, and uses dummy credentials only in tests.

- [x] None
- [ ] Yes — disclosure:

## Related issues

N/A


mvanhorn:
Maintainer amendment preserves your refresh work and the useful repeated-401 regression from your #1064. Corrected refresh transport to Bearer authorization, restored one fresh login when refresh credentials are missing/rejected, disabled automatic HTTP retries for refresh, and kept transient failures from echoing token-bearing error text. All 56 Bluesky tests pass, including both second-401 paths and malformed/rotated token cache cases. ce-code-review complete (latest20-1065), sequential inline, no actionable findings; no real sessions touched. Maintainer should inspect the first expired-session run after merge for one bounded recovery and actionable failures; revert on repeated login loops or fresh authentication regressions. #1064 stays open until this lands.

PR #1068, opened 2026-08-28
fix(cookies): parse Windows Firefox profiles.ini as UTF-16

### LINKED ISSUE #1067: Windows: FROM_BROWSER X auth always fails — profiles.ini is UTF-16, UnicodeDecodeError escapes the except tuple
## Summary

On Windows, X/Twitter cookie auth via `FROM_BROWSER` always fails when the user's Firefox is the cookie source, even when they are fully logged in to x.com. Firefox on Windows writes `profiles.ini` as **UTF-16 LE with a BOM**, but `_find_default_profile()` reads it with a hardcoded `encoding="utf-8"`. The resulting `UnicodeDecodeError` is not in the `except` tuple, so it propagates out of `extract_firefox_cookies()` and the documented `_fallback_find_profile()` path never runs. The whole Firefox cookie route dies on one unhandled exception.

Same class as #714 and #549 (Windows encoding assumptions), in a file those didn't cover. Related to #498, which fixed profile *selection* in this function but not the parse.

## Environment

- last30days v3.21.1
- Windows 10 Pro 19045, Python 3.12.0
- Firefox with multiple profiles; `profiles.ini` is UTF-16 LE (`ff fe` BOM)
- `FROM_BROWSER=auto` and `FROM_BROWSER=firefox` both affected

## Reproduction

1. On Windows, be logged in to x.com in Firefox.
2. Set `FROM_BROWSER=firefox` in `~/.config/last30days/.env`.
3. Run an X-only research pass:

```
python skills/last30days/scripts/last30days.py "any topic" --search x
```

Actual:

```
[Resolve] No web search backend available, skipping resolve
✗ Error: No sources are available for this run.
RuntimeError: No sources are available for this run.
```

Calling the extractor directly shows the real error, which the pipeline swallows into "no sources":

```python
from lib import cookie_extract as ce
ce._find_default_profile(ce._get_firefox_profiles_dir())
# UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
ce.extract_firefox_cookies("x.com", ["auth_token", "ct0"])
# UnicodeDecodeError: same
```

Confirming the file encoding:

```
$ head -c 16 "C:\Users\<user>\AppData\Roaming\Mozilla\Firefox\profiles.ini" | xxd
00000000: fffe 5b00 4900 6e00 7300 7400 6100 6c00  ..[.I.n.s.t.a.l.
```

Decoded as UTF-16, the file is well-formed and points directly at the correct profile:

```
[Install308046B0AF4A39CB]
Default=Profiles/en3ndvop.default-release
Locked=1
```

That profile holds a valid `auth_token`, `ct0` and `twid`. Nothing is wrong with the user's session; the ini is simply never parsed.

## Root cause

`skills/last30days/scripts/lib/cookie_extract.py`

- **L100** — `config.read(str(ini_path), encoding="utf-8")` assumes UTF-8. Firefox on Windows writes UTF-16 LE.
- **L121** — `except (configparser.Error, OSError) as exc:` does not include `UnicodeDecodeError` (a `ValueError`), so the error escapes instead of degrading to `_fallback_find_profile()`.

The second point is what makes it fatal rather than merely suboptimal. With the except widened alone, the fallback scan would find a usable profile.

## Suggested fix

```diff
             config = configparser.ConfigParser()
-            config.read(str(ini_path), encoding="utf-8")
+            try:
+                config.read(str(ini_path), encoding="utf-8")
+            except UnicodeDecodeError:
+                # Firefox on Windows writes profiles.ini as UTF-16 LE with a BOM.
+                config.read_string(ini_path.read_text(encoding="utf-16"))
```

```diff
-        except (configparser.Error, OSError) as exc:
+        except (configparser.Error, OSError, UnicodeDecodeError) as exc:
             logger.debug("Failed to parse profiles.ini: %s", exc)
```

Verified locally against v3.21.1. After the patch:

```
default profile -> ...\Firefox\Profiles\en3ndvop.default-release
x.com: FOUND ['auth_token', 'ct0']
✓ Research complete (2.8s) - X: 5 posts
```

## Impact

Any Windows user whose Firefox writes a UTF-16 `profiles.ini` cannot use `FROM_BROWSER` for X at all, which is the only free, keyless X route. It also fails quietly: `--diagnose` reports `x` in `available_sources` with `x_pending_browser_auth: true`, because it runs in `plan_only` and skips cookie reads by design (#692). The user is told X is configured, then every X-only run dies with a generic "No sources are available for this run." Surfacing the decode failure in the run log would have made this a one-minute diagnosis instead of a code read.

Happy to open a PR with the above plus a UTF-16 `profiles.ini` fixture if useful.

khsaurabh:
## Summary

Firefox on Windows writes `profiles.ini` as UTF-16 LE. `_find_default_profile()` read it as UTF-8, `UnicodeDecodeError` was not in the `except` tuple, and `FROM_BROWSER` X auth died before the documented profile fallback.

Retry UTF-16 on decode failure so the Install/Default profile is used, and treat leftover decode errors as malformed-ini (fallback), not a crash.

## Testing

- [x] `uv run pytest`
- [x] Added `test_utf16_profiles_ini_uses_install_default` (UTF-16 fixture; decoy profile would win if we only widened the except)
- Sabotage: dropping the UTF-16 retry makes that test fail (`decoy_token` vs `utf16_token`)

## Changelog

- [x] Added `changelog.d/1067.fixed.md`

### AI review

Checked encoding fallback vs silent fallback to `sorted()` first profile, sibling `encoding="utf-8"` sites (only this one), and that cookie DB handling is unchanged. Independent review subagent was not dispatched for this small parse change.

### Security

Cookie files are still read only after a valid profile path; this change only how `profiles.ini` is decoded. No new network, subprocess, or secret logging. N/A beyond that.

## Notes

Does not change Chrome/Chromium cookie paths. Does not surface the decode failure in `--diagnose` (called out on the issue; out of scope).

- [x] None
- [ ] Yes — disclosure:

## Related issues

Fixes #1067

### DIFF
--- a/changelog.d/1067.fixed.md
+++ b/changelog.d/1067.fixed.md
@@ -1,0 +1,1 @@
+Windows Firefox `FROM_BROWSER` X auth no longer dies on a UTF-16 `profiles.ini` (`UnicodeDecodeError` used to skip the fallback and kill the only keyless X route).
--- a/skills/last30days/scripts/lib/cookie_extract.py
+++ b/skills/last30days/scripts/lib/cookie_extract.py
@@ -86,6 +86,16 @@ def _get_firefox_profiles_dir() -> Optional[Path]:
     return path if path.is_dir() else None
 
 
+def _load_profiles_ini(ini_path: Path) -> configparser.ConfigParser:
+    """Parse Firefox profiles.ini, retrying UTF-16 LE used on Windows (#1067)."""
+    config = configparser.ConfigParser()
+    try:
+        config.read(str(ini_path), encoding="utf-8")
+    except UnicodeDecodeError:
+        config.read(str(ini_path), encoding="utf-16")
+    return config
+
+
 def _find_default_profile(profiles_dir: Path) -> Optional[Path]:
     """Parse profiles.ini to find the default profile directory.
 
@@ -96,8 +106,7 @@ def _find_default_profile(profiles_dir: Path) -> Optional[Path]:
 
     if ini_path.is_file():
         try:
-            config = configparser.ConfigParser()
-            config.read(str(ini_path), encoding="utf-8")
+            config = _load_profiles_ini(ini_path)
 
             # First pass: Install* section (Firefox >= 67 format, takes priority)
             for section in config.sections():
@@ -118,7 +127,7 @@ def _find_default_profile(profiles_dir: Path) -> Optional[Path]:
                     resolved = _resolve_profile_path(profiles_dir, config, section)
                     if resolved and resolved.is_dir():
                         return resolved
-        except (configparser.Error, OSError) as exc:
+        except (configparser.Error, OSError, UnicodeDecodeError) as exc:
             logger.debug("Failed to parse profiles.ini: %s", exc)
 
     # Fallback: scan directory for anything that looks like a profile
--- a/tests/test_cookie_extract.py
+++ b/tests/test_cookie_extract.py
@@ -281,6 +281,39 @@ class TestExtractFirefoxCookies:
         assert result is not None
         assert result["auth_token"] == "fallback_token"
 
+    def test_utf16_profiles_ini_uses_install_default(self, mock_firefox_env):
+        """Firefox on Windows writes UTF-16 LE profiles.ini (#1067)."""
+        default_profile = "en3ndvop.default-release"
+        decoy = "aaa111.decoy"
+        profiles_dir = mock_firefox_env(
+            default_profile=default_profile,
+            profiles={
+                decoy: [
+                    (".x.com", "auth_token", "decoy_token"),
+                ],
+                default_profile: [
+                    (".x.com", "auth_token", "utf16_token"),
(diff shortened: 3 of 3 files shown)


mvanhorn:
Merged as `40b486071f8a76f6324d12c648263b1077092530` — thanks @saurabhkagent-lab!

Verified against current `main` before merging: the UTF-8-only `config.read` in `_find_default_profile` was still there, so a UTF-16 LE `profiles.ini` (as Windows Firefox writes it) raised `UnicodeDecodeError` outside the `except` tuple and killed `FROM_BROWSER` before the profile-scan fallback could run. The UTF-8 → UTF-16 retry plus adding `UnicodeDecodeError` to the malformed-ini fallback is exactly right, and the fix stays scoped to Firefox `profiles.ini` parsing — Chrome/Keychain paths untouched. The new UTF-16 fixture test (with the decoy profile) passes on top of current main. Fixes #1067.

Issue #986, opened 2026-08-13
Instagram source always returns HTTP 404 while ScrapeCreators endpoints return 200 (v3.18.4)


ecik666:
## Summary

The Instagram source **always fails with `HTTP 404: Not Found`** (`Instagram: 0 items (error: HTTPError: HTTP 404: Not Found)`), even though the exact ScrapeCreators endpoints the engine uses return **HTTP 200 with valid data** when called directly with the same API key.

- Plugin version: **3.18.4**
- Host: Claude Code on macOS (Python 3.14)
- ScrapeCreators key: valid, plenty of credits remaining (confirmed via direct calls)
- Reproduced on multiple runs with different topics/handles

## Repro

Run any research with an Instagram creator + keyword, e.g.:

```
last30days.py "Prohern tattoo brand" --plan <plan> \
  --ig-creators=prohern_official --tiktok-hashtags=prohern
```

Engine output (every run):

```
## Source Coverage
- Instagram: 0 items (error: HTTPError: HTTP 404: Not Found (run doctor for fixes))

## Source Errors
- Instagram: HTTPError: HTTP 404: Not Found
```

`doctor --postmortem` only prints `instagram — error: HTTPError: HTTP 404: Not Found` with `fix: doctor` (a self-referential loop, no actionable fix).

## Evidence the endpoints themselves are fine (HTTP 200)

Both Instagram endpoints referenced in `scripts/lib/instagram.py` return **200** with real data using the *same* `SCRAPECREATORS_API_KEY`:

**1. `/v1/instagram/user/reels` (used by `--ig-creators`)**
```
GET https://api.scrapecreators.com/v1/instagram/user/reels?handle=prohern_official
→ HTTP 200
{"success":true,"credits_charged":1,"items":[{"media":{...}}, ...]}   # 12 reels returned
```

**2. `/v2/instagram/reels/search` (keyword search)**
```
GET https://api.scrapecreators.com/v2/instagram/reels/search?query=prohern
→ HTTP 200
{"success":true,"reels":[{"shortcode":"...","caption":"Prohern Black – the black that doesn't fade...", ...}]}
```

For contrast, the profile endpoint also confirms the account is public and exists:
```
GET https://api.scrapecreators.com/v1/instagram/profile?handle=prohern_official → HTTP 200 (is_private:false)
```

[...]

PR #1071, opened 2026-08-29
fix(reddit): prioritize most-commented threads for enrichment slots

### LINKED ISSUE #906: Comment-enrichment slots follow relevance-dominant result order, so the highest-comment thread can miss enrichment

SomSamantray:
## Summary

Keyless Reddit comment-enrichment slots (`ENRICH_LIMITS` 3/5/8 per depth) now go to the most-commented threads within each entity-priority tier. Previously the slots were allocated in score-first order, so a high-discussion thread ranked low by score could miss `## Top Community Comments` coverage entirely while near-empty threads consumed the scarce slots (reported in #906 with a 45-comment thread ranked 7th missing every slot). The entity-match tier still claims slots first, ties keep the stable score-first order, and the final display sort key (`_relevance_rank_key`) is untouched, so user-visible ranking semantics are unchanged.

## Testing

- [x] `uv run pytest` — 4097 passed, 6 skipped, 66 subtests passed
- [x] Added 6 regression tests (red observed on the pre-fix code, then green) covering: highest-comment thread gets a slot at default depth, comment-count ordering within the match tier, entity-tier priority over comment counts, stable ties, unknown counts treated as 0, and leftover slots in the miss tier ordered by comments
- [x] Coverage gate held: `uv run pytest --cov` — TOTAL 88.93% (fail_under 84%)

## Changelog

- [x] Added `changelog.d/906.fixed.md` (`fixed` type)
- [ ] Skip changelog

### AI review

This change was reviewed by a multi-lens agent review (correctness, project-standards, testing, learnings, adversarial). Main risks checked: entity-tier priority preservation, stable tie-break behavior, the `or 0` None guard on comment counts, and display-order invariance. The review flagged one test gap (miss-tier ordering when the entity-match tier does not fill the slot budget); that finding was independently validated and the regression test was added in the same branch. Advisory notes carried forward: slot ordering uses discovery-time comment counts (inherent to the signal), and comment-first ordering front-loads slower comment fetches into the existing 45s enrichment budget.

### Security

N/A — no input handling, command execution, path handling, auth, secrets, or dependency changes. Tests use dummy values only.

## Notes

Entity-miss threads with many comments still lose slots to entity matches with few comments; that is the intended entity-priority contract, not a regression. Follow-up work is tracked in #986/#1020 (Instagram) separately — unrelated to this change.

- [x] None
- [ ] Yes — disclosure: 

## Related issues

Fixes #906

### DIFF
--- a/changelog.d/906.fixed.md
+++ b/changelog.d/906.fixed.md
@@ -1,0 +1,1 @@
+Keyless Reddit comment-enrichment slots now go to the most-commented threads within each entity-priority tier, so a high-discussion thread no longer misses `## Top Community Comments` coverage while near-empty threads consume the scarce slots ([#906](https://github.com/mvanhorn/last30days-skill/issues/906)).
--- a/skills/last30days/scripts/lib/reddit_keyless.py
+++ b/skills/last30days/scripts/lib/reddit_keyless.py
@@ -256,6 +256,21 @@ def _enrich(posts: List[Dict[str, Any]], depth: str) -> List[Dict[str, Any]]:
     return enriched + rest
 
 
+def _by_comments(posts: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """Stable-sort posts by comment count descending for enrichment slots.
+
+    Ties (equal counts, including unknown counts treated as 0) preserve the
+    incoming order, which search_and_enrich's provisional score-first sort
+    establishes. Mirrors _relevance_rank_key's `or 0` guard so a present-but-
+    None count is treated as 0 rather than raising.
+    """
+    def _comment_count(post: Dict[str, Any]) -> int:
+        eng = post.get("engagement") or {}
+        return eng.get("num_comments") or post.get("num_comments") or 0
+
+    return sorted(posts, key=_comment_count, reverse=True)
+
+
 def _slot_priority(topic: str, posts: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
     """Order posts for enrichment slots: entity-matching posts first.
 
@@ -268,9 +283,11 @@ def _slot_priority(topic: str, posts: List[Dict[str, Any]]) -> List[Dict[str, An
     post text) so slots go to posts likely to survive final ranking — keying
     on the same head token keeps the two paths from diverging. Falls back to
     token-overlap relevance when the topic yields no usable primary entity.
-    Within each tier the incoming
-    (score-first) order is preserved. Never raises; on any failure the
-    incoming order is returned unchanged.
+    Within each tier posts are ordered by comment count descending (stable:
+    equal or unknown counts preserve the incoming score-first order), so the
+    scarce slots go to the threads with the most discussion rather than to
+    near-empty threads that merely ranked higher by score. Never raises; on
+    any failure the incoming order is returned unchanged.
     """
     try:
         from . import relevance, rerank
@@ -292,7 +309,7 @@ def _slot_priority(topic: str, posts: List[Dict[str, Any]]) -> List[Dict[str, An
         misses: List[Dict[str, Any]] = []
         for post in posts:
             (matches if _matches(post) else misses).append(post)
-        return matches + misses
+        return _by_comments(matches) + _by_comments(misses)
     except Exception:
         return posts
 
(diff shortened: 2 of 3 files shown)


mvanhorn:
Thank you @SomSamantray. All three of your commits landed on main verbatim through #1083 (squash-merged with a Co-authored-by credit), including the miss-tier ordering test. The enrichment limits were raised to 4/8/12 in the same change, so two of your fixtures gained a few posts to keep exercising the cut; the ordering contract is unchanged. Closing this as superseded by #1083; #906 closes with it.

mvanhorn:
Landed via #1083.

PR #1077, opened 2026-08-30
fix: continue research after optional X skip

arc-web:
## Summary

Declining or skipping X/browser-cookie access no longer strands a supplied research topic. The skill continues with available sources, preserves explicit cookie consent, retains configured X when active, and reports an unavailable X lane only after useful results.

## Testing

- [ ] `uv run pytest`

Focused regression set: 275 passed. Full suite: 4,106 collected, 4,099 passed, 6 skipped, and 1 unrelated host-sensitive failure in `tests/test_backend_descriptors.py::TestGrokExpiryStates::test_no_grok_cli_is_missing`; this machine has Grok at `/opt/homebrew/bin/grok` while the subprocess PATH omits that directory, and the fixture expects Grok to be entirely absent. Python compilation and `git diff --check` passed.

## Changelog

- [x] Added `changelog.d/+x-skip-continuation.fixed.md`

### AI review

Codex and delegated agents checked consent handling, first-run fall-through, active-source planning, X-active preservation, post-result messaging, setup status copy, quality scoring, and genuine error-reporting paths. The review found stale X repair-prompt expectations in two existing test areas; those expectations were updated, and a report-shape cleanup removed an unnecessary new return field before the final focused run.

### Security

Browser-cookie reads remain gated by explicit consent. The no-cookie path passes `--no-browser-cookies`, does not add credential writes, and does not change command execution or dependency behavior. Added lines were checked for credential patterns.

## Notes

This does not remove or disable X. Configured X remains an active source. It changes only the behavior when X is optional and unavailable or declined: continue the supported research first, then report the omission without another setup prompt.

- [x] None
- [ ] Yes — disclosure:

## Related issues

N/A

### DIFF
--- a/changelog.d/+x-skip-continuation.fixed.md
+++ b/changelog.d/+x-skip-continuation.fixed.md
@@ -1,0 +1,1 @@
+Declining or skipping X/browser-cookie access now continues the requested research with available sources, and reports X only as an optional omission after useful results.
--- a/skills/last30days/SKILL.md
+++ b/skills/last30days/SKILL.md
@@ -534,7 +534,9 @@ When both `LAST30DAYS_API_KEY` and `LAST30DAYS_API_BASE` are set, the engine run
 
 ## Step 0: First-Run Setup Wizard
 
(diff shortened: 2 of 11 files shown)


mvanhorn:
Closing as superseded.

The continue-after-optional-X-skip behavior landed on main as #1089 (squash `dc25291`), rewritten onto the extras-host box-chrome NUX from #1087 so MacBook daily-driver last30days stays unchanged. Original idea and credit: @arc-web.

Merging this branch as-is would have skipped that NUX and conflicted with current `SKILL.md`.

PR #1034, opened 2026-08-20
fix(reddit): comment slots go to relevant threads, and bots stop taking them (conflict resolved)

nish3451:
## Summary

Resolves the merge conflict in #907 that left it DIRTY/CONFLICTING for 537h.

- **reddit_arctic.py**: kept HEAD's `datetime`/`timezone`/`Optional` imports (required by the listing-lane features that landed on `main` after #907 branched) alongside #907's `counts_verified` changes. HEAD already imported `Any`, so no import was lost.
- **test_reddit_keyless.py**: kept both sides — #907's new `TestSlotPriority` methods (entity-match tier ordering, verified-zero vs unknown-count, slot/display order parity) and HEAD's `TestScoredListingsFallback` class (arctic-shift listing fallback). The PR's methods belong inside `TestSlotPriority`; HEAD's class follows after.
- All other files (`reddit_keyless.py`, `reddit_listing.py`, `reddit_shreddit.py`, and their tests) auto-merged cleanly.

This PR contains the full content of #907 rebased onto current `main` with the two textual conflicts resolved. Once green it can replace #907.

#### Test plan
- [x] `uv run pytest tests/test_reddit_keyless.py` — 36 passed (local smoke)
- [x] `uv run pytest tests/test_reddit_arctic.py` — 21 passed (local smoke)
- [ ] CI full suite

Generated with [Devin](https://devin.ai)

PR #1083, opened 2026-08-31
fix(reddit): restore the keyless Reddit lane (comments, ranking, pacing)

### LINKED ISSUE #906: Comment-enrichment slots follow relevance-dominant result order, so the highest-comment thread can miss enrichment

### LINKED ISSUE #985: Reddit: keyless token bucket (5 req/sec) outpaces the unauthenticated surface — every run loses sub-requests to 429 with no in-lane retry

mvanhorn:
## Summary

A default keyless run now shows Reddit's most-discussed on-topic threads and their top-voted comments, finishes without dropping Reddit sub-requests to rate limits, and prints no error text in the user-facing footer. On the 2026-08-31 Kanye West run the engine had collected the 3,293-upvote "Vladimir Putin doesn't care about black people" comment and never rendered it; the 16K-upvote "Heads to Putin's Russia" thread never became a candidate at all.

Four independent mechanisms produced that, and each is fixed here:

- Fusion de-duplicated a candidate's source items by `(source, item_id)`, but keyless Reddit and X assign per-stream ids (`R1`, `X1`) that collide across subqueries, so the enriched copy of a thread was discarded. Identity is now the normalized URL and enrichment merges into the retained copy (`fusion.merge_source_items`, also applied at report finalize).
- Per-stream ranking is 65% title relevance and each stream is cut to 12 before pooling, so a 16K-upvote thread with 0.19 relevance lost to a one-upvote post. Reddit streams now hold slots for their top three on-topic threads by engagement, and the fused pool reserves 2/3/4 slots by depth for the same class (generic entity head tokens like "ai" require the higher 0.25 floor).
- `## Top Community Comments` and `## Best Takes` read only the eight visible clusters; they now read every cluster that clears the relevance floor.
- The keyless bucket ran at 5 req/s against a host that tolerates ~1 req/s, and every subquery re-fetched the same subreddit listings. Pacing lands via #1037 and #1048; a per-command memo with in-flight sharing cuts a four-subquery run from ~184 reddit.com requests to ~50; future timeouts are sized from the bucket's queue depth; enrichment covers 4/8/12 threads per depth with 12 comments each.

The emoji-tree footer carries counts only. Run diagnostics stay in `## Partial Coverage` (model-facing), the saved raw file, `--emit=json`, and `doctor --postmortem`, which now reads `Succeeded: reddit (22 items; 8 sub-requests rate-limited (HTTP 429))` instead of branding a source that delivered items as partial.

Contributor PRs landed as cherry-picks with authorship preserved: #1048 (immy1991), #1037 (terry-li-hm), #1071 (SomSamantray), #959 (technicianofthesacred), and the bot-author filter from #1034 (nish3451's rebase of b33eep's #907). The rest of #1034 (`counts_verified` slot rework) conflicts with #1071's comment-first ordering and is deferred; #1028's cooldown and old.reddit fallback and #961's arctic comment fallback are prior art for follow-up, not merged here.

## Testing

- [x] `uv run pytest` - full suite green except `tests/test_backend_descriptors.py::TestGrokExpiryStates::test_no_grok_cli_is_missing`, which fails identically on untouched `origin/main` on this machine (grok is installed at `/opt/homebrew/bin`); coverage 89.12% against the 84% floor
- [x] Added or updated tests that would catch a regression: `test_fusion_v3.py` (enriched-copy retention, URL collapse), `test_pipeline_finalize.py`, `test_top_comments_relevance.py` (cluster-eleven comment renders), `test_reddit_stream_keepers.py` (keepers, pool reservation, limit contract), `test_reddit_keyless_memo.py` (memo, in-flight sharing, owner-failure election, bounds), `test_reddit_keyless_wait_allowance.py`, `test_reddit_transport_outcomes.py` and `test_source_outcomes.py` (ok-plus-detail, rate-limited state carry-through), `test_render_footer.py` (counts-only footer, compact drops source errors), `test_reddit_ratelimit_reset.py` (epoch reset conversion, 60s cap), `test_reddit_shreddit.py` (bot filter, 4/8/12 limits), `test_registers.py` pins updated

Live validation, same flags and four-subquery plan as the failing 2026-08-31 run, residential IP:

| Measure | Before (main) | After |
|---|---|---|
| Reddit ranked candidates | 4 of 33 | 5 of 33 |
| Reddit comments in `## Top Community Comments` | 0 | 5 (Putin line leads) |
| 16K-upvote "Heads to Putin's Russia" thread | not a candidate | ranked, 12 comments attached |
| Feed futures dropped to timeouts | 6 | 0 |
| Footer Reddit line | `⚠ partial after 36 items: HTTP 429 ... (run doctor for fixes)` | counts only |
| Reddit outcome | `partial` | `ok` with `8 sub-requests rate-limited (HTTP 429)` detail |
| Run wall clock | 170-180s | 137s |

Two plan targets fell short and are recorded honestly: 5 Reddit candidates against a target of 6, and 8 unrecovered 429s against a target of 5 or fewer.

## Changelog

- [x] Added `changelog.d/` fragments: `906.fixed.md`, `985.fixed.md`, `+reddit-ratelimit-reset.fixed.md`, `+reddit-bot-comments.fixed.md`, `+source-outcome-swallowed-lane-failure.fixed.md`, `+reddit-ok-with-lane-detail.fixed.md`, `+fusion-keeps-enriched-copy.fixed.md`, `+top-comments-all-clusters.fixed.md`, `+reddit-engagement-floor.added.md`, `+reddit-run-memo.fixed.md`, `+reddit-enrich-limits.changed.md`, `+quiet-footer.changed.md`, `+reddit-queue-aware-timeouts.fixed.md`, `+reddit-review-hardening.fixed.md`
- [ ] Skip changelog

### AI review

Planned with a document review (coherence, feasibility, scope, adversarial lenses) and reviewed with a ten-finding code review before shipping. Findings fixed in the last commit: an uncapped `x-ratelimit-reset` wait that would sleep a GitHub 429 (epoch timestamp) for 1.7e9 seconds and a Reddit 429 for 540s (now capped at 60s, epoch converted to delta); memo waiters stampeding when the owner's fetch failed (waiters now re-elect one owner); the memo gate wait and enrichment budget ignoring the shared bucket's queue; the Reddit pool reservation silently cut by the final slice on small pools; stream keepers exceeding `--max-per-source`; a swallowed-429 source being re-fetched by the thin-source retry and reported as `no-results` when filtering emptied it. One review suggestion was tried and reverted on live evidence: letting the lane retry sleep out the header wait inside a worker dropped 14 feeds versus 6, so keyless 429s fail fast and the lane retries once through the bucket.

### Security

No credential, auth, or path handling changed. New parsing of untrusted input is limited to two response headers (`Retry-After`, `x-ratelimit-reset`) coerced through `float()` in a `try/except` with a 60-second cap, so a hostile header cannot produce an unbounded sleep. The per-command memo holds fetched HTML bodies in process memory only (bounded at 512 entries) and is never written to disk. No dependency changes.

## Notes

Known residuals from review, deliberately not addressed here:

- `fusion.raw_engagement` duplicates the shape of `rerank.discovery_engagement_total`; a shared helper is a small follow-up.
- `render._format_outcome` sniffs substrings of the detail string; fine for doctor output, worth a typed field later.
- The Reddit-only reservations sit on top of engagement-blind generic truncation. The plan chose bounded Reddit floors over re-weighting every source's score, which would need eval-baseline renegotiation.
- Deferred contributor work: #1034's `counts_verified` provenance, #1028's host cooldown and old.reddit fallback, #961's arctic comment fallback, #976's Scrapling fallback.
- `SKILL.md` pre-flight subreddit choice was the other half of the weak Kanye run (r/popculturechat and r/Fauxmoi returned nothing; r/NAFO and r/entertainment carried the story by accident). Out of scope for the engine.

- Run `/last30days` on a busy person topic with a dedicated subreddit and check the compact stdout: `## Top Community Comments` should carry Reddit comments, the footer Reddit line should have no `⚠`, and `doctor --postmortem` should list Reddit under Succeeded with any rate-limit detail.
- Watch stderr for `[RedditRSS] feed future failed:` / `[RedditListing] listing future failed:` (empty message = future timeout). Healthy: zero. Trigger to revisit `REDDIT_KEYLESS_CONTENTION_SECONDS` or `LAST30DAYS_REDDIT_KEYLESS_RATE`: any run with more than two.
- Watch the `sub-requests rate-limited` count in `doctor --postmortem`. Healthy: under ten on a four-subquery default run. Trigger to lower the default rate: sustained counts above twenty.
- Validation window: the next ten real runs after merge; owner: repo maintainer.

- [x] None
- [ ] Yes

## Related issues

Fixes #906
Fixes #985
Related: #1048, #1037, #1071, #959, #1034, #907, #1028, #961

---


mvanhorn:
Merged. Thank you to everyone whose work is in this change: @immy1991 (#1048), @terry-li-hm (#1037 and the #985 measurements), @SomSamantray (#1071), @technicianofthesacred (#959), @nish3451 (#1034 and the #1028 prior art), and @b33eep (#907). The squash commit carries Co-authored-by credits for the cherry-picked work.

PR #1092, opened 2026-09-01
fix(hooks): use bare CLAUDE_PLUGIN_ROOT in SessionStart

### LINKED ISSUE #1074: `hooks.json`: `${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}` resolves to the session cwd, and slips past Claude Code's own missing-plugin-root guard
## Summary

`hooks/hooks.json` runs the SessionStart hook with:

```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh\""
```

If neither variable is set, that resolves to `./hooks/scripts/check-config.sh`, relative to the session's working directory. Any repository that happens to contain that path gets its own script executed at session start, in place of the plugin's.

I want to lead with the argument I think actually matters, because on its own the above reads like a theoretical nitpick:

**Claude Code already refuses this exact condition, and the shipped string is the one form that slips past the refusal.** Before 2026-05-17 this file used the canonical form and inherited that refusal for free. The current form gives it up.

## The guard this opts out of

From the Claude Code 2.1.241 bundle. Identifiers are minified; the error strings are verbatim:

```js
for (let [xe, Fe] of [["CLAUDE_PLUGIN_ROOT", u || p], ["CLAUDE_PLUGIN_DATA", u]]) {
  if (Fe) continue;
  let Ye = "${" + xe + "}";
  if (!e.command.includes(Ye) && !e.args?.some((Ue) => Ue.includes(Ye))) continue;
  let $e = L$e(e);
  throw Error(p
    ? `Hook command references \${${xe}} but only \${CLAUDE_PLUGIN_ROOT} is available for skill hooks (\${CLAUDE_PLUGIN_DATA} is plugin-only). Command: ${$e}`
    : `Hook command references \${${xe}} but the hook is not associated with a plugin. This variable is only available in hooks defined in a plugin's hooks/hooks.json file, not in settings.json. Command: ${$e}`)
}
```

The test is `e.command.includes("${CLAUDE_PLUGIN_ROOT}")`, an **exact literal** match. `${CLAUDE_PLUGIN_ROOT:-...}` does not contain that substring: what follows `ROOT` is `:-`, not `}`. So the `continue` fires, the command is never examined, and the shell resolves the `:-` default instead.

For context, the same executor is why the variable is normally present at all. It is exported into the child environment rather than only substituted textually:

```js
if (u) { if (F.CLAUDE_PLUGIN_ROOT = k(u), d) F.CLAUDE_PLUGIN_DATA = k(Bgt(d)) }
...
if (p) F.CLAUDE_PLUGIN_ROOT = k(p);
```

`F` is the environment passed to the spawned shell. There is also a separate refusal if the plugin directory is missing:

```js
if (u) { if (!await kEw(m, u)) throw Error(`Plugin directory does not exist: ${u}` + ...) }
```

So with the canonical form, a missing plugin root gives a clear error naming the plugin. With the current form, it gives silent execution of whatever sits at `./hooks/scripts/check-config.sh`.

## When it changed

`hooks/hooks.json` carried the canonical form until two commits landed a minute apart on 2026-05-17:

```
5b0308b  Fix path-quoting in SessionStart check-config hook (handles spaces in CLAUDE_PLUGIN_ROOT)
e1017e9  fix: make hooks.json polyglot for Gemini CLI and Claude Code compatibility
```

`e1017e9` is a one-line change:

```diff
-"command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/scripts/check-config.sh\"",
+"command": "bash \"${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh\"",
```

The multi-host goal is entirely reasonable. It is the `:-.` tail specifically that has the side effect, and as far as I can tell that tail is not what buys the multi-host support.

## Reproduction, shell level

```bash
mkdir -p /tmp/l30d-probe/hooks/scripts
cat > /tmp/l30d-probe/hooks/scripts/check-config.sh <<'SH'
#!/bin/bash
echo "DECOY RAN from $PWD" > /tmp/l30d-decoy.txt
SH
chmod +x /tmp/l30d-probe/hooks/scripts/check-config.sh
cd /tmp/l30d-probe

# current form, both variables unset
env -u CLAUDE_PLUGIN_ROOT -u extensionPath \
  bash -c 'bash "${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh"'
cat /tmp/l30d-decoy.txt
# -> DECOY RAN from /tmp/l30d-probe

# pre-2026-05-17 form, same conditions
rm -f /tmp/l30d-decoy.txt
env -u CLAUDE_PLUGIN_ROOT \
  bash -c 'bash "${CLAUDE_PLUGIN_ROOT}/hooks/scripts/check-config.sh"'
# -> bash: /hooks/scripts/check-config.sh: No such file or directory
# -> decoy did not run; the old form fails closed
```

## Severity: Low on Claude Code, and I would rather say so than oversell it

I checked whether this is actually reachable on Claude Code before filing, and **it is not**, on the version I tested.

I built a directory containing a decoy at exactly the path the fallback resolves to and launched a real Claude Code session with that directory as the working directory. To make a negative result mean something, the probe carried its own positive control in the same session: `LAST30DAYS_MEMORY_DIR` was pointed at a scratch path, since `check-config.sh` line 23 does `mkdir -p` on it. That gave three distinguishable outcomes rather than two.

```
scratch memory dir EXISTS  -> the real check-config.sh ran, plugin root resolved
decoy marker absent        -> the fallback did not fire
```

The real hook ran and the decoy did not, with the decoy sitting at the exact fallback path. That matches the code above: for a hook declared in a plugin's own `hooks/hooks.json`, the plugin root is the directory the manifest was read from, so it is non-empty whenever the hook exists at all.

**Launch conditions, since a result without them is not reproducible:** Claude Code 2.1.241, the packaged binary invoked directly, non-interactive print mode, working directory outside any configured project. I did **not** test every launch wrapper, and I am marking that as inferred rather than measured: the plugin root is read from the plugin registry rather than from any settings file, so I do not expect a wrapper to change it, but I did not execute that variant.

I also confirmed that `hooks/scripts/check-config.sh` is not registered in any `settings.json`, which per the error string above is the configuration that would leave the hook with no associated plugin.

So: **not exploitable on Claude Code as it ships today.** What is left is a defense-in-depth deviation, and one that costs nothing to fix. I am filing it because the failure direction is the bad one. If a host ever invokes this manifest without setting either variable, the difference between the two forms is the difference between a clear error and executing a file out of the user's current directory.

## Suggested fix

**Preferred, and it is a revert of one line:**

```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/scripts/check-config.sh\""
```

This is the only form that gets the platform guard, because the guard matches that exact literal.

**Please do not reach for `${CLAUDE_PLUGIN_ROOT:?...}`.** It looks like the obvious hardening and it is strictly worse than both alternatives here. It does fail closed, but it fails closed *in bash*, one layer too late: the platform guard has already skipped the command for not containing the literal, so the user gets a shell diagnostic on the hook's stderr instead of Claude Code's message naming the plugin and telling them to reinstall. It also still does not contain the literal, so it inherits none of the platform behaviour. It buys a worse error message than doing nothing special.

**If the `extensionPath` branch must stay in one shared file,** the minimal change is to delete only the final default:

```json
"command": "bash \"${CLAUDE_PLUGIN_ROOT:-$extensionPath}/hooks/scripts/check-config.sh\""
```

With both unset that expands to `/hooks/scripts/check-config.sh`, an absolute path that fails closed exactly as the pre-2026-05-17 form did. This keeps polyglot support and removes the cwd resolution. It still does not get the platform guard, so it is second best, but it is a one-character-class change and it closes the actual issue.

## Does anything depend on the `extensionPath` branch?

Possibly not, which is why I am suggesting the plain revert first rather than the minimal change. Measured on the current tree:

- `extensionPath` appears in **exactly one place in the whole repository**: `hooks/hooks.json` line 9.
- `gemini-extension.json` has four keys, `name`, `version`, `description`, `settings`. There is no hooks key.
- None of `.codex-plugin/plugin.json`, `.grok-plugin/plugin.json`, `.grok-plugin/marketplace.json` or `.agents/plugins/marketplace.json` mentions hooks.

What I cannot determine from outside is whether Gemini CLI discovers `hooks/hooks.json` by convention without the manifest declaring it. You will know that immediately and I do not. If it does, the minimal change above is the right one; if it does not, the fallback has no live consumer and the plain revert costs nothing.

## Unrelated, and happy to file separately rather than clutter this

The sibling commit `d530c90`, one minute after `e1017e9`, removed `"timeout": 5` from the same hook entry with the message "remove timeout field per PR description intent". The hook now runs unbounded at every session start, and it does eight `security find-generic-password` probes on macOS. In normal conditions that is fast, around 140 ms measured over five runs, so this is minor. But a SessionStart hook with no ceiling is a session that can hang on a blocked keychain call, and the timeout looks like it was dropped as PR cleanup rather than as a decision. Say the word and I will open a separate issue instead of leaving it as a footnote here.

## Environment

- Plugin: `3.21.1`, installed from the marketplace, running at commit `d05389d`
- Claude Code: `2.1.241`
- macOS, `bash` as the hook shell

khsaurabh:
## Summary

`hooks/hooks.json` SessionStart used a defaulting form for CLAUDE_PLUGIN_ROOT that does not contain the literal `${CLAUDE_PLUGIN_ROOT}`. Claude Code's missing-plugin-root guard matches that exact substring, so the guard never fired and the `.` default resolved relative to the session cwd (any repo with `hooks/scripts/check-config.sh` could run instead of the plugin's).

Preferred fix from #1074: bare `${CLAUDE_PLUGIN_ROOT}` so the platform guard applies again. `extensionPath` appears nowhere else in the tree; Gemini uses `GEMINI_EXTENSION_DIR` in gemini-extension.json, not this hook path.

## Testing

- [x] `uv run pytest tests/test_plugin_contract.py tests/test_check_config_env_safety.py`
- [x] New contract test fails if the old cwd-default form is restored (sabotage)
- [x] Shell repro: old form runs a cwd decoy; new form fails closed with both vars unset

## Changelog

- [x] Added changelog.d/1074.fixed.md

## Agent disclosure

- **Harness:** Hermes Agent (opensource-builder)
- **Model:** grok-composer-2.5-fast (xAI)

## Relationship

No paid relationship with the maintainer or competing products.

## Related issues

Closes #1074

### DIFF
--- a/changelog.d/1074.fixed.md
+++ b/changelog.d/1074.fixed.md
@@ -1,0 +1,3 @@
+SessionStart hook must use the bare ${CLAUDE_PLUGIN_ROOT} form so Claude Code's
+missing-plugin-root guard can match it. A default of `.` resolves relative to the
+session cwd and can execute a decoy check-config.sh from the user's project.
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash \"${CLAUDE_PLUGIN_ROOT:-${extensionPath:-.}}/hooks/scripts/check-config.sh\""
+            "command": "bash \"${CLAUDE_PLUGIN_ROOT}/hooks/scripts/check-config.sh\""
           }
         ]
       }
--- a/tests/test_plugin_contract.py
+++ b/tests/test_plugin_contract.py
@@ -126,5 +126,16 @@ class TestPluginContract(unittest.TestCase):
 
         self.assertEqual([], offenders)
 
+    def test_session_start_hook_uses_bare_claude_plugin_root(self) -> None:
+        # Claude Code's missing-plugin-root guard matches the exact literal
+        # ${CLAUDE_PLUGIN_ROOT}. Defaults like ${CLAUDE_PLUGIN_ROOT:-.} slip past
+        # that check and resolve relative to the session cwd (issue #1074).
+        hooks = _json(ROOT / "hooks" / "hooks.json")
+        command = hooks["hooks"]["SessionStart"][0]["hooks"][0]["command"]
+        self.assertIn("${CLAUDE_PLUGIN_ROOT}", command)
+        self.assertNotIn(":-", command)
+        self.assertNotIn("extensionPath", command)
+        self.assertNotIn(":-.}", command)
+
 if __name__ == "__main__":
     unittest.main()


mvanhorn:
Maintainer review complete (ce-code-review run latest20-1092); no actionable findings. 9 plugin contract tests pass. Manual decoy CWD did not execute with root unset; configured plugin root executed only plugin fixture. Claude host missing-root guard was not exercised in a live host; literal contract and shell behavior verified. Review was sequential inline, not independent agents.

Issue #1051, opened 2026-08-25
Grok X backend: all responses rejected as "no items parsed" — grok CLI invoked without --output-format json


PDB06:
### Summary

With `LAST30DAYS_X_BACKEND=grok` and a signed-in Grok CLI, every X search is discarded with `rejected response: no items parsed`, and the run ends with `X: 0 posts`. The Grok CLI itself returns good data — the engine just can't parse it, because `grok` is invoked without `--output-format json`.

### Environment

- last30days v3.21.1 (Claude Code plugin install)
- macOS 14.6 (darwin 23.6.0), Python 3.14.6
- `grok` CLI 1.0.5, signed in (`~/.grok/auth.json` present, `grok_x.stored_auth_status()` → `('ok', ...)`)

### Repro

```
LAST30DAYS_X_BACKEND=grok python3 last30days.py "Seedance" --search x --no-browser-cookies --emit=md
```

Output:

```
[Grok] searching: Seedance since:2026-07-26 until:2026-08-25
[Grok] rejected response: no items parsed
[Grok] searching: Seedance since:2026-07-26 until:2026-08-25 min_faves:5
[Grok] rejected response: no items parsed
...
✓ Research complete (84.4s) - X: 0 posts
```

### Root cause

`lib/grok_x.py:693` invokes the CLI in free-text mode:

```python
result = subprocess.run(
    [binary, "-p", prompt, "--permission-mode", "bypassPermissions"],
    ...
)
```

Grok 1.0.5 narrates its tool use before emitting the fenced JSON block, e.g.:

```
I'll search X for Seedance posts from the last 7 days and return only the JSON array.
Checking the X search workflow first.The X recent-search path is the right one...
```json
[ { "url": "https://x.com/...", "text": "...", "likes": 2041 }, ... ]
```

The data is correct and complete — the preamble is what the parser trips on.

### Suggested fix

The CLI already supports structured output:

- `--output-format json`
- `--json-schema <SCHEMA>` — "the model is constrained to produce JSON matching this schema. Implies `--output-format json`"

Passing a schema for the expected item shape removes the parsing ambiguity entirely. Confirmed working locally with a schema of `{posts: [{url, author, text, likes, reposts, date}]}`.

### Secondary observation: call volume on the free tier

[...]

Issue #1043, opened 2026-08-23
v3.21.1: YouTube items dropped after successful transcript fetch (final count always 0)


T-B-Byte:
## Summary

On v3.21.1, YouTube videos are discovered and their transcripts fetched successfully, but zero YouTube items survive into the final results. Every run reports `YouTube: 0 videos`.

## Reproduction

Environment: macOS (Darwin 25.6.0), Apple M3, Python 3.13, Claude Code, skill installed as a git checkout at tag `v3.21.1`. `FUN_LEVEL` and `ELI5_MODE` unset; config contains only `SETUP_COMPLETE`, `BROWSER_CONSENT`, `FROM_BROWSER`. `--diagnose` reports `status: ready`, `safe: true`, youtube in `available_sources`, x_backend `xai`.

```
python3 scripts/last30days.py "Informa TechTarget" --emit=compact \
  --save-dir=<dir> --save-suffix=sally2 \
  --plan <planfile> --subreddits=B2BMarketing,martech,marketing,sales
```

The plan listed `youtube` in `sources` for both subqueries.

## Observed

Log shows transcripts fetched three times:

```
Got transcripts for 6/6
Got transcripts for 6/6
Got transcripts for 6/6
```

Final line of the same run:

```
Research complete (262.1s) - Reddit: 14 threads, X: 10 posts, YouTube: 0 videos,
HN: 0 stories, Polymarket: 0 markets, Web: 0 results, Github: 2 results, Jobs: 0 results
```

No YouTube items appear anywhere in the saved raw file.

## Scope

Not topic-specific and not new. Across 7 runs on v3.3.1 (B2B company topics: HubSpot, Apollo.io, Ingram Micro, Arrow Electronics, TechTarget, DemandScience, Foundry IDG) the logs showed 19 successful transcript fetches:

- 10x `Got transcripts for 6/6 videos`
- 5x `Got transcripts for 5/6 videos`
- 3x `Got transcripts for 2/2 videos`
- 1x `Got transcripts for 2/6 videos`

All 7 runs ended with `YouTube: 0 videos`. The same behavior reproduces on v3.21.1.

Low YouTube volume is plausible for some B2B company topics, but not zero across every run *after* transcripts were successfully fetched. Something between transcript fetch and ranking/dedup appears to be discarding the items, which also means the transcript fetch cost is being paid for nothing.

## Two other issues, now fixed, noted for completeness

[...]

PR #1093, opened 2026-09-01
fix(planner): honor operator --plan subquery sources at default/deep

### LINKED ISSUE #1073: Explicit --plan: per-subquery `sources` is required by validation and discarded outside --quick
### Summary

`validate_external_plan()` requires a non-empty `sources` array on every subquery of an explicit `--plan`, and `_trim_subqueries_for_depth()` then discards it at any depth other than `--quick`. A caller who passes a plan through `--plan` gets a hard validation error for omitting the field, and silent replacement of the value when they provide it.

`pipeline.py` calls the explicit plan a contract right before this happens:

```python
# External plan provided (e.g., from Claude Code via --plan flag).
# Explicit input is a contract: validate it before permissive sanitization.
planner.validate_external_plan(external_plan)
```

The override itself looks deliberate and reasonable for LLM-authored plans, and the comment says so. It just doesn't distinguish them from an operator-authored `--plan`.

### Steps to Reproduce

Deterministic, no network needed:

```python
from lib import planner

raw = {
    "intent": "opinion",
    "freshness_mode": "balanced_recent",
    "cluster_mode": "debate",
    "subqueries": [{
        "label": "primary",
        "search_query": "late diagnosed autism adults",
        "ranking_query": "late diagnosed autism adults",
        "sources": ["reddit", "x", "youtube"],
        "weight": 1.0,
    }],
}
avail = ["reddit", "x", "youtube", "hackernews", "polymarket", "github"]

for depth in ("quick", "default", "deep"):
    plan = planner._sanitize_plan(raw, "late diagnosed autism adults", avail, None, depth)
    print(depth, plan.subqueries[0].sources)
```

Output on v3.18.3:

```
quick    ['reddit', 'x']
default  ['reddit', 'x', 'youtube', 'hackernews', 'polymarket', 'github']
deep     ['reddit', 'x', 'youtube', 'hackernews', 'polymarket', 'github']
```

Removing `"sources"` from the subquery instead raises `ValueError: subqueries[0].sources must be a non-empty string array` (`planner.py:197-201`).

The relevant block is `planner.py:583-594`:

```python
if depth != "quick":
    expanded_sources = _default_sources_for_intent(intent, available_sources)
    return [schema.SubQuery(..., sources=expanded_sources, ...) for subquery in subqueries]
```

I diffed the function against `v3.21.1` and it is byte-identical to `v3.18.3` (73 lines each), so this is current behavior, not something already fixed.

### Expected Behavior

One of these would resolve the mismatch, in rough order of how little they change:

1. **Document it.** Say in `SKILL.md` that per-subquery `sources` is advisory outside `--quick`, and that `--search` is the way to restrict. The plan examples in `SKILL.md:1322-1352` currently show three subqueries with three *different* source lists, which teaches routing that the default path does not perform.
2. **Stop requiring the field** when it is going to be discarded, so the contract is at least self-consistent.
3. **Honor it for external plans.** Keep the expansion for LLM-authored plans, where the rationale in the comment applies, and pass a flag through `_sanitize_plan` so an operator-supplied `--plan` is respected. `_trim_subqueries_for_depth` already honors per-subquery sources on the `quick` branch, so the machinery exists.

I am not arguing the override is wrong by default. Letting fusion decide quality makes sense for a plan the model just invented. It reads differently when a human wrote the plan on purpose.

Practical effect on our side: two runs on a Brazilian-Portuguese neurodivergence topic queried `hackernews`, `polymarket` and `github` on every subquery and got 0 items from all three, because the plan's restriction to `reddit`/`x`/`youtube` was dropped. `--search reddit,x,youtube` works as a workaround, though it is global rather than per-subquery.

### Error / Traceback

No traceback. The value is replaced silently; the only visible trace is the planner log line showing every subquery with the full source list:

```
[Planner]   sq1 label=primary search="..." sources=[reddit,x,youtube,hackernews,polymarket,github]
```

### Install Method

Manual (git clone), vendored copy of `skills/last30days/` at v3.18.3.

### OS

macOS 26.5.2, Python 3.13.3

TheGusishere:
## Summary

Honor per-subquery `sources` on an operator-supplied `--plan` at default and deep depth. `_trim_subqueries_for_depth` was replacing those lists with the full available source set, so a plan that named `reddit`/`x`/`youtube` silently queried `hackernews`/`polymarket`/`github` too. Engine-internal LLM plans still expand narrow lists so fusion can decide quality.

## Testing

- [x] `uv run pytest` — local `python3 -m pytest tests/` green (this environment has pytest via pip, not `uv`); the five new #1073 cases passed

New coverage:
- `tests/test_planner_v3.py` — issue #1073 snippet with `honor_plan_sources=True` (keeps `reddit,x,youtube` at default/deep) and without the flag (still expands); distinct per-subquery lists preserved
- `tests/test_internals_v3.py` — `_trim_subqueries_for_depth(..., honor_plan_sources=True)` skips expansion
- `tests/test_pipeline_v3.py` — `pipeline.run(..., external_plan=...)` at default does not fetch the extra sources

## Changelog

- [x] Added `changelog.d/1073.fixed.md` (types: `added`, `changed`, `fixed`, `removed`, `deprecated`, `security`)

### AI review

Implemented by a Cursor cloud coding agent on TheGusishere's fork. The issue's option 3 / hunch was verified against current `planner._sanitize_plan` and `_trim_subqueries_for_depth`: without a flag, the #1073 snippet still expands at default/deep. The flag is keyword-only on `_sanitize_plan` and defaults to `False`, so `plan_query` (engine-internal LLM) is unchanged. Pipeline's external `--plan` path passes `honor_plan_sources=True`.

Main risks checked:
- LLM-authored plans still expand (`test_llm_plan_still_expands_narrow_sources_at_default_and_deep` plus existing `test_default_how_to_expands_past_llm_narrow_source_weights`).
- Hosting models that pass `--plan` (LAW 7) now get their per-subquery lists honored, including secondary subquery routing that SKILL.md already documented. Added one sentence in Step 0.75 so that contract is explicit.
- Did not change `validate_external_plan` (still requires non-empty `sources`). Did not touch #1043 or #1051. Did not edit `CHANGELOG.md` or bump versions.

### Security

N/A. No new credentials, command execution, or path handling. Subquery sources are still filtered to available (and `--search`) sources before the honor path returns them.

## Notes

Quick depth is unchanged: it already honored per-subquery sources, then capped. `--search` remains a global restriction on top of the plan. Pipeline safety nets (grounding / jobs / corpus injection) are unchanged and sit after sanitization.

Disclose employment, contracting, equity, or other paid ties to a company/product/service this PR adds or meaningfully promotes (example: you work at the API vendor being integrated).

- [x] None
- [ ] Yes — disclosure:

## Related issues

Fixes #1073


mvanhorn:
Closed the remaining fallback hole: unavailable-only external subqueries are skipped with non-TTY-visible diagnostics; an all-empty plan fails before retrieval instead of searching unrequested sources. Six regression cases failed before the fix across quick/default/deep; the seven new tests and planner, pipeline, internals, source-outcome and source-log suites pass after it. ce-code-review complete (latest20-1093), sequential inline, no actionable findings. Maintainer should inspect the first explicit narrow plan after merge: backend calls must remain a subset of its requested sources; revert on unintended expansion or rejection of a valid available source.

PR #1108, opened 2026-09-06
fix(auth): prefer Codex auth over OPENAI_API_KEY fallback

Crich1187:
Prefers Codex's own authenticated session over falling back to `OPENAI_API_KEY`, so a machine with Codex auth configured doesn't silently use a stale/unintended API key.

## Changes
- `scripts/lib/env.py`: auth-precedence fix (prefer Codex auth) plus a related cleanup that removes now-redundant OpenAI key fallback paths, and a small set of additive provider-availability stubs (Bluesky, TruthSocial, XQuik, YouTube SC, Xiaohongshu) that were sitting on the same branch.
- `tests/test_codex_auth.py`: new coverage for the auth-precedence behavior.

## Notes
This branch was developed and committed locally some time ago; this PR is the first time it's being pushed upstream (prior push attempts hit auth issues on our end, now resolved). The three commits are pushed as originally authored, unmodified.


mvanhorn:
Thanks @Crich1187. The optional Codex-auth idea is worth keeping, but closing this implementation: the auth/config replacement drops existing API-key and source-availability behavior, breaks unrelated source paths, and couples us to private token/backend handling previously removed from main.

Attributed follow-up roadmap (idea retained from this PR; no code merged): an explicitly selected adapter around the supported noninteractive Codex CLI for scheduled/scripted engine runs whose host is not already doing planning/judging. Use normal CLI-managed login; preserve existing providers, defaults, and source credentials. Never silently fall back to a billed provider when login, quota, or output fails. Normal /last30days inside Codex already has a host model, so spawning another model should not be the default.

This is deferred feature design, not implementation in the current PR cleanup. Reference: https://developers.openai.com/codex/noninteractive/ .

PR #1109, opened 2026-09-07
harden: detected non-static command inside command in run.go

anupamme:
## Summary
Harden input handling in `mcp/internal/engine/run.go` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | go.lang.security.audit.dangerous-exec-command.dangerous-exec-command |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `go.lang.security.audit.dangerous-exec-command.dangerous-exec-command` |
| **File** | `mcp/internal/engine/run.go:87` |
| **Assessment** | Defensive hardening |

**Description**: Detected non-static command inside Command. Audit the input to 'exec.Command'. If unverified user data can reach this call site, this is a code injection vulnerability. A malicious actor can inject a malicious script to execute arbitrary code.

## Threat Model Context

This appears to be an internal/admin endpoint with restricted access. This is a Python library - vulnerabilities affect applications that import this code.

## Changes
- `mcp/internal/engine/run.go`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/changelog.d/1109.security.md
+++ b/changelog.d/1109.security.md
@@ -1,0 +1,3 @@
+Require an absolute path for automatically discovered MCP Python interpreters,
+including when Go's built-in relative-path protection is explicitly disabled.
+Explicit interpreter overrides retain their existing behavior.
--- a/mcp/internal/engine/run.go
+++ b/mcp/internal/engine/run.go
@@ -121,7 +121,12 @@ func resolvePython(override string) (string, error) {
 		return override, nil
 	}
 	path, err := exec.LookPath(DefaultPythonBinary)
-	if err == nil {
+	// Reject relative results: on older Go runtimes (and PATH entries that
+	// include "."), LookPath can resolve to a file in the current working
+	// directory instead of a trusted install, letting an attacker who
+	// controls the CWD substitute an arbitrary binary for the interpreter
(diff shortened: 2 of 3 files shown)


mvanhorn:
Maintainer amendment: kept the original guard, narrowed the explanation to Go's ErrDot opt-out, added the required security fragment, and added regression tests for relative PATH rejection, absolute PATH acceptance, and explicit overrides. The rejection test failed before the guard; all Go tests and go vet pass after it.

Review: ce-code-review complete, run latest20-1109, no actionable findings. Sequential inline review per project-owner instructions; no independent/cross-model review claimed. Windows-specific executable coverage was not run locally.

Owner: maintainer, first MCP smoke run after merge. Confirm a normal absolute PATH starts Python, an explicitly configured interpreter still works, and relative-only auto-discovery fails with the install guidance. Investigate new `engine: python3 not found on PATH` reports from otherwise valid absolute installations; revert this narrow change if that regression is confirmed.

PR #1118, opened 2026-09-08
fix(cookies): anchor Safari cookie host matching to the domain

iliaal:
## Summary

Safari cookie extraction matched hosts with an unanchored substring test, `domain in c["url"]`. Callers pass the dotted form (`env.COOKIE_DOMAINS`: `.x.com`, `.truthsocial.com`), so a stored cookie on an attacker-registrable host such as `x.com.evil.tld` or `api.x.com.evil.tld` satisfied a request for `.x.com`, while a legitimate host-only `x.com` cookie was **missed**. Both directions are fixed by matching on host equality or a true subdomain suffix.

## Testing

- [x] `uv run pytest`

Probed old versus new against the real caller value `.x.com` rather than the docstring's undotted example:

| stored host | old | new |
|---|---|---|
| `x.com` (legitimate host-only) | no match | match |
| `.x.com` (legitimate domain cookie) | match | match |
| `api.x.com` (legitimate subdomain) | match | match |
| `.x.com.evil.tld` | match | no match |
| `api.x.com.evil.tld` | match | no match |

28 parametrized cases in `TestDomainMatching` cover exact host, leading-dot storage on either side, subdomain, case folding, empty host or empty requested domain, parent-domain rejection, and eight hosts that merely contain the domain text. Negative control: restoring `domain in c["url"]` fails 11 of them.

`test_domain_substring_match_with_leading_dot` was renamed to `test_leading_dot_stored_host_matches_bare_domain` since "substring" no longer describes the contract; its assertions are unchanged.

## Changelog

- [x] Added `changelog.d/+safari-cookie-domain-match.security.md` (`security`)

### AI review

Found during a full-codebase review. The matcher was probed in both directions against the real caller values, and the consumption path was traced to the end before rating it.

### Security

This is cookie **substitution**, not exfiltration, and the difference matters for how alarming it is. The domain is the lookup key and the destination is fixed per caller, so the user's real X cookie never travels anywhere new. What the old test allowed is the reverse: a cookie named `auth_token` or `ct0` sitting on a host the attacker can register gets read out of the local Safari store and sent to X as the user's credentials (`env.py:940-942` → `bird_x.py:75,82-86` → the vendored client's `cookie` header and `x-csrf-token`). Truth Social has the same shape via `_session_id` → `TRUTHSOCIAL_TOKEN` → `Authorization: Bearer`. The realistic outcomes are a failed auth or acting as the wrong session, and it requires a hostile cookie already in the user's Safari store.

The missed host-only cookie is the plainer bug: users whose Safari stored `x.com` without a leading dot silently got no credentials at all.

`notx.com` and `evilx.com` match only under the docstring's undotted `x.com` form, which no caller passes today. They are covered by tests anyway, because the docstring advertised that form and the next caller could use it.

No new input handling, command execution, or dependency change. Tests use obviously fake cookie values only.

## Notes

`_host_matches` strips a leading dot from both sides and lowercases before comparing, because Safari stores domain cookies as `.x.com` and host-only cookies as `x.com` and either spelling can arrive from either side. The docstring now states the real contract instead of promising a substring match.

### Scope rationale

Not addressed here: `extract_safari_cookies_macos` still takes the domain as a plain string, so nothing stops a future caller passing a public suffix such as `com` and matching everything. The fix rejects `com` for the current callers, but a typed or validated domain argument is a broader change and wants its own PR.

- [x] None

## Related issues

N/A

### DIFF
--- a/changelog.d/+safari-cookie-domain-match.security.md
+++ b/changelog.d/+safari-cookie-domain-match.security.md
@@ -1,0 +1,1 @@
+Safari cookie extraction now matches the cookie host exactly or as a true subdomain, so a session cookie stored for an unrelated host such as `x.com.evil.tld` is no longer picked up for `x.com`.
--- a/skills/last30days/scripts/lib/safari_cookies.py
+++ b/skills/last30days/scripts/lib/safari_cookies.py
@@ -98,7 +98,10 @@ def extract_safari_cookies_macos(
     Extract cookies from Safari on macOS.
 
     Args:
-        domain: Domain to match (substring match, e.g. "x.com")
+        domain: Registrable host to match, with or without a leading dot
+            (e.g. "x.com" or ".x.com"). A stored cookie host matches when it
+            equals the domain or is a subdomain of it; unrelated hosts that
+            merely contain the text (e.g. "x.com.evil.tld") do not.
         cookie_names: List of cookie names to extract (e.g. ["auth_token", "ct0"])
 
     Returns:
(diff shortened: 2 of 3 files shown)


mvanhorn:
Speaking as Matt's firstmate:

PR #1119, opened 2026-09-08
fix(reddit): clamp engagement before log10 so downvoted posts don't crash ranking

iliaal:
## Summary

Reddit ranking crashed on downvoted posts. `_relevance_rank_key` computes `math.log10(total_engagement + 1)`, and Reddit reports negative scores for downvoted posts, so a post at score `-1` made the argument exactly `0` and anything below that made it negative. Both raise `ValueError: math domain error`. The key is used by `all_items.sort(key=_relevance_rank_key)` at `reddit.py:651` with no guard, so one downvoted post aborted the entire Reddit search.

## Testing

- [x] `uv run pytest`

Reproduced on unfixed code at both sites before fixing:

```
reddit          score=-1,c=0  -> ValueError: math domain error
reddit          score=-10,c=3 -> ValueError: math domain error
reddit_keyless  score=-1,c=0  -> ValueError: math domain error
reddit_keyless  score=-10,c=3 -> ValueError: math domain error
```

Producers pass negative scores through unclamped: `reddit.py:228` `_extract_score` returns `ups`/`score`/`votes` verbatim, `reddit_listing.py:126` uses `int(_attr(tag, "score") or 0)`, and `reddit_arctic.py:57` returns the archived score as-is. `reddit_public.py:184` clamps only inside its own relevance computation, not the stored score.

Three tests, each parametrized across both implementations. Negative control: with the clamp reverted, the negative-engagement cases fail on both. The zero and non-negative cases pass either way by design: they pin that scores for existing inputs are unchanged, asserting the exact value `0.3 + log10(8)/20`.

## Changelog

- [x] Added `changelog.d/+reddit-negative-engagement.fixed.md` (`fixed`)

### AI review

Found during a full-codebase review. Both call sites were traced to their producers to confirm negative values actually reach the log, and the crash was reproduced rather than inferred from the arithmetic. The blast radius was checked too: the unguarded `sort` is what turns a bad score into a failed search rather than a mis-ranked post.

### Security

`N/A`. No input handling, command execution, path handling, or dependency change.

## Notes

`max(0, ...)` clamps at the log call rather than at the producers, so a downvoted post scores the engagement floor and still ranks by relevance. Clamping in `_extract_score` instead would have discarded the sign everywhere it is read, including anywhere a caller legitimately wants to know a post was downvoted.

### Scope rationale

Not addressed here: there is no test at the `search_reddit` level proving a mixed result set with one downvoted post completes. The crash source is covered directly, and a search-level test would need the whole fetch path mocked for no additional signal about this defect.

- [x] None

## Related issues

N/A

### DIFF
--- a/changelog.d/+reddit-negative-engagement.fixed.md
+++ b/changelog.d/+reddit-negative-engagement.fixed.md
@@ -1,0 +1,1 @@
+Reddit ranking no longer crashes with a `math domain error` when a downvoted post has a negative score; such posts now receive the minimum engagement bonus.
--- a/skills/last30days/scripts/lib/reddit.py
+++ b/skills/last30days/scripts/lib/reddit.py
@@ -264,7 +264,7 @@ def _relevance_rank_key(item: Dict[str, Any]) -> float:
     ~0) above an on-topic one (relevance >= RELEVANCE_FLOOR).
     """
     rel = item.get("relevance") or 0.0
-    eng_bonus = min(0.25, math.log10(_total_engagement(item) + 1) / 20.0)
+    eng_bonus = min(0.25, math.log10(max(0, _total_engagement(item)) + 1) / 20.0)
     return rel + eng_bonus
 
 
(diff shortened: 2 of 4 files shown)


mvanhorn:
Speaking as Matt's firstmate:

PR #1121, opened 2026-09-08
fix(env): strip trailing inline comments from unquoted .env values

iliaal:
## Summary

`.env` values kept everything after `=`, so a trailing comment became part of the value. `CONFIGURATION.md:36` documents exactly the shape that breaks:

```
LAST30DAYS_MEMORY_DIR=~/Documents/Last30Days                      # POSIX — defaults to this path when unset
```

That line stored `~/Documents/Last30Days                      # POSIX — defaults to this path when unset` as the save path. A user who uncommented the documented example got a broken directory.

## Testing

- [x] `uv run pytest`

Probed `load_env_file` on a temp file rather than reasoning about it. Old parser versus new, on that documented line:

```
OLD: {'LAST30DAYS_MEMORY_DIR': '~/Documents/Last30Days                      # POSIX — defaults to this path when unset'}
NEW: {'LAST30DAYS_MEMORY_DIR': '~/Documents/Last30Days'}
```

Only the trailing-comment form was wrong; `BAZ=value#nothash`, `QUX="value # not a comment"`, leading whitespace, whole-line comments, and empty values were already handled correctly and now have tests pinning them so this change cannot quietly alter them.

10 tests, 5 of which fail against `origin/main`'s `env.py`; the other 5 pin the already-correct behaviour and pass either way. Also ran the other direct `load_env_file` callers (`tests/test_env_*.py`, `test_secret_hygiene.py`, `test_setup_wizard.py`, `test_check_config_env_safety.py`): 183 passed.

## Changelog

- [x] Added `changelog.d/+env-inline-comment-parsing.fixed.md` (`fixed`)

### AI review

Found during a full-codebase review. The parser was run against all six line shapes before anything was changed, so the fix is scoped to the one form that was actually wrong rather than to a rewrite of the parser.

### Security

This function loads credentials, so the risk of a parsing change is corrupting a key rather than leaking one. The convention was chosen to be conservative in that direction: `#` opens a comment only as the first non-blank character or when preceded by whitespace, so a key containing `#` (`TOKEN=abc#123`) is untouched, and nothing inside matching quotes is stripped. An unterminated quote falls through to the pre-existing handling unchanged.

A value that is unquoted *and* legitimately contains ` #` would now be truncated. That is the documented `.env` convention across shells and python-dotenv, and `CONFIGURATION.md` now tells users to quote such values. No secret can be set to empty as a result: empty values were already ignored and still are.

Tests use obviously fake values only. No command execution, path handling, or dependency change.

## Notes

`CONFIGURATION.md` gains a "File syntax" paragraph stating the convention explicitly, including the `LAST30DAYS_YT_PLAYER_CLIENT=` empty-value exception so it does not contradict the existing text at `:171`.

### Scope rationale

Not addressed here: `.env` still supports no backslash escapes and no `$VAR` expansion. That is now documented rather than implemented, since adding either changes what existing files mean.

- [x] None

## Related issues

N/A


mvanhorn:
Speaking as Matt's firstmate:

Captain OK via Firstmate (FM-L30-0912C): merging this PR as corrective with green CI.

PR #1123, opened 2026-09-08
test(backends): isolate the grok off-PATH probe from the real filesystem

iliaal:
## Summary

`test_no_grok_cli_is_missing` failed on any machine with grok installed in an installer directory. `_x_env` patches `lib.backends.which` to simulate an absent CLI, but `backends._grok_finding` then calls `health._off_path_binary("grok")`, which scans real directories including `~/.local/bin`. On such a host the probe took the off-PATH branch and returned "grok is installed at ... but that directory is not on this process's PATH" instead of "not found on PATH", so the test failed locally while passing in CI.

## Testing

- [x] `uv run pytest`

The failure reproduced on this host, which has `~/.local/bin/grok`. Negative control: with the new patch removed, both the original test and the new one fail; with it in place the file is green.

`_x_env` gains a `grok_off_path` parameter so the branch is driven explicitly, which also gives the off-PATH path its first deliberate coverage. Until now it only ever executed on developer machines that happened to have grok installed off PATH, where it made an unrelated test fail rather than asserting anything.

`test_backend_descriptors.py` was the only test file touching this code without isolating it; the other seven that reference `_off_path` or `installer_bin_dirs` already patch them.

## Changelog

- [ ] Added `changelog.d/<pr-or-issue>.<type>.md` (types: `added`, `changed`, `fixed`, `removed`, `deprecated`, `security`)

### AI review

Found while running the suite across several branches during a full-codebase review: the same test failed on every branch and on untouched `main`, which ruled out the branches as the cause. Both product branches were then read and confirmed correct; the defect is entirely in the test's isolation, so no product code changes here.

### Security

`N/A`. Test-only change. No input handling, command execution, path handling, or dependency change.

## Notes

Only `grok` is exposed to this: `backends.py` has an off-PATH branch for grok and not for `xurl`, so the neighbouring `xurl` assertions on "not found on PATH" cannot be affected by the filesystem.

### Scope rationale

Not addressed here: `health._off_path_binary` reads real directories by design, and other suites patch it individually rather than through a shared autouse fixture. A repo-wide fixture would be the durable answer to this class, but it changes every backend test's environment at once and wants its own PR.

- [x] None

## Related issues

N/A

### DIFF
--- a/tests/test_backend_descriptors.py
+++ b/tests/test_backend_descriptors.py
@@ -19,6 +19,7 @@ Covers the plan's U2 scenarios:
      (env.x_backend_chain()[0]) across three config permutations.
 """
 
+from pathlib import Path
 from unittest import mock
 
 import pytest
@@ -57,6 +58,7 @@ def _x_env(
     grok_installed=False,
     grok_authed=False,
     grok_expired=False,
+    grok_off_path=None,
 ):
     """Context managers configuring the X-chain probe environment.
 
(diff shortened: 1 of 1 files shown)


mvanhorn:
Speaking as Matt's firstmate:

Captain OK via Firstmate (FM-L30-0912C): merging this PR as corrective with green CI.

PR #1127, opened 2026-09-11
fix(tests): make the both-mode outcome fixture date relative

dzivkovi:
## Summary

`test_pipeline_records_both_mode_semantic_leg_failure_as_partial` hardcodes `"date": "2026-08-10"` on its fixture item. Once that date falls outside the 30-day window the item is dropped, the pipeline reports ERROR instead of PARTIAL, and the test fails on every branch on the same calendar day. It has been red since 2026-09-10. This PR makes the fixture date relative to today so the test keeps asserting the intended behavior.

## Testing

- [x] `uv run pytest`

Verified against current `main` (`ca9d415`): the test fails with the fixed date and passes with the relative one; the rest of `tests/test_source_outcomes.py` (55 tests) is green, and a full-suite failure-list diff before and after shows this test as the only change. The date is computed as five days before today in UTC, so it stays well inside the window regardless of the runner's timezone.

## Changelog

### AI review

Prepared with Claude Code. Risk checked: that the relative date still lands inside the window under every timezone the test could run in (it uses a fixed offset of a few days, not the boundary). No other files touched.

### Security

N/A

## Notes

Test-only change. A maintainer will need to add the `skip-changelog` label; a fork contributor cannot set labels on this repo.

- [x] None
- [ ] Yes — disclosure:

## Related issues

N/A

### DIFF
--- a/tests/test_source_outcomes.py
+++ b/tests/test_source_outcomes.py
@@ -1,5 +1,6 @@
 import socket
 import urllib.error
+from datetime import datetime, timedelta, timezone
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


mvanhorn:
Speaking as Matt's firstmate:

Merging — corrective test fix so the both-mode PARTIAL fixture date stays inside the 30-day window. CI green, Greptile clean, no product behavior change.

PR #1128, opened 2026-09-11
chore(tests): strip credential-shaped env vars so the suite is hermetic

dzivkovi:
## Summary

`env.get_config()` reads API keys and cookies straight from `os.environ`, so a developer machine with real credentials exported resolves config in ways a clean CI runner never does. With any `*_API_KEY` exported, six tests in `test_project_config_trust.py` and `test_check_config_env_safety.py` fail, and a real key value can appear in a pytest assertion message. This PR adds an autouse fixture in `tests/conftest.py` that strips credential-shaped variables for the duration of each test. Tests that need a credential still set it themselves.

## Testing

- [x] `uv run pytest`

Verified two ways against current `main` (`ca9d415`):

- Full suite on a developer machine with real API keys exported: the fixture turns six failing tests in `tests/test_project_config_trust.py` green and introduces no new failure anywhere else (failure list diffed before and after; the remaining failures on that box are Windows-only and identical on both sides).
- Isolated repro on `main` with only dummy `SCRAPECREATORS_API_KEY` and `XAI_API_KEY` exported: the two affected files go from 16 of 20 passing to 10 of 20; with the fixture they return to 16 of 20 under the same exported keys.

No fixture-specific test was added: the fixture is the regression guard for the existing tests.

## Changelog

### AI review

Prepared with Claude Code. Risks checked: that the suffix list does not remove a variable a test relies on implicitly (none found across the suite), and that `monkeypatch.delenv` restores the environment after each test (it does). Suffix list: `_API_KEY`, `_TOKEN`, `_KEY`, `_SECRET`, `_PASSWORD`, `_COOKIE`, plus the exact names `AUTH_TOKEN`, `CT0`, `SESSION_ID`.

### Security

This is a hygiene improvement: it stops a contributor's real credentials from leaking into test output or influencing test outcomes. No new input handling or command execution.

## Notes

Test-only change. A maintainer will need to add the `skip-changelog` label; a fork contributor cannot set labels on this repo.

- [x] None
- [ ] Yes — disclosure:

## Related issues

N/A

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import os
 import sys
 from pathlib import Path
 from unittest import mock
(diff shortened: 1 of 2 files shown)


dzivkovi:
The single CI failure here was `test_pipeline_records_both_mode_semantic_leg_failure_as_partial`, which is the stale fixture date fixed in #1127 and is red on `main` today independent of this change. I have rebased this branch on top of #1127 so the checks run green; once #1127 merges, this PR is back to its one conftest commit. Everything else passed: 4668 tests.

mvanhorn:
Speaking as Matt's firstmate:

Merging — test hermeticity fix stripping credential-shaped env vars in conftest so the suite matches clean CI. CI green, Greptile clean, no product behavior change.

PR #1137, opened 2026-09-14
feat: surface real reasoning-provider token usage in the agent JSON export

CrisRondaS:
## Summary

Headless callers (`--emit=json --json-profile=agent`, e.g. cron/CI/watchlist) have no way to know the real token cost of a run — the engine's internal planner + rerank calls to Gemini/OpenAI/xAI/OpenRouter never surfaced usage. This adds a `usage` field to the agent export with the real accumulated token usage for the run, `null` in mock/local-provider mode.

`ReasoningClient` now accumulates usage per call (`record_usage()`/`total_usage`). `GeminiClient` reads `usageMetadata` from the response (previously discarded by `extract_gemini_text`) — this is the one path verified end-to-end against a real API call. `OpenAIClient`/`XAIClient`/`OpenRouterClient` call `super().__init__()` so `total_usage` doesn't crash when one of them is the active provider, but their own usage capture isn't implemented yet — I don't have working keys for those to verify field names against a real response, and would rather ship a verified Gemini path than guess at the other three.

Additive only: `to_agent_export()` gains one new key, no `schema_version` bump, existing consumers ignore unknown keys.

## Testing

- [x] `uv run pytest`

Updated `tests/fixtures/agent_export_v1.json` (golden contract) to include the new `usage: null` key. Also ran a real (non-mock) invocation against the live Gemini API to confirm real numbers come through end-to-end, not just that the plumbing doesn't crash:

```
$ GEMINI_API_KEY=... uv run python3 skills/last30days/scripts/last30days.py "sourdough bread baking" --quick --days=30 --emit=json --json-profile=agent
...
"usage": {"calls": 1, "promptTokens": 712, "completionTokens": 422, "totalTokens": 1134}
```

## Changelog

- [x] Added `changelog.d/+reasoning-provider-token-usage.added.md`
- [ ] Skip changelog

### AI review

I'm an AI coding agent (Claude Code). I traced every real internal LLM call site in the engine (`planner.py`, `rerank.py` — confirmed there are only these two, both go through the same shared `ReasoningClient` instance resolved once in `resolve_runtime()`) before choosing where to accumulate usage, to avoid missing a call site or double-counting. I checked that `Report`/`Report.artifacts` aren't frozen dataclasses before mutating them, and that `artifacts` was already the established bag for this kind of out-of-band metadata (same pattern as `plan_source`/`corpus_in_export`) rather than introducing a new mechanism. I ran the full test suite plus a real API call before considering this done, not just a mock/unit pass.

### Security

No new input handling, command execution, path handling, auth, or dependency risk — this only reads an existing field (`usageMetadata`) out of a response the engine was already receiving and parsing, and threads it through existing, already-established data structures (`Report.artifacts`, the agent export dict). No new external calls, no new credentials.

## Notes

I use this engine downstream in an internal audit pipeline (Shockoe/Trovr) where the lack of real usage reporting was making our own cost accounting unreliable for the calls this engine makes on our behalf — this PR is the fix for that on our end, opened upstream since it's a real gap for any headless caller, not something specific to my setup.

Follow-up someone may want to pick up: real usage capture for OpenAI/xAI/OpenRouter (I left the accumulator wired but the actual capture unimplemented for those three, see above).

- [x] None

## Related issues

N/A — opened directly, no existing issue.


mvanhorn:
Speaking as Matt's firstmate: classified as opt-in (additive agent-export `usage` field). Greptile 3/5 (early usage snapshot; false zeros on non-Gemini; missing non-null tests), changelog `guard` red. Please address Greptile to 5/5 and restore green guard. Not proposing merge.

PR #1138, opened 2026-09-14
feat(meta-ads): opt-in Meta Ad Library source for a brand's live paid creative

mvanhorn:
## Summary

Adds an opt-in `meta_ads` research source that surfaces what a brand is *paying* to say this month, next to what everyone else is saying about it. It resolves the brand's advertiser page in the Meta Ad Library, pulls the creatives that launched inside the 30-day window, and reads back ad copy, launch date, placements, call to action, landing product, promo codes, and the spoken transcript of the newest video ads. A dedicated 📣 footer line names the advertiser page it resolved, so a wrong-company match is visible rather than silent.

Off by default and never inferred from topic shape, on the Amazon precedent: it needs `SCRAPECREATORS_API_KEY` **and** an explicit `--search meta_ads` or `INCLUDE_SOURCES=meta_ads`. A default-depth run spends at most 7 of the 10,000 free calls. Paid message only, never audience reaction: Meta publishes reach and spend for political ads alone, so commercial creatives carry no engagement.

Six units: adapter, normalization, pipeline wiring and CLI flag, rendering, doctor, docs.

## Testing

- [x] `uv run pytest` — 4,979 passed, 6 skipped; coverage 89.99% against the 84% floor

127 new tests across four files. The ones worth naming: the lane's paid-request ceiling is pinned so a silent-video page cannot blow past it; the resolver is pinned against reseller pages, generic shared words, initialism brands, and umbrella brands advertising under product-line names; each of the four honest empty states has its own assertion; and hostile advertiser names are asserted unable to forge footer rows.

Two live smoke runs against the real API were run by hand (about 9 credits). They are deliberately **not** in the suite: they depend on what a brand happens to be advertising today, so they would fail for reasons unrelated to the code.

## Changelog

- [x] Added `changelog.d/+meta-ads.added.md`

### AI review

Nine reviewers ran over the branch: eight in-process lenses (correctness, security, reliability, testing, maintainability, agent-native, project standards, institutional learnings) plus an independent cross-model adversarial pass on a different model family. Main risks checked: wrong-company attribution, paid-request budgeting under failure, partial-failure honesty, untrusted upstream text reaching the report, and activation gating.

Nine findings were applied. The ones that mattered:

- **The transcript cap counted successes, not paid requests.** An upstream with no transcripts kept the lane calling for every video creative, each one billed, blowing the documented credit ceiling while the tally still read zero. There is now a test that pins the ceiling.
- **The busiest partial name match won the topic.** A reseller page, or an unrelated page sharing one generic word, could claim a brand. Resolution is now strength-tiered, with ad volume breaking ties only inside the winning tier — volume measures spend, not identity.
- **A rate limit or expired credential mid-run discarded creatives already fetched.** Stopping calls and throwing away paid-for evidence are different things; enrichment now keeps what it has and reports partial.
- **The footer treated every non-OK outcome as failure.** Since the pipeline stamps a no-results state on any zero-item source, all four honest empty states collapsed into one generic line, and a cut-short run could print a definitive "nothing new" conclusion its own data did not support.
- **An initialism brand could never resolve**, even against its own identically-named page, because the token floor emptied both sides of the match. This is the same failure class as an existing entry in `docs/solutions/`, resurfacing in a new module; the learnings reviewer caught it by cross-referencing that entry.

Residual items are recorded in the section below rather than silently dropped.

### Security

Reviewed: a third-party API credential resolved through the shared config layer and sent as a request header (never on a command line, never logged); model- and user-supplied strings reaching request parameters; and an advertiser-page override validated to digits or an Ad Library URL parameter, with vanity URLs rejected because one live check resolved a brand-looking handle to a private person's profile.

One real issue found and fixed: advertiser names, candidate names, and promo codes come from the upstream API and reached the footer raw, so a name carrying a newline could forge extra rows in the emoji-tree footer the model relays verbatim. All are now control-character collapsed, separator-neutralized, and clipped, with tests asserting a hostile name cannot forge a row. No secrets, cookies, or real credentials appear in any fixture.

## Notes

The source resolves one advertiser page per run. Two limits are worth knowing:

- Substring containment is the weakest match tier and cannot distinguish a product line from a coincidence. It is tried last, the footer says when it was used, and `--meta-ads-page` overrides it.
- An explicit `--meta-ads-page` is trusted verbatim, so the footer cannot detect a wrong id.

Follow-up work, deliberately out of scope: per-competitor page overrides in comparison plans, multi-page enrichment for umbrella brands, and ranking long-running pre-window creatives at deep depth.

- [x] None

## Related issues

N/A


mvanhorn:
@greptileai review

mvanhorn:
Both findings fixed in 5d70f9e9.

**Generic tokens misidentify advertisers** — a shared word no longer establishes identity on its own. Within a match tier, candidates are now ranked by how distinctive the matched word is across the result set before ad volume is considered at all, and a match resting only on a word several advertisers carry is rejected outright when a word of the topic that no candidate carries went unmatched. So `Acme Kitchen` resolves to `Acme Kitchen` even when `Kitchen World` runs ten times the ads, and resolves to nothing when only the `Kitchen *` pages exist. Rarity is measured against the candidates themselves rather than a stopword list, because the same word is a category in one search and a brand's own name in another. An exact normalized-name match is exempt, since it is unambiguous.

**Ended ads labeled active** — the wording now follows the stored state instead of asserting `Active` for everything: running creatives read as running, and one that ended reads as having run until its end date. The adapter keeps ended-in-window creatives deliberately (a one-week promo push is the signal this source exists for), which is exactly why mislabeling them was wrong.

Seven tests cover the two behaviors, including the umbrella-brand case that the first cut of the distinctiveness rule broke: a brand's own name also appears across its product-line pages, so counting how many pages share a word cannot separate a brand from a category. Full suite: 4,953 passed, 6 skipped.

mvanhorn:
Good catch, and confirmed by reproducing it: a search for `Acme Kitchen` returning only `Kitchen World` did resolve that page, because with one lookalike in the results `kitchen` has a spread of one and reads as distinctive. Counting how many advertisers share a word cannot carry this rule.

Replaced the spread-based guard with the head-word requirement the engine already uses for entity grounding, where grounding keys on the head token of the primary entity because trailing words are usually descriptors. A multi-word topic carries its identity in the leading word, so a page matching only the trailing word has matched what the brand sells rather than who it is. `Acme Kitchen` against only `Kitchen World` now resolves nothing regardless of how many lookalikes are present, while `Acme Supply` still resolves and single-word topics keep their umbrella reach over product-line pages. Spread is still used, but only to order candidates inside a tier, which is what it is actually good for.

Four regression tests cover it, including the lone-lookalike case. Full suite: 4,957 passed, 6 skipped.

mvanhorn:
Fixed. The head word is now taken from the topic's primary entity rather than the raw phrase: it skips the same descriptor vocabulary the engine's other adapters strip before searching, so `best Acme grills` keys on `acme` and the brand's own page resolves. When every word is a descriptor it falls back to the leading word rather than matching anything at all.

Checking that by hand surfaced a second bug in the same guard, which I have also fixed: for a containment match it credited every topic word whenever any page word appeared anywhere in the topic, so `Grill World` could claim `best Acme grills` on the strength of `grill` and drag the brand word along with it. The participating word is now identified rather than assumed.

Seven resolution cases verified end to end, each asserted in tests: descriptor-led topics resolve the brand; a category lookalike is rejected with or without a leading descriptor; an umbrella topic still reaches its product-line pages; and an initialism still resolves through the exact tier. Full suite: 4,962 passed, 6 skipped.

mvanhorn:
Fixed, and the `Kitchen by Crate & Barrel` example is now a test case.

Position was the wrong test to lean on. A page whose entire name the topic spells out is being named by that topic, wherever in the phrase the name sits, so that now resolves on its own and the head-word rule only applies when it does not hold. The lookalike case still fails it for a concrete reason rather than a positional one: `Kitchen World` carries `world`, which a topic about `Acme Kitchen` never mentions, so the topic is not naming that company.

Nine resolution cases are asserted, covering brand-leading, descriptor-leading, brand-trailing (`Kitchen by Crate and Barrel`, `smart home Acme`, `cookware from Acme Supply`), umbrella topics, initialisms, and both lookalike rejections. Full suite: 4,966 passed, 6 skipped.

Worth stating plainly, since this is the fourth round on this one finding: name matching alone cannot settle advertiser identity in every phrasing, so the lane is deliberately tuned to fail closed. When it is unsure it resolves nothing, spends no enrichment credits, names the closest candidate, and points at `--meta-ads-page`. That trades a recoverable miss for never silently attributing one company's ads to another, which is the failure this source cannot afford.

mvanhorn:
Correct, and this one was a regression from my previous fix rather than a pre-existing gap. I reproduced it first: a page named only `Kitchen` was trivially covered by the topic `Acme Kitchen` and claimed it.

The coverage shortcut now requires the page name to account for more than one word of the topic, unless it accounts for the topic entirely. So `Crate and Barrel` still resolves from `Kitchen by Crate & Barrel`, and `Acme Co` still resolves from the single-word topic `Acme`, while a page named just `Kitchen` or just `Grills` no longer claims a brand topic on the category word alone.

Twelve resolution cases are now asserted in both directions: brand-leading, descriptor-leading, brand-trailing, umbrella, initialism, single-word topics, and four distinct wrong-entity rejections. Full suite: 4,969 passed, 6 skipped.

mvanhorn:
I checked both examples against the code rather than reasoning about them, and they land differently.

`Cart Wheel` vs topic `Acme Cartwheel`: the coverage check returns False and resolution already returns no advertiser. That half of the finding does not reproduce.

`Acme AI` vs topic `Acme Kitchen`: that one is real, but it is structurally the same case as `Acme Supply` vs `Acme Kitchen`, which an earlier round of this review asked me to make resolve and which is now a test. Both are a page sharing the topic's brand word and nothing else. There is no lexical rule that accepts one and rejects the other, so tightening this would reopen the brand-trailing finding from two rounds ago.

What I could safely take from it is shipped: within a match tier, the page accounting for more of the topic now outranks the busier one, so when `Acme Kitchen Co` and `Acme AI` both appear, the former wins even at forty times fewer ads. Asserted in a test.

Beyond that I am going to stop tuning this rule. Advertiser identity cannot be settled from name text alone across every phrasing, and the last several rounds have traded false positives against false negatives in the same spot. The lane is deliberately tuned to fail closed: when it is unsure it resolves nothing, spends no enrichment credits, names the closest candidate, and points at `--meta-ads-page`. When it does resolve on a weak tier the footer says so, and the resolved advertiser is always named so a wrong match is visible rather than silent.

Full suite: 4,970 passed, 6 skipped.

mvanhorn:
Both of these are now fixed, and both are asserted as tests.

You were right about the two mechanics. The coverage check was skipping words shorter than four characters and was matching against the topic with word boundaries removed, so a partial name could stand in for a whole one. It now requires every word of the advertiser's name to appear in the topic as a word, short words included and boundaries intact. `Acme AI` is no longer covered by `Acme Kitchen`, and `Cart Wheel` is no longer covered by `Acme Cartwheel`.

I also tightened the surrounding guard rather than leaving the head word as a way around it. A multi-word topic now has to be matched as a name: the topic either spells the advertiser's name out, or the page carries every identifying word of the topic. Sharing one word is what a category lookalike and a same-word different company both look like, and neither should receive another firm's creatives. A single-word topic is its own identity, so an umbrella brand still reaches its product-line pages.

This costs some recall on purpose. A page like `Acme Supply` against a topic about `Acme Kitchen` now resolves nothing rather than guessing; the footer names it as the closest candidate and `--meta-ads-page` recovers it in one flag. Six of my own tests asserted the looser behavior and have been rewritten to the stricter rule.

Fourteen resolution cases now pass in both directions, covering brand-leading, descriptor-leading, brand-trailing, umbrella topics, initialisms, single-word topics, and six distinct wrong-entity rejections. Full suite: 4,973 passed, 6 skipped.

PR #1140, opened 2026-09-16
fix(mcp): discover and verify compatible Python interpreters

opita04:
On hosts where the first `python3` is older than 3.12, the MCP launcher starts an unsupported interpreter even when a compatible Python is installed later on PATH. This change verifies candidates before launching research and reports how to select or install a compatible interpreter when none is found.

- Discover `python3`, versioned `python3.N`, and `python` executables in absolute PATH directories, preserving directory precedence and checking their actual versions.
- Honor `LAST30DAYS_PYTHON` as a verified explicit choice; reject invalid overrides without silently selecting another interpreter.
- Bound discovery and version probes, honor caller cancellation, and retain the trusted Go `RunOptions.PythonPath` injection seam.
- Document configuration and cover an old default, later compatible defaults, future minor-version names, invalid overrides, relative-path rejection, and probe cancellation.

This retains the Python 3.12+ requirement; it does not add Python 3.9 compatibility or install runtimes. Interpreters outside PATH can be selected explicitly. Windows executable names are recognized, but the executable-fixture tests and runtime smoke were run on macOS.

Validation:
- `go test ./...` in `mcp/` passed.
- Local macOS smoke selected Python 3.13.5 and successfully executed a script through `engine.Run` while the default `python3` was 3.9.6.
- `go vet ./...`, `go test -race ./internal/engine`, and `git diff --check` passed.


mvanhorn:
Speaking as Matt's firstmate: classified as corrective. Greptile 4/5 (PATH scan cancellation + CONFIGURATION.md placement), CI `tests` red, merge state CONFLICTING. Please resolve conflicts, address Greptile to 5/5 with zero unresolved, and get CI green. Not proposing merge.

PR #1141, opened 2026-09-17
fix: keep the runtime preflight intact under Claude Code argument substitution

alexandercroft:
## Summary

In Claude Code, `/last30days` with a multi-word topic breaks the Runtime Preflight: the harness replaces `$<digit>` tokens in `SKILL.md` with words from the arguments, so `candidate="$1"`, `path="$1"` and the `awk … $2` version lookup reach the model rewritten, and Python 3.12+ detection fails. This PR writes them as `${1}` and `$(2)` — the same shell/awk semantics, with nothing for the substitution to match — and keeps the preflight inline, so the existing contract tests stay as they are.

This is the argv half of #1024 in the smallest form I could make. That PR moves the preflight into `scripts/preflight.sh`, which trips `test_candidate_selection_accepts_any_python_3_12_or_newer` (it pins `try_last30days_python()` in `SKILL.md`), and its checks have been red since it was opened. The RECOMMENDATIONS-template half of #1024 is independent and not touched here.

## Testing

- [x] `uv run pytest` — 10 failures, and they are the same ten that fail on `main` @ 30d47be here: `test_check_config_quiet.py` (8), `test_env_doc_contract.py::test_documented_env_keys_are_registered_in_get_config` (`LAST30DAYS_QUIET` from #958 is documented but not registered in `get_config`) and `test_source_outcomes.py::test_pipeline_records_both_mode_semantic_leg_failure_as_partial`. Nothing else fails, and the new test passes.
- [x] Added or updated tests that would catch a regression: `test_code_has_no_bare_positional_parameters` scans fenced and inline code in `SKILL.md`; it fails on `main` with `['$1', '$1', '$2', '$2']` and passes here.

Also checked in plain bash (macOS awk): old and new snippets give identical results — the version lookup returns `3.21.1`; `try_last30days_python python3`, a missing path and `windows_path_to_unix '/c/Program Files'` produce the same output. In Claude Code 2.1.271 the substitution rule is `/\$(\d+)(?!\w)/g` (argument words indexed from 0, unmatched indices left alone); with a 30-word argument the patched file has no token left to rewrite. The `$100` / `$200` in prose would need 101+ words.

## Changelog

- [x] Added `changelog.d/+argv-substitution-preflight.fixed.md`

### AI review

Written with Claude Code (Opus 5) and reviewed by me. Checked that `${1}` and awk `$(2)` behave exactly like `$1` / `$2` (the `gsub` target is still an lvalue), that no other code in `SKILL.md` carries a bare positional token, and that no version strings or manifests changed.

### Security

N/A — three existing lines change spelling only; no new input handling, command execution, paths, secrets or dependencies.

## Notes

`\$1` would also work in Claude Code, but the same `SKILL.md` ships to Codex and Grok, which don't substitute and would show the backslash.

- [x] None
- [ ] Yes — disclosure:

## Related issues

Relates to #1024 (argv half). Background: anthropics/claude-code#94709.


alexandercroft:
CI note: the red `tests` job is inherited from `main`. Validate on `main` @ 30d47be ends with `2 failed, 5005 passed` (`test_env_doc_contract.py::test_documented_env_keys_are_registered_in_get_config` and `test_source_outcomes.py::test_pipeline_records_both_mode_semantic_leg_failure_as_partial`). This PR ends with the same two failures and `5006 passed`; the extra pass is the new contract test.

The eight `test_check_config_quiet.py` failures from my local run, counted in the description, pass in CI, so they are specific to my machine (most likely my global `~/.config/last30days/.env`).

mvanhorn:
Speaking as Matt's firstmate: classified as corrective. Greptile 5/5, mergeable, but Validate `tests` is red (UNSTABLE). Please push a green CI or rebase once main is green; author/CI blockers stay off the captain desk. Not proposing merge.

